PR #67, opened 2022-11-04
Add air-to-async pass for cpu lowering

fifield:
Lower air dialect to async dialect.
There is currently basic support for `air.herd, air.execute, air.wait_all`
I don't think address space conversion is fully working. The plan is to transform all memref types to have address space 0.


fifield:
rebased and moved to personal branch: https://github.com/Xilinx/mlir-air/pull/125

PR #122, opened 2022-12-15
Cmake

stephenneuendorffer:


CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ include(ExternalProject)
 
 find_package(MLIR REQUIRED CONFIG)
 find_package(AIE REQUIRED)
+set(LibXAIE_ROOT ${Sysroot}/usr ${Sysroot}/opt/xaiengine ${Sysroot}/opt/xaienginev2)
```
fifield, line 43: I don't think these is needed or wanted. The same defaults are set in https://github.com/Xilinx/cmakeModules/blob/master/FindLibXAIE.cmake or the location can be passed in ([for example in the CI](https://github.com/Xilinx/mlir-air/blob/main/.github/workflows/buildAndTest.yml#L87))

CMakeLists.txt
```diff
@@ -109,38 +110,23 @@ add_custom_target(check-all)
...
-else()
-	set(AIRHOST_TOOLCHAIN_OPT ${CMAKE_TOOLCHAIN_FILE})
-endif() 
-
-if(AIR_BUILD_RUNTIME)
-	message("AIRHOST_TOOLCHAIN_OPT=${AIRHOST_TOOLCHAIN_OPT}")
-	message("Sysroot=${Sysroot}")
-	ExternalProject_Add(air_runtime_lib
-		PREFIX ${CMAKE_CURRENT_BINARY_DIR}/runtime_libTmp
+foreach(target ${AIR_RUNTIME_TARGETS})
```
fifield, line 113: Where is AIR_RUNTIME_TARGETS set?  The runtime didn't build at all in the github action.
I see,
```
CMake variable LibXAIE_ROOT is set to:

    /usr;/opt/xaiengine;/opt/xaienginev2
```
instead of,
```
[8/185] Performing configure step for 'air_runtime_lib'
...
-- Found LibXAIE: /home/runner/work/mlir-air/mlir-air/aienginev2/install/include  
-- Configuring done
-- Generating done
...
[1/26] Building CXX object airhost/CMakeFiles/airhost.dir/queue.cpp.o
[2/26] Building CXX object airhost/CMakeFiles/airhost.dir/memory.cpp.o
...
```

fifield (changes requested):

CMakeLists.txt
```diff
@@ -109,38 +110,23 @@ add_custom_target(check-all)
...
+			-DSysroot=${${target}_Sysroot} 
 			-DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
 			-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
 			-DCMAKE_ASM_COMPILER=${CMAKE_ASM_COMPILER}
 			-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
 			-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
-			-Dpybind11_DIR=${pybind11_DIR}
-			-DLibXAIE_DIR=${LibXAIE_DIR}
 			-DAIE_DIR=${AIE_DIR}
 			-DBUILD_AIR_PCIE=${BUILD_AIR_PCIE}
```
fifield, line 131: Should these go in the target toolchain file?  That seems like the only place to pass different parameters depending on the target without adding a bunch of ad-hoc logic in the foreach loop.

CMakeLists.txt
stephenneuendorffer: It seems to me like the default hints compose with CMAKE_SYSROOT... Or at least I couldn't get it to work for me.  I'd prefer to avoid plumbing through explicit paths, which requires us to maintain it in multiple places (acdc, air CI)

CMakeLists.txt
stephenneuendorffer: Perhaps... it wasn't clear to me whether these are per-target options?

fifield: Today BUILD_AIR_PCIE is roughly synonymous with !aarch64 in the runtime code, but not quite. We really have 3 sub-targets: build with PCIe support, build with /dev/mem support, and build with /dev/mem support but insert extra error messages because you're not building for aarch64. The latter is unused but would be useful for unit-testing parts of the runtime or as an emulation runtime.

stephenneuendorffer: It's not clear to me what exactly to do with this.   Any suggestions?

jgmelber: I like the idea of using toolchain files per target. It would make sense and enable outside-AMD users. Especially because they cannot use type 2. Rather than per board, it is per system: PCIe, embedded, hybrid.

jamesroxbypb: I'm not jgmelber, but my thoughts are that platforms we want to support now and in the future are will be split between the more embedded ARM+AIE, and then the data centerish x86 + (ARM+AIE), so I do think figuring this out for the current gen is useful, even if the result is proliferation to begin with

stephenneuendorffer:
Now that the tutorial is done, I'd like to merge this.  I think the key concern I have is about getting the CIs updated so that everything gets configured correctly.  Can someone who is using the runtimes (which don't seem to be compiled in the CI) verify that this works for you?

jgmelber:
This doesn't build any runtimes as it is. I think I need to specify a list of AIR_RUNTIME_TARGETS somewhere?

jgmelber:
> This doesn't build any runtimes as it is. I think I need to specify a list of AIR_RUNTIME_TARGETS somewhere?

I think I got it to build with `-DAIR_RUNTIME_TARGETS="x86" \`, testing...

stephenneuendorffer:
> This doesn't build any runtimes as it is. I think I need to specify a list of AIR_RUNTIME_TARGETS somewhere?

Yes, you need to configure with something like:

               -DAIR_RUNTIME_TARGETS:STRING=aarch64;x864_64
               -Daarch64_TOOLCHAIN_FILE:STRING=${aarch64_TOOLCHAIN_FILE}
               -Dx86_64_TOOLCHAIN_FILE:STRING=${x86_64_TOOLCHAIN_FILE}

It's unclear to me whether this is actually sufficient, however.  Observations:
1) we actually have a host target (e.g. aarch64 or x86_64) and an aie target (AIEv1, AIEv2).  we may need to compile differently based on *both*  For instance, I might have an AIEv1 and AIEv2 device in an X86 server.
2) we could define our own target-style scheme (e.g v70-x86_64-PCIE) that would imply some of these combinations and make it easier for users of aircc/aiecc to specify things.

jgmelber (approved):
This works for me, I built using the build-mlir-air-pcie.sh script (modified in this PR) and tested on VCK5000.

PR #124, opened 2022-12-16
Add a lowering of channel put/get to airrt memcpy

fifield:
Mostly this is a cut and paste of the dma memcpy lowering.

TODO: write a test


fifield:
@abisca this is the start of the runtime side of the channel get/put lowering. The objfifo lowering will still need to produce some metadata to inform the runtime, TBD.

AndraBisca:
Thank you for the pointer @fifield ! We can discuss what other metadata should be produced as part of the channel-to-objfifo lowering and add that to the runtime next.

abisca (approved):

PR #126, opened 2022-12-20
Update AIR dialect canonicalizers

fifield:
A small bit of refactoring after noticing overlap between the dialect canonicalization and `dependencyCanonicalizer`

@erwei-xilinx please have a look at the changes to `dependencyCanonicalizer::removeRedundantWaitAllOps`. We should favor putting "cleanup" canonicalizations in the dialect whenever possible so that they also run with ordinary `-canonicalize`. I made these changes before removal of `dependencyCanonicalizer::removeRedundantAIRHierarchyArgs` but that optimization is also covered by the dialect canonicalization. I did not move `dependencyCanonicalizer::removeUnusedExecuteOp` into the dialect because it doesn't seem safe in general (what if the execute block included an op with side effects?).  `dependencyCanonicalizer::removeDepListRepetition` is covered by the dialect canonicalization for hierarchy ops, but not yet for all other ops (TODO).

### DIFF
--- a/mlir/include/air/Util/Dependency.h
+++ b/mlir/include/air/Util/Dependency.h
@@ -56,9 +56,6 @@ Value getLoopCarriedTokenFromScfOp(scf::ParallelOp op);
 Value getLoopCarriedTokenFromScfOp(scf::ForOp op,
                                    std::string operand_or_argument = "operand");
 void addAsyncDependencyIfNew(air::AsyncOpInterface op, Value token);
-std::string getMemorySpaceAsString(Value memref);
-air::ChannelGetOp getTheOtherChannelOpThroughSymbol(air::ChannelPutOp put);
-air::ChannelPutOp getTheOtherChannelOpThroughSymbol(air::ChannelGetOp get);
 
 //===----------------------------------------------------------------------===//
 // Dependency graph parsed as a Boost graph object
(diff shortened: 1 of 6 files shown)


mlir/include/air/Util/Dependency.h
```diff
@@ -56,9 +56,6 @@ Value getLoopCarriedTokenFromScfOp(scf::ParallelOp op);
 Value getLoopCarriedTokenFromScfOp(scf::ForOp op,
                                    std::string operand_or_argument = "operand");
 void addAsyncDependencyIfNew(air::AsyncOpInterface op, Value token);
-std::string getMemorySpaceAsString(Value memref);
-air::ChannelGetOp getTheOtherChannelOpThroughSymbol(air::ChannelPutOp put);
```
erwei-xilinx, line 60: Could leave the declaration of getTheOtherChannelOpThroughSymbol() here, as runner includes Dependency.h and makes use of those helper functions.

erwei-xilinx (changes requested):
I agree with your point of merging those methods in dependencyCanonicalizer to the ops' canonicalization patterns. Might be helpful to leave the getTheOtherChannelOpTHroughSymbol helper function there for runner codebase to use it.

mlir/include/air/Util/Dependency.h
erwei-xilinx: Ok I see that the declaration is under Utils.h instead. This is okay and the PR looks good to me. Thanks

erwei-xilinx (approved):

PR #129, opened 2022-12-21
AIR Dependency: Fixup channel hoisting issue with having parent nested for loops

erwei-xilinx:


fifield (approved):
I just ran into the bug solved by this, can we merge it?

erwei-xilinx:
> I just ran into the bug solved by this, can we merge it?

Sure. Will merge it soon today after CI build and test passes.

PR #135, opened 2023-01-06
AIR Runner: automatically infer the number of AIE cores from herd size

erwei-xilinx:


fifield:
I thought the function was modeling cost of a single linalg op in a single worker of a herd. If that's the case, isn't the correct number of cores 1?

https://github.com/Xilinx/mlir-air/blob/main/mlir/lib/Util/Runner.cpp#L405

erwei-xilinx:
> I thought the function was modeling cost of a single linalg op in a single worker of a herd. If that's the case, isn't the correct number of cores 1?
> 
> https://github.com/Xilinx/mlir-air/blob/main/mlir/lib/Util/Runner.cpp#L405

At the moment the runner is modelling a linalg op running on an entire herd. A herd is currently the minimum modelling granularity in the runner. I agree that as a next step we should model each core in a herd individually.

PR #136, opened 2023-01-06
Fix compilation

Ralender:
When trying to compile I had the following error:
```
ld.lld: error: undefined symbol: boost::throw_exception(std::exception const&, boost::source_location const&)
>>> referenced by Dependency.cpp
>>>               Dependency.cpp.o:(boost::xpressive::detail::matchable_ex<__gnu_cxx::__normal_iterator<char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>>::repeat(boost::xpressive::detail::quant_spec const&, boost::xpressive::detail::sequence<__gnu_cxx::__normal_iterator<char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>>&) const) in archive lib/libAIRUtil.a
```
This PR fixes this issue

### DIFF
--- a/mlir/lib/Transform/AIRDependency.cpp
+++ b/mlir/lib/Transform/AIRDependency.cpp
@@ -42,6 +42,12 @@
 void boost::throw_exception(std::exception const &e) {
   llvm_unreachable("boost exception");
 }
+#if BOOST_VERSION >= 107300
+void boost::throw_exception(std::exception const &e,
+                            boost::source_location const &) {
+  llvm_unreachable("boost exception");
+}
+#endif
 
 // boost graph
 #include <boost/graph/adjacency_list.hpp>


keryell:
Does it break the compilation with old Boost version? If not, why checking?

fifield:
> Does it break the compilation with old Boost version? If not, why checking?

It does break,
https://github.com/Xilinx/mlir-air/actions/runs/3858203432/jobs/6608578855#step:15:607
Maybe there is a different fix?

Ralender:
> I think this happens for boost 1.73.0 and newer. Can the patch check the boost version?

The boost version I had installed was 1.74 and the patch notes mention in as a change for 1.73.
So I added a macro to only add it for boost 1.73 onward

fifield:
@Ralender please run clang-format then we can merge this PR. Thanks.

PR #139, opened 2023-01-11
Add a reduction pipelining transform using affine.if

fifield:
This transformation splits a linalg on memrefs reduction into pipeline stages with partial results passed from one stage to the next. The reduction pipeline is mapped to a Nx1 or 1xN herd.

### DIFF
--- a/mlir/include/air/Dialect/AIR/AIRTransformOps.h
+++ b/mlir/include/air/Dialect/AIR/AIRTransformOps.h
@@ -9,6 +9,7 @@
 #ifndef MLIR_AIR_TRANSFORM_OPS_H
 #define MLIR_AIR_TRANSFORM_OPS_H
 
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/PDL/IR/PDLTypes.h"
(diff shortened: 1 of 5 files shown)


jamesroxbypb (commented):
This looks handy.  Presumably the tile-size option can be n-dimensional and the type is inferred from the type of the tensor on the reduction dimension.  Are there ever any cases where you have more than 1 reduction dimension on a linalg op?

fifield:
> This looks handy. Presumably the tile-size option can be n-dimensional and the type is inferred from the type of the tensor on the reduction dimension. Are there ever any cases where you have more than 1 reduction dimension on a linalg op?

In the current form it only tiles along the inner dimension, which must be a reduction dimension, and tile-size applies to that dimension. There are linalg ops such as convolutions with more than one reduction dimension.

jamesroxbypb:
Does the depth of 2 include the special last stage to do just summation?

fifield:
> Does the depth of 2 include the special last stage to do just summation?

No, this does an even split of the linalg op across the pipeline stages. A special stage at the end is not emitted.

PR #140, opened 2023-01-11
Platform version

eddierichter-amd:
Adding a register to the VCK5000 platform to report the platform version. Used for software to determine what platform is programmed on the machine.


platforms/xilinx_vck5000_air/vivado/xilinx_vck5000_air_xsa.tcl
```diff
@@ -64,7 +64,8 @@ if { [string first $scripts_vivado_version $current_vivado_version] == -1 } {
    return 1
 }
 
-set_param board.repoPaths /wrk/xcohdnobkup1/jmelber/vck5000/boards/vck5000/production_silicon/1.1
+#set_param board.repoPaths /wrk/xcohdnobkup1/jmelber/vck5000/boards/vck5000/production_silicon/1.1
+set_param board.repoPaths /group/xrlabs2/eddier/board_files/vck5000/production_silicon/1.1
```
jgmelber, line 68: We should put these under `/group/xrlabs2/platforms/board_files`

eddierichter-amd: Good point, will do.

runtime_lib/controller/main.cpp
```diff
@@ -59,7 +59,7 @@ int shim_dma_cols[NUM_SHIM_DMAS] = {2,  3,  6,  7,  10, 11, 18, 19,
 int col_dma_cols[NUM_COL_DMAS] = {7, 8, 9, 10};
 #define NUM_DMAS (NUM_SHIM_DMAS + NUM_COL_DMAS)
 
-#define CHATTY 0
+#define CHATTY 1
```
jgmelber, line 62: We still want this CHATTY 0

jgmelber:
It would be a good idea to include the ARM firmware build inside the Make process for the vck5000 platform rather than upload the elf to the repository

runtime_lib/controller/main.cpp
jamesroxbypb: Definitely.  CHATTY=1 needs to be an active decision made, not default

jamesroxbypb:
For the "upload elf to the repo" issue, we should just make a separate PR for this, as its a more general issue, rather than related to the version register, which we really want

runtime_lib/controller/main.cpp
eddierichter-amd: Good points. Will change.

platforms/xilinx_vck5000_air/vivado/xilinx_vck5000_air_xsa.tcl
fifield: Why aren't these checked in somewhere?

jgmelber: They are available on the Early Access Site: [https://www.xilinx.com/member/vck5000.html#vivado](https://www.xilinx.com/member/vck5000.html#vivado)

fifield: got it. Related question, what happened to the Xilinx BoardStore? https://github.com/Xilinx/XilinxBoardStore/pull/560 What's XHUB?

jamesroxbypb:
And should test 271 be extended to read this? Seems like each PR should at least have some sort of test

jgmelber:
It should have a test, and the [firmware](runtime_lib/controller/main.cpp) should verify the firmware is compatible (at least some version)

runtime_lib/airhost/host.cpp
```diff
@@ -462,7 +462,11 @@ hsa_status_t air_iterate_agents(hsa_status_t (*callback)(air_agent_t agent,
       return HSA_STATUS_ERROR;
 
     uint64_t *bram_base = reinterpret_cast<uint64_t *>(
-        mmap(NULL, 0x1000, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0));
+        mmap(NULL, 256*1024, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0));
+
+    // Platform version is a constant register after the BRAM
+    uint32_t platform_version = (uint32_t)bram_base[PLATFORM_VERSION_REG / sizeof(uint64_t)];
+    std::cout << "Platform version: " << platform_version << std::endl;
```
fifield, line 469: In general the runtime shouldn't print anything. Can we put this behind a tested C API for now? and maybe an hsa_agent_info query later?

fifield (changes requested):

eddierichter-amd:
> And should test 271 be extended to read this? Seems like each PR should at least have some sort of test

Good points I will add both a get_info query tested in 271 and firmware support. I was thinking that I didn't want it exposed to an application, but that is a nice way of doing it.

runtime_lib/airhost/host.cpp
eddierichter-amd: Good point and inline with Phil's comment. Will do.

runtime_lib/controller/main.cpp
eddierichter-amd: Done in commit 482971b

platforms/xilinx_vck5000_air/vivado/xilinx_vck5000_air_xsa.tcl
eddierichter-amd: Done in commit f448895

runtime_lib/airhost/host.cpp
eddierichter-amd: Printing done in commit b644d1c. Adding to an agent info query and tested in 271 in commit fca3217.

runtime_lib/controller/main.cpp
```diff
@@ -897,7 +897,7 @@ void unlock_uart() {
 int queue_create(uint32_t size, queue_t **queue, uint32_t mb_id) {
   uint64_t queue_address[1] = {base_address + sizeof(dispatch_packet_t)};
   uint64_t queue_base_address[1] = {queue_address[0] +
-                                    sizeof(dispatch_packet_t)};
+                                    2 * sizeof(dispatch_packet_t)};
```
fifield, line 900: This seems like an important fix separate from the platform version. I think we are skipping over the queue data structure here. To avoid future bugs maybe it should be something like:
```C++
ceil((float)sizeof(queue_t) / (float)sizeof(dispatch_packet_t)) * sizeof(dispatch_packet_t)
```
note `queue_base_address` must be aligned to the packet size per hsa spec.

muwyse-amd: what about something like this? It avoids using floating point, at the cost of the loop, but the number of iteration should be small

```
uintptr_t queue_base_address_min = queue_address[0] + sizeof(queue_t);
uintptr_t queue_base_address[1] = {queue_address[0] + sizeof(dispatch_packet_t);}
while (queue_base_address[0] < queue_base_address_min) {
  queue_base_address[0] += sizeof(dispatch_packet_t);
}
```

PR #143, opened 2023-01-18
AIR Runner: Add support for pipeline simulation

erwei-xilinx:
- Tracing channel data consumption and production as symbolic tokens.
- Clang-format.
- Runner unit testing infrastructure.
- Make air-runner support for loops with multiple tokens; use multi-token for loops to represent for-loop pipelining.


erwei-xilinx:
@fifield in this PR I have fixed the issue with linalg op latency estimation.

PR #146, opened 2023-01-26
Add a pattern to remove unnecessary copies after buffer promotion

fifield:
The RemoveAllocCopyLinalgOpCopyPattern pattern in AirLinalgCodegen tries to remove unnecessary memref.copy operations inserted by linalg subview promotion. The copies are to write-only operands of a linalg operation.


jamesroxbypb (approved):
This looks great and makes sense.  Tests look good too, especially as test_copy_remove is the pattern we saw yesterday.

PR #147, opened 2023-01-27
Remove deprecated memcpy variants

fifield:
This is some overdue cleanup. The uses of these were removed from the tests and compiler a long time ago.

### DIFF
--- a/runtime_lib/airhost/memory.cpp
+++ b/runtime_lib/airhost/memory.cpp
@@ -210,199 +210,25 @@ uint64_t air_dev_mem_get_pa(void *buff_va) {
   return offset + (uint64_t)AIR_BBUFF_BASE;
 }
 
-namespace {
-
-int64_t shim_location_data(air_herd_shim_desc_t *sd, int i, int j, int k) {
+static int64_t shim_location_data(air_herd_shim_desc_t *sd, int i, int j,
+                                  int k) {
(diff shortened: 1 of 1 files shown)


runtime_lib/airhost/memory.cpp
```diff
@@ -210,199 +210,25 @@ uint64_t air_dev_mem_get_pa(void *buff_va) {
...
-  }
-}
-
-template<typename T0, int R0, typename T1, int R1>
-void air_mem_cdma_nd_memcpy_queue_impl(tensor_t<T0, R0>* t0, tensor_t<T1, R1>* t1, uint32_t space0, uint32_t space1, 
-                                 uint64_t offset_3, uint64_t offset_2, uint64_t offset_1, uint64_t offset_0,
-                                 uint64_t length_4d, uint64_t length_3d, uint64_t length_2d, uint64_t length_1d,
-                                 uint64_t stride_4d, uint64_t stride_3d, uint64_t stride_2d) {
+static template <typename T0, int R0, typename T1, int R1>
+void air_mem_cdma_nd_memcpy_queue_impl(tensor_t<T0, R0> *t0,
```
jgmelber, line 224: This could be removed. None of our current platforms will really make use of this. It was mainly for L3/L2 transfers emulating MemTiles in URAM.

jgmelber (approved):

Issue #114, opened 2022-12-06
make petalinux sysroot target .PHONY


muwyse-amd:
Add `.PHONY: sysroot` to the petalinux directory Makefile, otherwise the `sysroot` target will always be up-to-date as the target name is the same as the SYSROOT directory name.

This target must execute before the further steps of the platform build occur.

https://github.com/Xilinx/mlir-air/blob/4b3a639f0d84285f07de280cc69b80d0101bee3d/platforms/xilinx_vck190_air/petalinux/Makefile#L33

FredKellerman:
Thank you @muwyse-amd for the fix here!

PR #153, opened 2023-02-01
Update petalinux Makefile with PHONY sysroot

FredKellerman:
This is one way to fix issue #114 and re-enable builds of PYNQ for VCK190 .


fifield:
Thanks!

fifield (approved):

PR #154, opened 2023-02-01
Fix physical device extra enumeration bug

jgmelber:
Skip device enumeration if it is already done


fifield (approved):
OK for a short term fix, but really we need to fix when air_get_physical_devices is called

PR #156, opened 2023-02-02
AIRToAIE Lowering for Channels to ObjectFifos (L1 to L1)

AndraBisca:
This PR adds a new test-pattern to AIRToAIE.cpp called "lower-air-channels".  For now, the lowering is done for data movement from L1 to L1. The pattern should be called after applying air-to-aie test patterns "to-aie-mlir" and "specialize-affine-if", then "canonicalize".

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -869,6 +869,7 @@ struct LowerAIRChannelsPattern : public OpRewritePattern<air::ChannelPutOp> {
 
   LogicalResult matchAndRewrite(air::ChannelPutOp put,
                                 PatternRewriter &rewriter) const override {
+    auto aie_module = put->getParentOfType<ModuleOp>();
     // retrieve put/get of channel
     ChannelOp channel = getChannelDeclarationThroughSymbol(
         dyn_cast<air::ChannelInterface>(put.getOperation()));
@@ -887,17 +888,19 @@ struct LowerAIRChannelsPattern : public OpRewritePattern<air::ChannelPutOp> {
     // find AIE tiles and their cores based on memory hierarchy levels
     AIE::TileOp producerTile;
     AIE::TileOp consumerTile;
+    AIE::CoreOp producerCore;
+    AIE::CoreOp consumerCore;
     if (src_space == (int)air::MemorySpace::L1 &&
         dst_space == (int)air::MemorySpace::L1) {
 
-      AIE::CoreOp producerCore = put->getParentOfType<AIE::CoreOp>();
+      producerCore = put->getParentOfType<AIE::CoreOp>();
       if (!producerCore)
         return failure();
       producerTile = producerCore.getTileOp();
       if (!producerTile)
         return failure();
 
-      AIE::CoreOp consumerCore = get->getParentOfType<AIE::CoreOp>();
+      consumerCore = get->getParentOfType<AIE::CoreOp>();
       if (!consumerCore)
         return failure();
       consumerTile = consumerCore.getTileOp();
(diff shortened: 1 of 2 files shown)


jgmelber (changes requested):
Do you have a unit test and milr test to check in?

AndraBisca:
Added a test in mlir/test/Conversion/AIRToAIE/

fifield:
>  As "to-aie-mlir" seems to delete air.channel operations, they must be manually added back in before applying the channel lowering pattern.

Please check if https://github.com/Xilinx/mlir-air/pull/159 helps with this.

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -845,6 +846,153 @@ void allocL1Buffers(ModuleOp m,
...
+            objFifo, 1);
+    rewriter.setInsertionPointAfter(producerAcq);
+    AIE::ObjectFifoSubviewAccessOp producerAccess =
+        rewriter.create<AIE::ObjectFifoSubviewAccessOp>(
+            rewriter.getUnknownLoc(), elementType, producerAcq.getSubview(),
+            rewriter.getIntegerAttr(rewriter.getI32Type(), 0));
+
+    // replace uses of alloc with result of acquire
+    if (auto a = dyn_cast<memref::AllocOp>(put.getSrc().getDefiningOp()))
+      a.getMemref().replaceAllUsesWith(producerAccess.getOutput());
```
fifield, line 935: Avoid mixing manual IR changes with `rewriter `based changes. The `rewriter` changes are staged and unwound if the pass return `failure()`, but this `replaceAllUsesWith `is immediate.

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -845,6 +846,153 @@ void allocL1Buffers(ModuleOp m,
...
+
+      AIE::CoreOp consumerCore = get->getParentOfType<AIE::CoreOp>();
+      if (!consumerCore)
+        return failure();
+      consumerTile = consumerCore.getTileOp();
+      if (!consumerTile)
+        return failure();
+
+    } else {
+      assert(false && "Unsupported data movement");
```
fifield, line 901: return failure instead of assert?

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -845,6 +846,153 @@ void allocL1Buffers(ModuleOp m,
...
+        rewriter.create<AIE::ObjectFifoReleaseOp>(
+            dealloc->getLoc(), AIE::ObjectFifoPort::Consume, objFifo, 1);
+        dealloc.erase();
+      }
+    }
+
+    // erase the channel and its put/get
+    channel.erase();
+    put.erase();
+    get.erase();
```
fifield, line 980: Use `rewriter.eraseOp()` here and a few places above

mlir/test/Conversion/AIRToAIE/air_channel_to_objectFifo_L1toL1.mlir
```diff
@@ -0,0 +1,79 @@
+//===- air_channel_to_objectFifo_L1toL1.mlir --------------------*- MLIR -*-===//
+//
+// Copyright (C) 2022, Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: air-opt %s -air-to-aie='test-patterns=specialize-affine-if,lower-air-channels' | air-opt --canonicalize | FileCheck %s
```
fifield, line 8: Test could be simplified by using the output of `specialize-affine-if` instead of running it here.

mlir/lib/Conversion/AIRToAIEPass.cpp
AndraBisca: I didn't know this, thank you for explaining! I replaced the function call with one to the rewriter.

mlir/test/Conversion/AIRToAIE/air_channel_to_objectFifo_L1toL1.mlir
AndraBisca: I changed the test so it is the output of `specialize-affine-if` and also `canonicalize`.

jgmelber (approved):

fifield (approved):

PR #158, opened 2023-02-03
Align queue packet array address

jnider:
The packet array must come after the queue header and it must be aligned to the packet size (dispatch_packet_t). Make sure there is enough space for the queue header.

Signed-off-by: Joel Nider <joel.nider@xilinx.com>


jnider:
It seems this has been found & fixed several times but never merged:
https://github.com/muwyse-amd/mlir-air/pull/6/commits/a40ab490317d83c22295c1da72751bc1ed931257
https://github.com/Xilinx/mlir-air/pull/140#discussion_r1072915925

My fix does not use loops or floating point.

fifield (approved):

PR #159, opened 2023-02-03
Preserve channel op when outlining aie cores

fifield:


AndraBisca (approved):
Tested it out on my end and it works. Thank you for the fix!

PR #167, opened 2023-02-09
VCK5000 DDR size patch for latest board files

jgmelber:
Updates the VCK5000 platform build script for the latest board files correctly reflecting the DDR capacity


platforms/xilinx_vck5000_air/vivado/xilinx_vck5000_air_xsa.tcl
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 proc numberOfCPUs {} {
-    return 8
+    return 16
```
keryell, line 6: Perhaps a comment in the code would help understanding the relation to the PR subject.

jgmelber: This is just the number of threads used by Vivado

PR #168, opened 2023-02-10
Load PDI and ELF separately to VCK5000

jnider:
It is useful to be able to load the firmware separately from the PDI. In some cases, the PDI file does not contain the firmware for the command processor (often ARM). The firmware may be provided in a separate ELF file. The bash script loads both the PDI and then the ELF (if it is provided). In other cases such as firmware or driver development, the firmware can be reloaded without resetting the board or disturbing the rest of the system.

The firmware can be reloaded like this:
ELF_FILE=acdc_agent.elf xsct load.tcl

A specific card can be targeted when more than one is installed: CARD_IDX=1 ELF_FILE=acdc_agent.elf xsct load.tcl

All options are passed as environment variables. The shell script sets the environment which is then shared with the TCL script, or environment variables can be passed to the TCL script directly (without using the bash script) as shown in the examples above.

Signed-off-by: Joel Nider <joel.nider@xilinx.com>

### DIFF
--- a/platforms/scripts/load.tcl
+++ b/platforms/scripts/load.tcl
@@ -1,0 +1,108 @@
+# Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+#
+# Load either a PDI file or ELF file (or both) to a VCK5000
+# Intended to be used with xsct, with JTAG over USB
+#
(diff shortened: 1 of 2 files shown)


jgmelber (changes requested):
Thanks for this. Please document the usage and arguments in the script headers

jgmelber (approved):

PR #169, opened 2023-02-10
Qemu

jnider:
Update QEMU options to support Ubuntu and TAP networking, primarily to help during driver development.


jgmelber:
Could you add a README to the scripts directory to orient potential users

jgmelber (approved):
This looks great. README is very helpful

PR #173, opened 2023-02-15
driver: Initial implementation

jnider:
Add a kernel-mode driver for VCK5000. It provides memory-mapped access to the PCIe BARs. This initial implementation is very simple but forms the basis for future features. The device and module are called 'amdair'. Runtimes and low-level apps should use the /dev/amdair character device for communication.

Only a single device is supported at this time. This is planned to be extended in the future.

The VCK5000 currently exposes 3 PCIe BARs (memory ranges). The BARs can be mapped by using the mmap system call from user space. The different BARs can be selected by using an offset, similar to the io_uring driver. Bar 2 (AIE) is deprecated but still in use. To discourage future use, a module parameter enable_aie must be used to gain access to the BAR.

Use GPL license so the string passes some Linux makefile magic that only accepts certain values.

### DIFF
--- a/driver/chardev.c
+++ b/driver/chardev.c
@@ -1,0 +1,293 @@
+/*
+ * Copyright 2023 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER(S) OR AUTHOR(S) BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <stddef.h>
+#include <linux/device.h>
+#include <linux/export.h>
+#include <linux/err.h>
+#include <linux/fs.h>
+#include <linux/file.h>
+#include <linux/sched.h>
+#include <linux/slab.h>
+#include <linux/uaccess.h>
+#include <linux/compat.h>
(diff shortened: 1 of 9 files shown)


jgmelber (changes requested):
Can you add a short README.md covering building, installing, and using the driver?

driver/main.c
```diff
@@ -0,0 +1,169 @@
...
+
+module_init(vck5000_init);
+module_exit(vck5000_exit);
+
+module_param(enable_aie, bool, 0644);
+MODULE_PARM_DESC(enable_aie, "Enable debug access to AIE BAR");
+
+MODULE_LICENSE("GPL");
+MODULE_AUTHOR("Joel Nider <joel.nider@amd.com>");
+MODULE_DESCRIPTION("VCK5000 AIE driver");
```
jgmelber, line 168: VCK5000 AIR driver?

driver/chardev.c
```diff
@@ -40,6 +39,7 @@
 #include <linux/uaccess.h>
 #include <stddef.h>
 #include <uapi/linux/kfd_ioctl.h>
+#include "chardev.h"
```
fifield, line 42: You might need to add a blank line, otherwise clang-format wants alphabetical order.

jgmelber: Done

fifield:
> This clang formatting will cause issues later on, since it's not compatible with the Linux kernel coding standards. I think the easiest path forward is to continue to use clang format so I can keep the driver in this same repo (unless there is an easy way to give an exemption to these files?)

I think there are a few options including a custom configuration for that directory or turning off clang format for portions of the file(s):
https://www.kernel.org/doc/html/next/process/clang-format.html#clang-format

We can also just override the check on the PR but it will continue to cause failed checks on future changes.

jnider:
> > This clang formatting will cause issues later on, since it's not compatible with the Linux kernel coding standards. I think the easiest path forward is to continue to use clang format so I can keep the driver in this same repo (unless there is an easy way to give an exemption to these files?)
> 
> I think there are a few options including a custom configuration for that directory or turning off clang format for portions of the file(s): https://www.kernel.org/doc/html/next/process/clang-format.html#clang-format

That's great - I didn't even know Linux used clang-format. I have stolen their .clang-format file and stuck it in the driver directory. The license should be ok since its an unmodified redistribution of an open-source file. Hopefully git CI will honor that file. I will push a new version shortly.

driver/main.c
jnider: updated

fifield (approved):

jgmelber (approved):

driver/main.c
```diff
@@ -0,0 +1,179 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE COPYRIGHT HOLDER(S) OR AUTHOR(S) BE LIABLE FOR ANY CLAIM, DAMAGES OR
+ * OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
```
stephenneuendorffer, line 21: This is not a GPL license. can you update the license text in the file and also note in the toplevel license that the repo includes code that is GPL, please?

PR #174, opened 2023-02-16
Python binding exposed read32/write32 for debug

jgmelber:


jamesroxbypb:
I like this a lot.  I will also like removing this in the future when we have smarter wrapping for general debug/profile steps

fifield (changes requested):
Looks good. It should have a simple test of the C API in the unit test area.

jgmelber:
> Looks good. It should have a simple test of the C API in the unit test area.

Board-level test? or python/test?

fifield:
> > Looks good. It should have a simple test of the C API in the unit test area.
> 
> Board-level test? or python/test?

Ideally both but at least a "board-level" test since that's the only place we currently test the C APIs. We have an open issue on the general lack of testing of runtime python bindings.

fifield (approved):

jamesroxbypb (approved):
Looks good

PR #176, opened 2023-02-17
Update setup_python_packages.sh

jamesroxbypb:
We should use the Python3 built-in venv rather than required the optional virtualenv be installed.  Last machine I tried to use failed instantly due to the missing module


fifield (approved):
+1

PR #180, opened 2023-02-22
Channel to ObjectFifo Lowering L1toL3

abisca:
This PR changes the channel-to-objectFifo rewrite pattern in AIRToAIE.cpp to target ChannelOps directly. With the addition of two new helper functions, the channel symbol is used to search for both the associated ChannelPut and ChannelGet. If either is missing, then the communication is done through L3. As objectFifos require physically placed tiles, for now, a simple ShimTileAllocator is used to instantiate new shim tiles when required.

One small change to AIR.td ChannelPut and ChannelGet ops is made such that code for the rewrite pattern can be written more cleanly: the getSrcMemref() and getDstMemref() are both changed to getMemref(). This change was applied to any source files currently using those functions.

For now, the channel-to-objectFifo lowering only supports lowering for one-to-one channels. Subsequent PRs will extend the lowering to support lowering for many-to-many channels without broadcast and one-to-many with broadcast.


erwei-xilinx:
If the other channel put/get is missing then it is assumed to be to/from L3. May I ask if it could also come from L2?

abisca:
After the "to-mlir-aie" part of the AIRToAIE lowering pass, anything related to the L3 memref is removed because the memref itself is removed. Since L2 will be introduced with the addition of MemTiles I think the allocs associated with L2 memory will not be removed so neither will the channel put/get. That's my understanding of it at least, but maybe @fifield could confirm?

fifield:
@abisca , yes one way to do L2 allocation is with a static buffer allocation in a MemTile the same as we do for L1 (tile) memory.

erwei-xilinx:
I see. Thanks for the explanation. Looks great to me.

erwei-xilinx (approved):

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -51,6 +51,20 @@ struct AIRToAIEOptions {
...
+// std::vector<int> s80_nmu_col_list{0, 0, 1, 1, 0, 0, 1, 1,
+//                                   0, 0, 1, 1, 0, 0, 0, 0,
+//                                   0, 0, 1, 1, 0, 0, 0, 0,
+//                                   0, 0, 1, 1, 0, 0, 0, 0,
+//                                   0, 0, 1, 1, 0, 0, 0, 0,
+//                                   0, 0, 1, 1, 0, 0, 1, 1,
+//                                   0, 0};
+std::vector<int> shim_dma_cols{2,  3,  6,  7,  10, 11, 18, 19,
+                               26, 27, 34, 35, 42, 43, 46, 47};
+const int shim_dma_channels = 2;
```
jgmelber, line 66: Not necessarily for this PR but it would be a good idea to start factoring device/generation specific information out of passes

jgmelber (approved):

mlir/include/air/Util/Util.h
```diff
@@ -60,6 +60,10 @@ DmaMemcpyNdOp getAIRDmaInBlock(mlir::Block *block);
 
 // Get channel declaration through channel symbol
 ChannelOp getChannelDeclarationThroughSymbol(ChannelInterface op);
+// Get ChannelPutOp from ChannelOp
+ChannelPutOp getChannelPutOpThroughSymbol(ChannelOp channel);
+// Get ChannelGetOp from ChannelOp
+ChannelGetOp getChannelGetOpThroughSymbol(ChannelOp channel);
```
fifield, line 66: There will be multiple gets or puts in the broadcast case. Can we make this API return multiple ops from the start?

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -78,6 +92,41 @@ AIE::TileOp getPhysTileOp(ModuleOp aie_module, int col, int row) {
...
+
+  ShimTileAllocator(std::vector<int> cols, int channels)
+      : shim_columns(cols), shim_dma_channels(channels) {}
+
+  AIE::TileOp getShimTile(ModuleOp aie_module, int src_memory_space,
+                          int dst_memory_space) {
+    bool isMM2S = (src_memory_space < dst_memory_space);
+    auto allocs = isMM2S ? &mm2s_allocs : &s2mm_allocs;
+
+    // return first available shim tile with a free channel
```
fifield, line 115: I know this is what the current allocator does, but it's not safe in general to just return the next channel.

mlir/include/air/Util/Util.h
erwei-xilinx: Yes I agree--this is something that I've been working on. Will make a PR on this soon.

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -862,134 +911,164 @@ AIE::ObjectFifoCreateOp createObjectFifo(OpBuilder &builder,
...
-
-      producerCore = put->getParentOfType<AIE::CoreOp>();
-      if (!producerCore)
-        return failure();
-      producerTile = producerCore.getTileOp();
-      if (!producerTile)
-        return failure();
+      getExists = false;
+
+    if (putExists || getExists) {
```
fifield, line 970: Handle multiple gets/puts and the channel indices or fail gracefully.

fifield (approved):
We need to handle the broadcast case before this is enabled by default.

mlir/lib/Conversion/AIRToAIEPass.cpp
fifield: Anything related to L2 can be deleted now. The shim tile information is needed until we get some kind of logical tile in aie dialect.

abisca: There's two L2 tests for AIRToAIE in `mlir/test/Conversion/AIRToAIE/`, air_aie_to_L2cpy.mlir and air_L2cpy_to_aie.mlir, should I remove those as well or disable them with a 'requires' statement?

jgmelber: They can be removed

abisca: Yes that would be better, I'll adjust that. Thank you!

abisca:
In the interest of keeping the PR more focused on a specific change, I will first add support for the lowering of only one-to-one channels with this PR, and expand the lowering to channels with more interesting indices in new PRs.

PR #182, opened 2023-02-22
Updates for cmake changes

fifield:
Make the github action build the runtime again after [recent cmake changes](https://github.com/Xilinx/mlir-air/pull/122).

Some notes:
* This moves `BUILD_AIR_PCIE` definition to the toolchain file and makes it the default for x86.
* Use the x86 toolchain file in the github builder
* Make clang-12 the x86 default to match github builder
* Don't hard-code `LibXAIE_ROOT`

None of the utility scripts are updated to properly build the runtime after https://github.com/Xilinx/mlir-air/pull/122, but this is a start.


fifield:
The builder logs now contain,
```
Compiling AIR runtime for x86
TOOLCHAIN=/home/runner/work/mlir-air/mlir-air/build_assert/../cmake/modules/toolchain_x86.cmake
```
and,
```
-- Found LibXAIE: /home/runner/work/mlir-air/mlir-air/aienginev2/install/include  
Building PCIe runtime (BUILD_AIR_PCIE=ON)
```

runtime_lib/CMakeLists.txt
```diff
@@ -30,8 +30,6 @@ project(air_runtime_lib LANGUAGES CXX C)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
 
-# Eventually we need to have some configuration options to select LibXAIEV2
-set(LibXAIE_ROOT ${Sysroot}/usr ${Sysroot}/opt/xaiengine ${Sysroot}/opt/xaienginev2)
```
stephenneuendorffer, line 34: Where do you think this search path should live?

stephenneuendorffer (approved):

runtime_lib/CMakeLists.txt
fifield: There are some defaults here:
https://github.com/Xilinx/cmakeModules/blob/master/FindLibXAIE.cmake#L36
Or it's given to cmake by the user.

jgmelber:
So the following can be removed from the `build-mlir-air-pcie.sh` script? 

```
    -DVitisSysroot="" \
    -DLibXAIE_DIR=${LibXAIE_DIR} \
    -DARM_TOOLCHAIN_OPT="" \
    -DBUILD_AIR_PCIE=ON \
```

PR #184, opened 2023-02-23
Removed L2 mentions in AIRToAIE.

abisca:


runtime_lib/airhost/memory.cpp
```diff
@@ -485,43 +391,3 @@ mlir_air_dma_nd_memcpy(3d1f32, 3, 1, float);
 mlir_air_dma_nd_memcpy(4d1f32, 4, 1, float);
 
 } // extern "C"
-
-#define mlir_air_nd_memcpy(mangle, rank0, space0, type0, rank1, space1, type1) \
```
fifield, line 489: This is the "full memcopy" API. It is emitted by the compiler when the runtime is responsible for both sides of the copy (i.e. it is not a half-dma). Lets leave the macro and function instantiations and only remove the call to `air_mem_cdma_nd_memcpy_queue_impl`

fifield (changes requested):

fifield (approved):

Issue #30, opened 2022-10-24
aircc can reference wrong version of llvm tools.


fifield:
aircc.py assumes that the clang, opt, et. al. in the user's PATH are the correct ones to use.
There is also a hard coded assumption for the path of mlir-aie: https://github.com/Xilinx/mlir-air/blob/main/python/air/compiler/aircc/main.py#L356

PR #185, opened 2023-02-23
Adding air_init() to unit tests

eddierichter-amd:


test/02_mb_dispatch/Makefile
```diff
@@ -0,0 +1,30 @@
+# Copyright (C) 2022, Xilinx Inc.
```
fifield, line 1: This checks in a bunch of makefiles but the tests are built and run via llvm-lit. We could have makefiles if we refactored the lit infrastructure to build via makefiles. Otherwise the makefiles are untested and get out of sync and we shouldn't have them.

python/air/compiler/aircc/main.py
```diff
@@ -358,6 +358,7 @@ def run_flow(opts):
       cmd += [f'-I{thispath}/../../../../runtime_lib/airhost/include']
       cmd += [f'-I{thispath}/../../../../runtime_lib']
       cmd += [f'-I{thispath}/../../../../../aie/runtime_lib']
+      cmd += [f'-I{thispath}/../../../../../utils/mlir-aie/runtime_lib']
```
fifield, line 361: This just adds a second hard coded path to mlir-aie? I'd rather fix this properly (see https://github.com/Xilinx/mlir-air/issues/30)

fifield (changes requested):
The makefiles should not be checked in unless they are tested.

test/02_mb_dispatch/Makefile
eddierichter-amd: Didn't realize that. I can delete them, some of the tests like 13 and 21 do have Makefiles idk if we want to do anything with those.

python/air/compiler/aircc/main.py
fifield: If mlir-aie and mlir-air are installed to the same place then line 359 should be sufficient. i.e. they will share a runtime_lib install directory.

eddierichter-amd: I don't believe the building documentation states to install them in the same directory. I was thinking of just adding some logic similar to the Makefile (https://github.com/Xilinx/mlir-air/blob/main/test/13_mb_add_one/Makefile#L5) to find where mlir-aie was installed. Thoughts?

fifield: A few thoughts:
 - Put this in for now but maybe we should update the documentation to not suggest such a strange install location for mlir-aie? This assumes that aircc is installed relative to the mlir-air source directory which in turn has mlir-aie installed inside it.
 - The relative path might be different for the lit tests because testing runs out of the build directory not an install (because test comes before install). I think line 360 even assumes a certain test setup.
 - Finding the path with `which aie-opt` is an option and is effectively how we find `aie-opt, aie-translate,` and `aiecc.py` But cmake also knows the location of the mlir-aie it's building against and can set the path at compile time.

test/02_mb_dispatch/Makefile
eddierichter-amd: Removed all added Makefiles in [e2cdfad](https://github.com/Xilinx/mlir-air/pull/185/commits/e2cdfade011aaa08ef6d4ac2765de032d8a40f03).

fifield (approved):

PR #189, opened 2023-03-03
Add channel unit-tests from jgmelber's mlir-air fork.

abisca:


fifield (changes requested):
These are not very useful or maintainable in their current form. They need to be turned into lit + FileCheck tests. If they are for testing MLIR passes then they need to be sorted into `mlir/test`. If they are end-to-end tests for running in system then they need host code.

fifield:
This very out of date, with no recent activity.  Closing.

PR #195, opened 2023-03-09
AIR Runner: Refactor codebase for new resource model backend

erwei-xilinx:
This is a rewrite of air-runner simulator's resource model, with the introduction of a graph of objects representing different resources. Resource hierarchies are special resource objects which represent coarse-grained resource models, including device, column and tile. Current plan is to have air.launch reserve resources on device, air.segment reserve entire columns, and air.herd reserve entire tiles.

- Incremental code development ensuring existing runner tests pass with the new resoruce backend.
- Resource hierarchy contention modelling. Launch will fail if not enough resources are allocated to it.
- Resource contention, including memories and ports.


mlir/include/air/Util/Util.h
```diff
@@ -38,6 +38,13 @@ scf::ParallelOp getParallelRegionInitValsOwner(Operation *op, Value val);
 HerdOp getHerdArgOwner(Value val);
 // Get the parent air.hierarchy op of a tile id
 HierarchyInterface getHierarchyArgOwner(Value val);
+// Get the scf parent op from scf.yield op
+template <typename T> Operation *getScfParentOpFromYieldOp(scf::YieldOp op) {
+  if (auto scfop = dyn_cast<T>(op->getParentOp())) {
+    return scfop.getOperation();
+  }
+  return nullptr;
+}
```
fifield, line 47: If you make this return type `T` then the user doesn't need to cast the returned value. If it also took an `Operation*` instead of a `scf::YieldOp` it would be general and a new form of `Operation::getParentOfType`.  But if it were used more generally then `dyn_cast_if_present` should be used because `getParentOp` can return `nullptr`

erwei-xilinx: Thanks for the suggestions, Jeff! I've modified this function accordingly.

PR #201, opened 2023-03-14
Update README.md

jnider:
Fix the broken links.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,9 +27,9 @@ air
 
 ### Getting Started
 
-- [Building](building.html)
-- [Running Examples](running.html)
-- [AIRCC](aircc.html)
+- [Building](building.md)
+- [Running Examples](running.md)
+- [AIRCC](aircc.md)
 
 ### Generated MLIR Documentation
 - [AIR Dialect](AIRDialect.html)
@@ -41,12 +41,12 @@ air
 ### AIR Platform Documentation
 #### [VCK190 platform](platforms/xilinx_vck190_air)
 #### [VCK5000 platform](platforms/xilinx_vck5000_air)
-#### [Building MicroBlaze firmware](vck190_microblaze_firmware.html)
+#### [Building MicroBlaze firmware](vck190_microblaze_firmware.md)
 
 ### Tutorials
-- [AIR Asynchronous Concurrency](AIRAsyncConcurrency.html)
-- [AIR Runner](AIRRunner.html)
+- [AIR Asynchronous Concurrency](AIRAsyncConcurrency.md)
+- [AIR Runner](AIRRunner.md)
 
 -----
 
-<p align="center">Copyright&copy; 2019-2022 Advanced Micro Devices, Inc.</p>
+<p align="center">Copyright&copy; 2019-2023 Advanced Micro Devices, Inc.</p>


fifield (changes requested):
The links are setup to work here:
https://xilinx.github.io/mlir-air/

There might be a solution to make it work on both github.io and browsing the repo but I don't know it.

stephenneuendorffer:
I think the .md links with relative paths are transformed to .html links with relative paths when published to github.io: https://github.blog/2016-12-05-relative-links-for-github-pages/

fifield:
The generated mlir docs will never work browsing the repo unless we check them in but the platform docs can probably be reworked to work in both places. Here's the code that deploys them: https://github.com/Xilinx/mlir-air/blob/main/.github/workflows/generateDocs.yml#L93-L99 if anyone wants to play with it.

jnider:
I see, this seems to be a bigger issue than I thought. According to the explanation in Steve's link, the correct form is to write:
`[a relative link](another-page.md)`
Which gets translated to the appropriate HTML page automatically when published.
Is there a way to test github.io without checking in this commit?

stephenneuendorffer:
> The generated mlir docs will never work browsing the repo unless we check them in but the platform docs can probably be reworked to work in both places. Here's the code that deploys them: https://github.com/Xilinx/mlir-air/blob/main/.github/workflows/generateDocs.yml#L93-L99 if anyone wants to play with it.

Maybe use this instead? https://github.com/actions/deploy-pages

stephenneuendorffer:
> I see, this seems to be a bigger issue than I thought. According to the explanation in Steve's link, the correct form is to write: `[a relative link](another-page.md)` Which gets translated to the appropriate HTML page automatically when published. Is there a way to test github.io without merging this commit?

I don't think so, but I'd support experimenting here for a bit. We can always revert the changes.

jackl-xilinx:
@jnider You can experiment on your own fork and it'll build an associated github.io for it.

jnider:
I tested it on my own fork as @jackl-xilinx suggested and it works, according to the formatting in the link posted by @stephenneuendorffer .
You can see the .md files browsing the source tree here: https://github.com/jnider/mlir-air/tree/main/docs
And the .html files deployed on github.io (redirected to my personal domain name) here: http://joel.nider.org/mlir-air/

So at least those links can be updated but it doesn't solve the problem for the generated mlir docs.

docs/README.md
```diff
@@ -27,26 +27,26 @@ air
...
-- [AIR Dialect](AIRDialect.html)
-- [AIRRt Dialect](AIRRtDialect.html)
-- [AIR Transform Passes](AIRTransformPasses.html)
-- [AIR Conversion Passes](AIRConversionPasses.html)
-- [Transform Dialect Extensions](AIRTransformOps.html)
+- [AIR Dialect]
+- [AIRRt Dialect]
+- [AIR Transform Passes]
+- [AIR Conversion Passes]
+- [Transform Dialect Extensions]
```
fifield, line 39: Revert this

docs/README.md
```diff
@@ -27,26 +27,26 @@ air
...
 
 ### AIR Platform Documentation
-#### [VCK190 platform](platforms/xilinx_vck190_air)
-#### [VCK5000 platform](platforms/xilinx_vck5000_air)
-#### [Building MicroBlaze firmware](vck190_microblaze_firmware.html)
+#### [VCK190 platform](../platforms/xilinx_vck190_air)
+#### [VCK5000 platform](../platforms/xilinx_vck5000_air)
```
fifield, line 43: Revert this

fifield (approved):

docs/README.md
fifield: The platform docs links will be broken after navigating to https://github.com/Xilinx/mlir-air/blob/main/docs/building.md directly, but when I set it up I thought it was nicer to have the platform specific docs be the `README.md` when navigating to `https://github.com/Xilinx/mlir-air/tree/main/platforms/<platform>`.  We could instead move the platform `README.md`s to the docs folder. Or maybe there are other suggestions?

PR #202, opened 2023-03-14
Update building.md

jnider:
Add more details to install prerequisites on fresh Ubuntu 20.04 install

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -11,7 +11,24 @@
   ninja 1.10.0
   Xilinx Vitis 2021.2 
 ```
+### Regular packages
+```
+sudo apt install ninja-build clang-10 clang lld-10 lld
+```
+### cmake
+Ubuntu 20.04 is released with:
+cmake/focal,now 3.16.3-1ubuntu1 amd64 [installed]
 
+But we require:
+3.20.6 (or newer)
+```
+wget https://cmake.org/files/v3.20/cmake-3.20.6-linux-x86_64.sh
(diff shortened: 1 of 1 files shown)


docs/building.md
```diff
@@ -11,7 +11,24 @@
   ninja 1.10.0
   Xilinx Vitis 2021.2 
 ```
+### Regular packages
+```
+sudo apt install ninja-build clang-10 clang lld-10 lld
```
fifield, line 16: We test (and clang-format) with clang-12. Can we put that here instead?

docs/building.md
```diff
@@ -11,7 +11,24 @@
...
+Ubuntu 20.04 is released with:
+cmake/focal,now 3.16.3-1ubuntu1 amd64 [installed]
+
+But we require:
+3.20.6 (or newer)
+```
+wget https://cmake.org/files/v3.20/cmake-3.20.6-linux-x86_64.sh
+chmod +x cmake-3.20.6-linux-x86_64.sh
+sudo ./cmake-3.20.6-linux-x86_64.sh --prefix=/usr/local
+cmake --version
```
fifield, line 28: I usually just pip install cmake with any other requirements. It's easier than building it and doesn't require any special permissions or `PATH`.

fifield (changes requested):

docs/building.md
jnider: That's fine - would you like to update the instructions with your method?

PR #205, opened 2023-03-22
ChannelToObjFifo - simple pingpong support

abisca:
Added integer attribute to ChannelOp to specify number of buffer resources that can be used by the channel-to-objfifo lowering.


jgmelber (approved):
Thanks! This is a useful initial hook for enabling ping-pong channels through one mechanism

fifield (approved):

PR #206, opened 2023-03-23
Jnider runtime

jnider:
Change the runtime_lib to use the driver interface rather than /dev/mem.

This has only been tested with test #13.


eddierichter-amd (commented):
Looks good to me.

PR #213, opened 2023-04-06
Channel lowering: channels with indices and broadcast

abisca:
This PR modifies the channel-to-objectFifo rewrite pattern in AIRToAIE.cpp to account for channel bundles, i.e. channels with a size attribute higher than [1, 1]. This is done in such a way that it will work for any size channels, not just with x and y dimensions, by using an std::vector<int> of indices, that represents the indices from channel put/get operations that refer to a specific channel of the bundle.

For each channel, all possible channel indices are computed based on the channel size attribute. Then, puts/gets using that channel are identified. For any indices of the channel that do not have a corresponding put/get operation they will be treated as accesses to/from external memory. Finally, objectFifos are created with the acquired information.

Additionally, this PR introduces a lowering for channels with a broadcast_shape attribute to objectFifos with one producer and multiple consumers.


mlir/test/Conversion/AIRToAIE/air_channel_to_objectFifo_L1toL1.mlir
```diff
@@ -10,17 +10,17 @@
 // CHECK: module @aie.segment_0 {
 // CHECK:   %0 = AIE.tile(1, 1)
 // CHECK:   %1 = AIE.tile(1, 2)
-// CHECK:   %2 = AIE.objectFifo.createObjectFifo(%0, {%1}, 1) : !AIE.objectFifo<memref<32xi32, 2>>
```
erwei-xilinx, line 13: Is the memory space label in memref removed inside aie.segment because we assume that L1 is the only memory hierarchy that can possibly appear here?

AndraBisca: In this code example the aie.segment is the label of the aie.module. I can see how this is confusing, but in mlir-aie there is no segment operation like in mlir-air. 

Regarding the memory space labels, for now the objectFifo abstraction only allocates memory in L1. It can be made aware of memory in L3 through the registration of aie.external_buffers but the lowering will not allocate those itself. This is, however, subject to change with the addition of MemTiles (L2) in the future.

abisca:
Features now introduced in #263

PR #217, opened 2023-04-10
Compile different runtime targets to different directories

stephenneuendorffer:
This allows us to have multiple runtime targets in a useful way.


stephenneuendorffer:
Users of the location of this library still need to be updated.

PR #221, opened 2023-04-20
AIR Async Dependency: Pipeline lowering

erwei-xilinx:


mlir/lib/Util/Util.cpp
```diff
@@ -522,4 +521,62 @@ Operation *air::getAffineIfNestAndSpatialLoopFromOp(
...
+char air::checkOpOperandReadOrWrite(mlir::OpOperand &op_operand) {
+  auto owner = op_operand.getOwner();
+  // If used in DmaMemcpy Op
+  if (auto dma = dyn_cast<xilinx::air::DmaMemcpyInterface>(owner)) {
+    if (op_operand.is(dma.getSrcMemref())) {
+      return 'r';
+    } else if (op_operand.is(dma.getDstMemref())) {
+      return 'w';
+    } else {
+      assert(false && "Unknown operand in air.dma");
```
keryell, line 545: This is not specific to this PR as it appears all over the place.
But I am curious about what is the expected plan for this when not compiling with `assert` on? To go on? Because there is no test in the `assert` so it it is obvious it will goes badly later. :-)
Would it be clearer to have a simple `std::terminate`? Or throwing an exception just to capture the error message even if it is compiled with `-fnoexcept` (I am unsure whether it can work).

mlir/test/Transform/AIRDependencyScheduleOpt/hoist_alloc.mlir
```diff
@@ -0,0 +1,47 @@
...
+          }
+          air.herd_terminator
+        }
+        air.segment_terminator
+      }
+      air.launch_terminator
+    }
+    return
+  }
+}
```
keryell, line 47: Perhaps you could update your editor configuration to add the EOL at the EOF?

mlir/include/air/Util/Util.h
```diff
@@ -100,6 +100,10 @@ struct LinalgTransforms {
   static const StringLiteral kLinalgTransformMarker;
 };
 
+// Check if an operand of an operation is read or write access
+char checkOpOperandReadOrWrite(mlir::OpOperand &op_operand);
+char checkOpOperandReadOrWrite(Value op_operand, Operation * owner);
```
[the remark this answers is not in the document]
erwei-xilinx: Thanks for the comment. Later I will apply clang-format tool to the codebase, so hopefully issues like this will go away.

mlir/lib/Util/Util.cpp
erwei-xilinx: I agree with you. Currently I used assertion here in the draft for debugging, but I will later review the codebase and use "emitError" MLIR diagnostic to replace those assertions before I push. Thanks for your comment.

mlir/test/Transform/AIRDependencyScheduleOpt/hoist_alloc.mlir
erwei-xilinx: I will later apply clang-format tool to the codebase when the draft is ready. That tool will handle the EOF properly.

PR #222, opened 2023-04-20
Jnider sysfs

jnider:
Add sysfs interface to replace direct access to AIE BAR.
Depends on changes to libxaie (https://github.com/jnider/embeddedsw/tree/vck5000)

Updated the license to GPLv2 and made a note in the top-level license, as requested.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,5 @@ air_project
 
 # build results
 libadf.a
+*.elf
+*.elf.*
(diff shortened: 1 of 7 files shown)


stephenneuendorffer:
Quick scan: LGTM

jnider:
@fifield I would like to get your input on this change, specifically regarding the 'embeddedsw' repo. I had to make changes to support the new backend but can't commit these changes upstream for several reasons. The result is that I have to pull from a fork for the tests to work and the CI build to pass.

fifield (approved):

jnider:
> > 
> 
> Makes sense. At some point we'll want to update to a newer version of the driver, e.g. https://github.com/Xilinx/aie-rt/tree/xlnx_rel_v2022.2

I agree - I started to look at it but it will take some work

PR #226, opened 2023-05-04
driver: revert license to MIT

jnider:
This partially reverts commit dc50c477.

It is easier to develop with the wider team if the driver uses the MIT license.

The module declares a dual license so we can continue to use GPL-only functions such as 'device_destroy'. This follows the same procedure as the AMDKFD driver.


atgutier (approved):
Thanks!

PR #228, opened 2023-05-08
Add a channel get/put hoisting pass

fifield:
Some experiments hoisting `air.channel.get/put` out of loops by adding an extra dimension to the op.

In addition this adds some methods to the `air::ChannelInterface` interface

### DIFF
--- a/mlir/lib/Transform/AIRMiscPasses.cpp
+++ b/mlir/lib/Transform/AIRMiscPasses.cpp
@@ -15,6 +15,8 @@
 #include "air/Transform/AIRMiscPasses.h"
 #include "PassDetail.h"
 #include "air/Dialect/AIR/AIRDialect.h"
+#include "air/Dialect/AIRRt/AIRRtDialect.h"
+#include "air/Dialect/AIRRt/AIRRtOps.h"
 #include "air/Transform/AIRTilingUtils.h"
 #include "air/Util/Dependency.h"
 #include "air/Util/Util.h"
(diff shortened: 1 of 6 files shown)


mlir/lib/Transform/AIRMiscPasses.cpp
```diff
@@ -879,6 +881,119 @@ void AIRLowerHerdParallelPass::runOnOperation() {
...
+
+    // channel op must be the only op in the loop
+    if (forOp.getBody()->getOperations().size() != 2)
+      return failure();
+
+    // scf.for must have constant bounds and step
+    auto lb = forOp.getLowerBound().getDefiningOp<arith::ConstantIndexOp>();
+    auto ub = forOp.getUpperBound().getDefiningOp<arith::ConstantIndexOp>();
+    auto step = forOp.getStep().getDefiningOp<arith::ConstantIndexOp>();
+    if (!lb || !ub || !step)
```
keryell, line 909:
```suggestion
    if (!(lb && ub && step))
```
might be easier to understand the precondition you expect.

mlir/lib/Transform/AIRMiscPasses.cpp
```diff
@@ -879,6 +881,119 @@ void AIRLowerHerdParallelPass::runOnOperation() {
...
+          offsets, sizes, strides);
+    else
+      rewriter.create<air::ChannelGetOp>(
+          loc, tys, deps, channel, chanOp.getIndices(), chanOp.getMemref(),
+          offsets, sizes, strides);
+    rewriter.eraseOp(op);
+    rewriter.eraseOp(forOp);
+    return success();
+  }
+};
```
keryell, line 974: Ouch, 85 lines of mostly uncommented code... Feel free to help some eventual readers. :-)

mlir/lib/Transform/AIRMiscPasses.cpp
fifield: I prefer the one I can parse left-to-right without a mental stack.

keryell: I did not think you were a Turing machine! ;-)

PR #230, opened 2023-05-11
Make channel name the objfifo name

fifield:
Set the symbol name of object fifos generated in air-to-aie to `"air_<channel_name>"`

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -899,11 +899,14 @@ void allocL1Buffers(ModuleOp m,
   (void)applyPatternsAndFoldGreedily(m, std::move(patterns));
 }
 
-AIE::ObjectFifoCreateOp
-createObjectFifo(OpBuilder &builder, AIE::AIEObjectFifoType datatype,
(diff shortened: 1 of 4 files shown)


mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -899,11 +899,14 @@ void allocL1Buffers(ModuleOp m,
   (void)applyPatternsAndFoldGreedily(m, std::move(patterns));
 }
 
-AIE::ObjectFifoCreateOp
-createObjectFifo(OpBuilder &builder, AIE::AIEObjectFifoType datatype,
-                 Value prodTile, std::vector<Value> consTile, int depth) {
+AIE::ObjectFifoCreateOp createObjectFifo(OpBuilder &builder,
+                                         AIE::AIEObjectFifoType datatype,
+                                         Value prodTile,
+                                         std::vector<Value> consTile, int depth,
```
keryell, line 905:
```suggestion
                                         cont std::vector<Value> &consTile, int depth,
```
might be enough?

fifield: fixed in https://github.com/Xilinx/mlir-air/pull/231

PR #224, opened 2023-05-01
Update mlir-aie

fifield:
The start of updates for `AIE.device`, `AIETargetModel` and AIE2.

* update mlir-aie to da8a0b3
* Generate `AIE.device` op during `air-to-aie`
* Add device option to `air-to-aie`
* Generate AIE2 locks in `air-to-aie` dma lowering

PR #231, opened 2023-05-11
Enable channel to object fifo lowering

fifield:
* Enable channel to object fifo conversion in air-to-aie
* Disallow mixing dma copies and channel get/put


fifield:
build is failing because it assumes https://github.com/Xilinx/mlir-air/pull/224

abisca (approved):

PR #232, opened 2023-05-18
AIR Runner: Replace assertion with mlir diagnostic or custom error mechanism

erwei-xilinx:


keryell (approved):
Definitely a better end-user experience.

PR #234, opened 2023-05-19
AIR Runner: Improvements and bugfixes on resource backend

erwei-xilinx:
- Differentiate the format for compute capacity per core between ops/cycle and macs/cycle.
- Resource allocation and deallocation will make some previously inexecutable ops on wavefront executable. Therefore, simulation needs to recurse to check for these knock-on effects.
- The ordering of ops on wavefront is important. For example, yield ops shall reset the execution status of other ops which are ready to execute at the same timestamp.
- Detected a corner case where the runner nodes fail to cover enough vertices as candidate events to be pushed to wavefront. Current solution is to keep a temporary buffer of neighbouring vertices to recently resetted vertices, which are sometimes victims of such corner cases.
- Given that in our async CDFG dependence to scf.for goes out from scf.yield, and that scf.yield is non-blocking, a air.wait_all barrier is inserted after for loops to protect the scheduling behaviour.

### DIFF
--- a/mlir/lib/Util/Runner.cpp
+++ b/mlir/lib/Util/Runner.cpp
@@ -604,8 +604,8 @@ private:
     std::string cpuops = "math.rsqrt;";
     cpuops += "arith.mulf;arith.divf;arith.addf;arith.subf;arith.truncf;"
               "arith.cmpf;arith.maxf;";
-    cpuops += "arith.muli;arith.divi;arith.addi;arith.subi;arith.trunci;"
-              "arith.cmpi;arith.maxi";
+    cpuops += "arith.muli;arith.divsi;arith.divsi;arith.addi;arith.subi;"
+              "arith.trunci;arith.cmpi;arith.maxi";
     cpuops += "std.select";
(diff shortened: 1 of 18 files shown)


mlir/lib/Util/Runner.cpp
```diff
@@ -236,14 +236,36 @@ class AIRRunner::AIRRunner_impl {
...
                     uint64_t time) {
 
+    LLVM_DEBUG(llvm::dbgs() << "\nNEW TIME STAMP @" << time - 1 << " runner "
+                            << air::to_string(c.ctrl_g->hierarchyOp) << " loc "
+                            << air::to_string(c.ctrl_g->position) << "'\n");
+
+    executeOpsFromWavefrontAndFreeResource(c, device_resource_node, time);
+    pushOpsToWavefrontAndAllocateResource(c, device_resource_node, time);
+
+    return c.wavefront.size() > 0;
```
keryell, line 249: Perhaps this API does exist?
```suggestion
    return !c.wavefront.empty();
```

PR #235, opened 2023-05-19
Add libxaie paths to aiecc command lines

fifield:
The defaults assumed by aiecc.py do not work for these tests. The issue is that aiecc sets the libxaie paths assuming libxaie was built as a subproject of mlir-aie, which is not always true.

### DIFF
--- a/test/03_mb_lock_rel/run.lit
+++ b/test/03_mb_lock_rel/run.lit
@@ -6,5 +6,5 @@
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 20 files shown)


stephenneuendorffer (changes requested):
I don't understand, the libxaie dependency should be managed by aiecc.  This isn't the way to fix this, IMO

fifield:
> I don't understand, the libxaie dependency should be managed by aiecc. This isn't the way to fix this, IMO

I agree. These changes are to make the tests pass on my vck190. I'll gladly not merge them if we can fix aiecc instead.

fifield:
tracking this issue here: https://github.com/Xilinx/mlir-aie/issues/501

fifield:
This is likely fixed by https://github.com/Xilinx/mlir-air/pull/246

fifield:
obsolete

PR #240, opened 2023-05-30
Keep output of air-to-aie in one module.

fifield:
The goal is to keep more of the lowering unified in one module (file) for longer. A new pass -air-split-devices can split out just the device code for input to aiecc.


fifield:
@erwei-xilinx FYI this PR adds a scope operand to the `getChannelPutOpThroughSymbol` `getChannelGetOpThroughSymbol` API to limit the search scope. The default is to use the enclosing `ModuleOp` which is the previous behavior.

PR #242, opened 2023-06-07
Runtime library support for channel put and get operations

zzzDavid:
## Summary
This PR adds support for `air.channel.put` and `air.channel.get` operations in the `aircpu` runtime library for functional simulation. 

## Changes
- New file: `runtime_lib/aircpu/channel.cpp`
- End2end example: `examples/air_to_async/channel_ops`

## Implementation
1. `air.channel` declarations are lowered to `i64` global memrefs.
2. `air.channel.put`: for each channel, we create a buffer in the heap memory, copy data from src memref to buffer, save the buffer address to the `i64` element in the global memref associated with the channel, and then immediately returns.
3. `air.channel.get`: we first get the buffer address, check if it is valid, copy data from buffer to dst memref, and then free the buffer.

Channel abstracts a wire-like connection instead of a FIFO with depth. I implemented put/get operations as "non-blocking" to work with async. For example, if one engine tries to get data from a channel but it has RAW dependency, this function call can be wrapped in a `async.execute` block and wait for a token.

## Ongoing
1. This example only works for 1D to 4D memrefs of int32 and float types, and the `air.channel.put` and `air.channel.get` operations should have all positional arguments filled. To support more cases, we need to handle their name mangling. This might result in too much boilerplate code.
2. Construct async examples for testing.
3. Currently, the `air-to-async` lowering pass does not handle `air.channel` lowering correctly. For example, an array of channels `air.channel [2, 2]` gets lowered to a `memref<i64>` instead of a `memref<2x2xi64>`. I plan to handle this in a different PR.


fifield:
> Currently, the air-to-async lowering pass does not handle air.channel lowering correctly. For example, an array of channels air.channel [2, 2] gets lowered to a memref<i64> instead of a memref<2x2xi64>. I plan to handle this in a different PR.

It is lowered to `i64` with the intent that it would be getting an opaque handle (i.e. a pointer) to a channel object allocated by the compiler or runtime. With that in mind all the type information of the channel (e.g. [2,2]) could be part of the channel object behind that opaque pointer. The semantics of `air.channel [2, 2]` aren't quite "array of channels" so lowering it to array of i64 might not be ideal.

fifield:
I don't fully understand the comments about implementing blocking with async blocks and tokens.
In general we're interested in lowering code like:
```
async {
   ...
   for (...) {
      air.channel.put @channel_0 [] ...
   }
   ...
}

async {
   ...
   for (...) {
      air.channel.get @channel_0 [] ...
   }
   ...
}
```
How does this get lowered with this scheme?

runtime_lib/aircpu/channel.cpp
```diff
@@ -0,0 +1,260 @@
...
+  mlir_air_channel_get_2d(M0I64_M0D2I32_I64_I64_I64_I64_I64_I64, int32_t);
+  mlir_air_channel_put_2d(M0I64_M0D2I32_I64_I64_I64_I64_I64_I64, int32_t);
+  mlir_air_channel_get_2d(M0I64_M0D2F32_I64_I64_I64_I64_I64_I64, float);
+  mlir_air_channel_put_2d(M0I64_M0D2F32_I64_I64_I64_I64_I64_I64, float);
+
+  // 1D
+  mlir_air_channel_get_1d(M0I64_M0D1I32_I64_I64_I64, int32_t);
+  mlir_air_channel_put_1d(M0I64_M0D1I32_I64_I64_I64, int32_t);
+  mlir_air_channel_get_1d(M0I64_M0D1F32_I64_I64_I64, float);
+  mlir_air_channel_put_1d(M0I64_M0D1F32_I64_I64_I64, float);
```
keryell, line 259: Just curious: would it be possible to use just explicit template instantiation instead of macros? I have the feeling it would synthesize automatically the right mangling according to the ABI. Or is it just some ACDC-private kind of mangling?

runtime_lib/aircpu/channel.cpp
```diff
@@ -0,0 +1,260 @@
...
+  {
+    offset[i] = _offset[i];
+    size[i] = _size[i];
+    stride[i] = _stride[i];
+  }
+  // allocate a buffer in heap memory
+  T *buffer = (T *)malloc(sizeof(T) * size[3] * size[2] * size[0] * size[1]);
+  if (VERBOSE)
+    printf("dst offset %lu, %lu, size %lu, %lu, stride %lu, %lu\n", offset[1],
+           offset[0], size[1], size[0], stride[1], stride[0]);
```
keryell, line 43: You love C. :-)
At least send debug/error message on `stderr` or use `std::cerr`?

fifield: sorry, we'll spell it `std::printf` next time.

examples/air_to_async/channel_ops/2mm/Makefile
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2023, Advanced Micro Devices, Inc.
```
stephenneuendorffer, line 1: I'd like to see unit tests that run as regressions in addition to examples.

runtime_lib/aircpu/channel.cpp
```diff
@@ -0,0 +1,240 @@
...
+    new_channel->_is_full = false;
+    channel->data[0] = (uint64_t)new_channel;
+  }
+
+  channel_t<T> *chan = (channel_t<T> *)channel->data[0];
+
+  // wait until the channel is empty
+  while (chan->_is_full)
+    ;
+
```
stephenneuendorffer, line 40: What are the concurrency assumptions about this code?  This would seem to not be very polite to just spin-wait?

keryell: Fortunately there is no race condition since it is equivalent to
```c++
const auto f = chan->_is_full;
while (f)
    /* If we loop this is for ever :-) */;
```
after loop-invariant code hoisting.

zzzDavid: `chan->_is_full` is not exactly loop-invariant. The assumption is that there is another async block that has channel.get which will empty the channel and set `chan->_is_full` to false.

stephenneuendorffer: Yeah, unfortunately, optimizing compilers will tend to treat this as loop invariant without 'volatile'.  Even then, an architecture can optimize memory accesses at runtime without additional memory barrier instructions.  Even then it seems that you are making assumptions about how async blocks are lowered, in particular that they are implemented as independent threads on separate cores with fairness guarantees, which may necessarily be true (or at least may not be performant)!  Perhaps there is an existing async library that we should be targetting here?  How does mlir/llvm lower its async primitives?

examples/air_to_async/channel_ops/2mm/mmult.mlir
```diff
@@ -0,0 +1,123 @@
...
+    %c0 = arith.constant 0 : index
+    %c32 = arith.constant 32 : index
+    %c1 = arith.constant 1 : index
+    %c0_i32 = arith.constant 0 : i32
+    %alloc = memref.alloc() {alignment = 64 : i64} : memref<32x32xi32>
+    linalg.fill ins(%c0_i32 : i32) outs(%alloc : memref<32x32xi32>)
+    %alloc_0 = memref.alloc() {alignment = 64 : i64} : memref<32x32xi32>
+    memref.copy %alloc, %alloc_0 : memref<32x32xi32> to memref<32x32xi32>
+    %0 = memref.get_global @channel_0 : memref<i64>
+    %1 = builtin.unrealized_conversion_cast %0 : memref<i64> to memref<i64>
```
erwei-xilinx, line 20: The PR looks great to me. Are those unrealized conversion casts generated from -air-to-std? If so, in one recent PR the -air-to-std pass has been extended to support channel puts and gets, without generating those ops.

erwei-xilinx: I agree with Steve's point that it would be helpful to add regression tests which test those features, so as to facilitate future code development on top of your work.

runtime_lib/aircpu/channel.cpp
zzzDavid: That makes sense, I'm taking a closer look at how async primitives are lowered in MLIR

keryell: Volatile would not solve the problem in a way compatible with the C++ memory model anyway since you want to have some causality between the `_is_full` and the rest of the program.
The C++ way to do it using polling (I do not say this is a good idea, but AMD has to sell a lot of CPU to pay our bills :-) ) would be to use some atomics,
for example
```c++
while (std::atomic_ref { chan->_is_full })
  ;
```

runtime_lib/aircpu/channel.cpp
```diff
@@ -0,0 +1,239 @@
...
+  if (channel->data[0] == 0) {
+    channel_t<T> *new_channel = new channel_t<T>(size);
+    channel->data[0] = (uint64_t)new_channel;
+  }
+
+  channel_t<T> *chan = (channel_t<T> *)channel->data[0];
+
+  // wait until the channel is empty
+  std::unique_lock<std::mutex> lock(chan->mtx);
+  chan->cv.wait(lock, [&chan] { return !chan->_is_full; });
```
zzzDavid, line 36: Synchronization between put and get is now implemented with conditional variables

examples/air_to_async/channel_ops/2mm/mmult.mlir
zzzDavid: The unrealized conversion casts are generated by `-air-to-async`. I have moved tests from examples to `mlir/tests`

runtime_lib/aircpu/channel.cpp
zzzDavid: Thanks for the suggestion! After some discussion with Steve, I implemented synchronization with condition variables. Please check out my newer commits.

keryell: Just thinking again at the interesting original code sample:
```c++
  while (chan->_is_full)
    ;
```
Since this is a correct C++, it is a race-free program, so `chan->_is_full` cannot change.
Then, since the abstract C++ machine executing a conformant program has to execute a step eventually (basically a side-effect), a C++ program cannot have an infinite loop.
So the compiler can prove `chan->_is_full` is always false and remove this loop. ;-)

fifield (approved):
I might move the tests around at some point. Otherwise looks good to me.

PR #249, opened 2023-06-22
controller: add command shell to UART

jnider:
Use the UART to enter commands in an interactive shell. This is a very powerful debug tool as it enables interacting with the device directly during operation.

This is first being used while debugging AIRBIN but has general applicability.

### DIFF
--- a/runtime_lib/controller/platform.h
+++ b/runtime_lib/controller/platform.h
@@ -1,34 +1,9 @@
-/******************************************************************************
-*
-* Copyright (C) 2008 - 2014 Xilinx, Inc.  All rights reserved.
-*
-* Permission is hereby granted, free of charge, to any person obtaining a copy
-* of this software and associated documentation files (the "Software"), to deal
-* in the Software without restriction, including without limitation the rights
-* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-* copies of the Software, and to permit persons to whom the Software is
-* furnished to do so, subject to the following conditions:
-*
-* The above copyright notice and this permission notice shall be included in
-* all copies or substantial portions of the Software.
-*
-* Use of the Software is limited solely to applications:
-* (a) running on a Xilinx device, or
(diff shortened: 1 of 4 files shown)


jgmelber (approved):

runtime_lib/controller/platform.h
```diff
@@ -1,34 +1,9 @@
...
-* in advertising or otherwise to promote the sale, use or other dealings in
-* this Software without prior written authorization from Xilinx.
-*
-******************************************************************************/
+//===- platform.h -----------------------------------------------*- C++ -*-===//
+//
+// Copyright (C) 2023, Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
```
fifield, line 6: Note this file was generated by a Xilinx tool with the former header. I'm not a lawyer so I have no opinion on changing it, just letting you know the reason for it.

runtime_lib/controller/shell.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+void shell(void) {
+  uint8_t in;
+
+  if (new_cmd) {
+    new_cmd = 0;
+    cmd_len = 0;
+
+    // good old DOS prompt
+    xil_printf("C:\\> ");
```
fifield, line 139: why?

jnider: This is an interactive shell, so it needs a prompt to tell users that they can enter commands. This particular prompt is chosen to be distinguishable from bash or other commonly used shells. When a user has many windows open, it helps identify that this shell is the UART to the device (as opposed to the the host machine).

fifield: I understand why there is a prompt. I don't understand why it's a fake DOS prompt instead of something that identifies what it actually is.

jnider: I explained: this is to differentiate it from bash that uses % or $. If you want to pick a different character I'm open to suggestions.

PR #250, opened 2023-06-22
Read write 32 squash

eddierichter-amd:
This is the squashed version of the read-write-32 PR.  

Adding functionality in the driver and runtime to generate read/write 32 packets to perform AIE configuration. In this implementation, the driver uses the first queue, which is mapped to the ARM, and can send read/write 32 packets to the firmware in response to read/write operations to the driver's sysfs interface. Currently still have the AIE BAR in the platform, which is unused but not hurting anything. Will remove that from the platform once we completely deprecate the MMIO configuration.


docs/building.md
```diff
@@ -120,9 +124,9 @@ export PATH=/path/to/mlir-air/install-pcie/bin:${PATH}
 
 The instructions for building the hardware platform designs are found in the mlir-air/platforms directory:
 
-- [xilinx_vck190_air](platforms/xilinx_vck190_air)
-- [xilinx_vck5000_air](platforms/xilinx_vck5000_air)
-- [xilinx_vck5000_air_scale_out](platforms/xilinx_vck5000_air_scale_out)
+- [xilinx_vck190_air](../platforms/xilinx_vck190_air)
```
jnider, line 127: This is somewhat problematic and keeps arising. @fifield has explained this to me in the past - the links work correctly on github.io (https://xilinx.github.io/mlir-air/building.html) but don't work correctly when browsing the repo directly. No known solution at this time.

docs/building.md
```diff
@@ -158,6 +162,12 @@ cd ..
 tar -cvf air_tools.tar.gz install-aarch64
 ```
 
+## Running on the VCK5000
+
+The first step required for running on the VCK5000 is building the hardware platform and loading it on your device. [Here](../platforms/xilinx_vck5000_air) we provide documentation on how to do that. Then, the AIR tools must be installed by running the [Building external projects on X86](#building-external-projects-on-X86) followed by the [Building on x86 with runtime for PCIe](#building-on-x86-with-runtime-for-PCIe) and [Environment setup](#environment-setup). Then, the [AIR PCIe driver](../driver) must be compiled and installed. Now you are ready to run tests using AIR on the VCK5000! For example, go to [Test 13](../test/13_mb_add_one) and run `Make` followed by `sudo ./test.elf` to run the application.
+
```
jnider, line 168: 'make' should be lowercase

runtime_lib/airhost/host.cpp
```diff
@@ -325,6 +325,13 @@ uint64_t air_segment_load(const char *name) {
 
 #ifdef AIR_PCIE
   XAie_Finish(&(_air_host_active_libxaie->DevInst));
+
+  // Setting the driver libxaie backend back up
+  // Currently only targetting device 0
+  _air_host_active_libxaie->AieConfigPtr.Backend = XAIE_IO_BACKEND_AMDAIR;
+  _air_host_active_libxaie->DevInst.IOInst =
+      (void *)"/sys/class/amdair/amdair/00";
```
jnider, line 333: I'm not really familiar with this function - what needs to change to make it work with multiple devices?

runtime_lib/controller/main.cpp
```diff
@@ -61,7 +61,7 @@ int shim_dma_cols[NUM_SHIM_DMAS] = {2,  3,  6,  7,  10, 11, 18, 19,
 int col_dma_cols[NUM_COL_DMAS] = {7, 8, 9, 10};
 #define NUM_DMAS (NUM_SHIM_DMAS + NUM_COL_DMAS)
 
-#define CHATTY 1
+#define CHATTY 0
```
jnider: This is why I don't like controlling features from inside source files :-)

driver/chardev.c
```diff
@@ -358,25 +361,71 @@ static ssize_t address_store(struct kobject *kobj, struct attribute *attr,
...
+  iowrite32((uint32_t)is_write,               pkt + PKT_ARG_ADDR_OFFSET + 0xC );  // Writing the is_write
+  iowrite32(header_type, pkt + PKT_HEADER_TYPE_OFFSET); // Writing dispatch header type and rw32 dude
+  iowrite32(0x00000001, pkt + PKT_COMPL_OFFSET); // Writing a 1 to the signal which will mark the completion
+
+  // Step 3: Ring the doorbell and poll on the read pointer
+  iowrite32(wr_idx, drv_priv->admin_queue + QUEUE_DOORBELL_OFFSET);
+  iowrite32(0, drv_priv->admin_queue + QUEUE_DOORBELL_OFFSET + 0x4);
+
+  read_completion = ioread32(pkt + PKT_COMPL_OFFSET);
+  while(read_completion != 0x00000000) {
```
jnider: (read_completion != 0) is enough (don't need all those 0000000's)
even "while (read_completion)" does the same thing, since this looks like a boolean value anyway

driver/chardev.c
```diff
@@ -358,25 +361,71 @@ static ssize_t address_store(struct kobject *kobj, struct attribute *attr,
...
+  header_type = (AIR_PKT_TYPE_RW32 << 16) | (HSA_PACKET_TYPE_AGENT_DISPATCH << HSA_PACKET_HEADER_TYPE);
+  iowrite32(HSA_PACKET_TYPE_INVALID, pkt + PKT_HEADER_TYPE_OFFSET);
+  iowrite32((uint32_t)(offset & 0xFFFFFFFF),  pkt + PKT_ARG_ADDR_OFFSET       );  // Writing the addr low bits
+  iowrite32((uint32_t)(offset >> 32),         pkt + PKT_ARG_ADDR_OFFSET + 0x4 );  // Writing the addr high bits
+  iowrite32(packet_value,                     pkt + PKT_ARG_ADDR_OFFSET + 0x8 );  // Writing the value, doesn't matter because read
+  iowrite32((uint32_t)is_write,               pkt + PKT_ARG_ADDR_OFFSET + 0xC );  // Writing the is_write
+  iowrite32(header_type, pkt + PKT_HEADER_TYPE_OFFSET); // Writing dispatch header type and rw32 dude
+  iowrite32(0x00000001, pkt + PKT_COMPL_OFFSET); // Writing a 1 to the signal which will mark the completion
+
+  // Step 3: Ring the doorbell and poll on the read pointer
```
jnider: please add a wmb() here to be sure all previous writes complete before hitting the doorbell. It is hard to debug when the device sees a stale value _sometimes_.

driver/chardev.c
```diff
@@ -388,10 +437,53 @@ static ssize_t value_store(struct kobject *kobj, struct attribute *attr,
...
+    header_type = (AIR_PKT_TYPE_RW32 << 16) | (HSA_PACKET_TYPE_AGENT_DISPATCH << HSA_PACKET_HEADER_TYPE);
+    iowrite32(HSA_PACKET_TYPE_INVALID, pkt + PKT_HEADER_TYPE_OFFSET);
+    iowrite32((uint32_t)(offset & 0xFFFFFFFF),  pkt + PKT_ARG_ADDR_OFFSET       );  // Writing the addr low bits
+    iowrite32((uint32_t)(offset >> 32),         pkt + PKT_ARG_ADDR_OFFSET + 0x4 );  // Writing the addr high bits
+    iowrite32(value,                            pkt + PKT_ARG_ADDR_OFFSET + 0x8 );  // Writing the value, doesn't matter because read
+    iowrite32((uint32_t)is_write,               pkt + PKT_ARG_ADDR_OFFSET + 0xC );  // Writing the is_write
+    iowrite32(header_type, pkt + PKT_HEADER_TYPE_OFFSET); // Writing dispatch header type and rw32 dude
+    iowrite32(0x00000001, pkt + PKT_COMPL_OFFSET); // Writing a 1 to the signal which will mark the completion
+
+    // Step 3: Ring the doorbell and poll on the read pointer
```
jnider: same here

driver/kernel_queue.c
```diff
@@ -0,0 +1,57 @@
...
+#include "hsa_defs.h"
+
+/*  
+  Get the address to queue 0 and point the admin queue of the device to it.
+*/
+hsa_status_t air_queue_create(uint32_t size, uint32_t type, struct vck5000_device *dev) {
+
+  // Get the base address of the queue, subtract the physical offset, 
+  // and point our admin queue pointer to it
+  dev->admin_queue = dev->bram_bar + HERD_CONTROLLER_BASE_ADDR(dev->bram_bar, 0) - BRAM_PADDR;;
```
jnider: too many ;;

jnider:
Thanks for squashing - 3 commits was much easier to review than 22. However, there are still some instances of files that change twice. For example, both the first commit (2aa3229a) and the 3rd commit (6fe0bd58) touch chardev.c, whereas ideally all of the changes to each file would be in a single commit.

jnider (changes requested):

runtime_lib/airhost/host.cpp
eddierichter-amd: Segments are a higher-level construct, and segment_load load is called within AIR-level tests. To support multiple devices, we would need the runtime to choose a device where it could load the segment. I have some ideas, lets discuss offline.

driver/kernel_queue.c
eddierichter-amd: Addressed in [391351c](https://github.com/Xilinx/mlir-air/pull/250/commits/391351c6ed10c4cca742dad0a07c25f1bdb0666c)

driver/chardev.c
eddierichter-amd: Addressed in [391351c](https://github.com/Xilinx/mlir-air/pull/250/commits/391351c6ed10c4cca742dad0a07c25f1bdb0666c)

driver/chardev.c
eddierichter-amd: Thanks for pointing this out, very good catch. Addressed in [391351c](https://github.com/Xilinx/mlir-air/pull/250/commits/391351c6ed10c4cca742dad0a07c25f1bdb0666c).

driver/chardev.c
eddierichter-amd: Fair enough, addressed in [391351c](https://github.com/Xilinx/mlir-air/pull/250/commits/391351c6ed10c4cca742dad0a07c25f1bdb0666c).

driver/chardev.c
```diff
@@ -35,7 +35,7 @@
 #define HERD_CONTROLLER_BRAM_SIZE 0x1000
 
 #define USE_RW32_PKTS
```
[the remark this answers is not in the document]
eddierichter-amd: Yeah it is a good point, it was more for debugging purposes when I first implemented it and then it just stuck around. Just going to remove it. Addressed in [391351c](https://github.com/Xilinx/mlir-air/pull/250/commits/391351c6ed10c4cca742dad0a07c25f1bdb0666c).

docs/building.md
eddierichter-amd: Thanks for the catch. Addressed in [b86101c](https://github.com/Xilinx/mlir-air/pull/250/commits/b86101c13f422a0484b17d933b6a2699fd8db9a4).

docs/building.md
eddierichter-amd: Do the relatively paths not work on github.io?

fifield: The relative paths do not work and the preference has been to have it work in the generated docs instead of browsing the repo. There is some directory reorganization needed to fix this. One option is to keep the platform docs in the docs area instead of in the respective platform folders. The current [deployment code](https://github.com/Xilinx/mlir-air/blob/main/.github/workflows/generateDocs.yml#L97).

eddierichter-amd: Thanks for pointing that out didn't realize that. The organizational structure sounds like a separate PR, I will change the docs back to what they were.

eddierichter-amd: Addressed in [447d44f](https://github.com/Xilinx/mlir-air/pull/250/commits/447d44ff94c68846fea0cd8344ac75836bb9946f).

jgmelber (approved):

PR #253, opened 2023-07-05
Segment and device initialization

eddierichter-amd:
* Adding NPI level 
* Changing AIE level tests to run device init and then herd init (before were running in the opposite order)
* Small fix to runtime shell to not block on reading from UART


runtime_lib/controller/main.cpp
```diff
@@ -323,7 +323,7 @@ void mlir_aie_print_shimdma_status(aie_libxaie_ctx_t *ctx, int col, int row) {
              col, row, dma_mm2s_status, dma_mm2s0_control, dma_mm2s1_control,
              dma_s2mm_status, dma_s2mm0_control, dma_s2mm1_control, dma_bd0_a,
              dma_bd0_control);
-  for (int bd = 0; bd < 8; bd++) {
```
keryell, line 326: Perhaps 16 could be a `constexpr` parameter somewhere?

runtime_lib/controller/main.cpp
```diff
@@ -820,6 +820,48 @@ void xaie_l2_dma_init(int col) {
   xaie::out32(xaie::getTileAddr(col, 0) + 0x00033008, 0xFF);
 }
 
+#ifdef ARM_CONTROLLER
+
+// Defining the NPI base and registers we use to reset the array
+uint64_t npi_base = 0xF70A0000UL;
+#define NPI_MASK_REG 0x0
```
keryell, line 827: Since you are lucky enough to use C++ instead of C :-)
```suggestion
constexpr uint64_t npi_base = 0xF70A0000UL;
constexpr auto NPI_MASK_REG = 0x0;
```
?

runtime_lib/controller/main.cpp
```diff
@@ -820,6 +820,48 @@ void xaie_l2_dma_init(int col) {
...
+#define NPI_VAL_REG 0x4
+#define NPI_LOCK_REG 0xC
+void xaie_array_reset() {
+
+  // Getting a pointer to NPI
+  volatile uint32_t *npib = (volatile uint32_t *)(npi_base);
+
+  // Performing array reset sequence
+  air_printf("Starting array reset\r\n");
+  npib[NPI_LOCK_REG >> 2] = 0xF9E8D7C6;
```
keryell, line 837: A lot of magical incantation for an open-source project. Perhaps adding some comments at least about what it means?

runtime_lib/controller/main.cpp
```diff
@@ -820,6 +820,48 @@ void xaie_l2_dma_init(int col) {
...
+
+// Defining the NPI base and registers we use to reset the array
+uint64_t npi_base = 0xF70A0000UL;
+#define NPI_MASK_REG 0x0
+#define NPI_VAL_REG 0x4
+#define NPI_LOCK_REG 0xC
+void xaie_array_reset() {
+
+  // Getting a pointer to NPI
+  volatile uint32_t *npib = (volatile uint32_t *)(npi_base);
```
keryell, line 833: You can avoid repeating the type in C++
```suggestion
  auto npib = (volatile uint32_t *)npi_base;
```

eddierichter-amd: Thanks for the pointer. Changed in [173d9da](https://github.com/Xilinx/mlir-air/pull/253/commits/173d9da6995d41a594bd9dc3e6d8972005afa95f).

runtime_lib/controller/main.cpp
eddierichter-amd: Don't want to go into too much detail of the content of the NPI registers, but added some general comments on when the reset is happening in [173d9da](https://github.com/Xilinx/mlir-air/pull/253/commits/173d9da6995d41a594bd9dc3e6d8972005afa95f)

jgmelber (approved):

jgmelber (changes requested):
Change from herd init to segment init. Add unit test for the reset behavior.

eddierichter-amd:
> Change from herd init to segment init. Add unit test for the reset behavior.

[0ac6900](https://github.com/Xilinx/mlir-air/pull/253/commits/0ac69009e821e361a506403077c38de0d52f06f5) is the unit test and [fcc0bf3](https://github.com/Xilinx/mlir-air/pull/253/commits/fcc0bf37f31ede6a2393cccf203c8820bddd4ee7) is the changing of herd init to segment init.

runtime_lib/airhost/host.cpp
```diff
@@ -345,11 +345,11 @@ uint64_t air_segment_load(const char *name) {
 
   wr_idx = queue_add_write_index(_air_host_active_segment.q, 1);
   packet_id = wr_idx % _air_host_active_segment.q->size;
-  dispatch_packet_t *herd_pkt =
+  dispatch_packet_t *segment_pkt =
       (dispatch_packet_t *)(_air_host_active_segment.q->base_address_vaddr) +
       packet_id;
-  air_packet_herd_init(herd_pkt, 0, 0, 50, 1, 8);
-  air_queue_dispatch_and_wait(_air_host_active_segment.q, wr_idx, herd_pkt);
+  air_packet_segment_init(segment_pkt, 0, 0, 50, 1, 8);
```
jgmelber, line 351: We should determine the information we need to only initialize/reset the resources we need for a given segment and replicate your unit test at the "AIR" level

jgmelber (approved):
Before merging, propagate the name change from herd_init to segment_init in the DU controller firmware.

runtime_lib/controller/main.cpp
```diff
@@ -820,6 +822,60 @@ void xaie_l2_dma_init(int col) {
...
+
+// Defining the NPI base and registers we use to reset the array
+constexpr uint64_t npi_base = 0xF70A0000UL;
+constexpr auto NPI_MASK_REG = 0x0;
+constexpr auto NPI_VAL_REG = 0x4;
+constexpr auto NPI_LOCK_REG = 0xC;
+void xaie_array_reset() {
+
+  // Getting a pointer to NPI
+  auto *npib = (volatile uint32_t *)(npi_base);
```
keryell, line 835:
```suggestion
  auto npib = (volatile uint32_t *)(npi_base);
```
In that case it is obvious it is a pointer, but if you want you can also keep the `*` to insist on it.

eddierichter-amd:
> Before merging, propagate the name change from herd_init to segment_init in the DU controller firmware.

Done in [79e8bf7](https://github.com/Xilinx/mlir-air/pull/253/commits/79e8bf713bf9c099977d5d46bcc419d0dffbfa40)

PR #256, opened 2023-07-17
AIR async channel to aie.locks

erwei-xilinx:
- Initial draft of adding an alternative pass to lower sync or async air.channel in `air-to-aie` pass.


mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -151,24 +288,18 @@ struct DMAAllocator {
     }
   }
 
-  AIE::TileOp getTile(AIE::DeviceOp aie_device, air::DmaMemcpyInterface &dmaOp,
+  template <typename T>
+  AIE::TileOp getTile(AIE::DeviceOp aie_device, T &memcpyOp,
```
fifield, line 292: It has occurred to me in the past that we could unify some or all of the ChannelInterface and DmaMemcpyInterface which is basically what's going on here via the template and all the new helper functions like `getSizesInterface` and `isTileInbound/isTileOutbound`.

fifield (approved):
Thanks for doing this. What do you think about using MLIR interface throughout the patch instead of templates? This patch seems to prove there is enough overlap between the channel and memcpy ops to write algorithms that work on either.

mlir/lib/Conversion/AIRToAIEPass.cpp
erwei-xilinx: I agree with you--unifying the two interfaces makes sense as they are almost the same. I will unify them in a future PR so that this PR only focuses on getting channel-to-locks pass working.

PR #258, opened 2023-07-19
AIR compilation scripts: have MLIR-AIR inherit LibXAIE path that MLIR-AIE was built against

erwei-xilinx:
- Fixup an issue where sysroot is not imported when building air-runtime as an external project to AIR.
- Bump mlir-aie.
- Have mlir-air inherit libxaie from mlir-aie, so that both projects are built against the same libxaie installation.


utils/cross-build-mlir-air.sh
```diff
@@ -69,6 +69,7 @@ cmake .. \
     -Daarch64_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE} \
     -DLLVM_DIR=${LLVM_DIR}/build-aarch64/lib/cmake/llvm \
     -DMLIR_DIR=${LLVM_DIR}/build-aarch64/lib/cmake/mlir \
+    -DLibXAIE_ROOT=${CMAKE_SYSROOT}/opt/xaienginev2 \
```
stephenneuendorffer, line 72: This I'm not so sure about, since it assume libXAIE is installed in the sysroot?

erwei-xilinx: Yes my understanding is that (i) with cross compile there must be a sysroot, and (ii) a sysroot always has an installation of libxaie under its opt dir. Are both statements true?

stephenneuendorffer: libxaie doesn't have to be installed under /opt.  In fact, in mlir-aie, we cross-compile our own version of libXAIE from source, so it seems like air should just use that?

erwei-xilinx: I see. I'll update the script to allow users to pass in their own installation of libxaie.

erwei-xilinx:
This PR makes mlir-air import libxaie path from mlir-aie. Depends on this PR from mlir-aie, which exports libxaie paths: https://github.com/Xilinx/mlir-aie/pull/549

utils/build-mlir-air-pcie.sh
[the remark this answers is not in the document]
erwei-xilinx: I modified the mlir-aie build script in mlir-air/utils, so that it now points to air's custom backend instead of using what mlir-aie builds as libxaie path. Hope that it works now.

My view is that since (i) mlir-aie is required to build mlir-air, and (ii) both mlir-aie and mlir-air must be built against the same libxaie installation, it makes sense to have only mlir-aie import libxaie path, while mlir-air should inherit whatever libxaie that mlir-aie was built against.

utils/cross-build-mlir-aie.sh
```diff
@@ -47,31 +43,21 @@ mkdir -p $INSTALL_DIR
...
-fi
-
 ninja install |& tee ninja-install.log
 
-ec=$?
-if [ $ec -ne 0 ]; then
-    echo "Build Error"
-    exit $ec
-fi
-
```
fifield, line 76: Revert the removal of error checking. This is required when I run this script in docker, otherwise it keeps going and I don't know where it failed.

utils/cross-build-mlir-aie.sh
```diff
@@ -47,31 +43,21 @@ mkdir -p $INSTALL_DIR
...
+    -DSysroot=${CMAKE_SYSROOT} \
+    -DArch=arm64 \
     -DLLVM_DIR=${LLVM_DIR}/build-aarch64/lib/cmake/llvm \
     -DMLIR_DIR=${LLVM_DIR}/build-aarch64/lib/cmake/mlir \
-    -DVitisSysroot=${CMAKE_SYSROOT} \
+    -DLLVM_USE_LINKER=lld \
+    -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} \
+    -DAIE_ENABLE_BINDINGS_PYTHON=ON \
+    -DCMAKE_BUILD_TYPE=Debug \
+    -Wno-dev \
```
fifield, line 58: Most of this should be reverted:
* The toolchain file should not be hard coded, or it should at least use the air aarch64 toolchain file.
* The build type should not be hard coded
* We don't need -DSysroot -DArch if using the air aarch64 toolchain file
* LLVM_USE_LINKER should be set by toolchain file

fifield (changes requested):

utils/build-mlir-aie-local.sh
```diff
@@ -8,52 +8,54 @@
...
-    -DCMAKE_MODULE_PATH=${CMAKEMODULES_DIR}/ \
+    -DLibXAIE_x86_64_DIR=${LibXAIE_x86_64_DIR} \
+    -DCMAKE_MODULE_PATH=${CMAKEMODULES_DIR}/modulesXilinx \
     -DCMAKE_INSTALL_PREFIX="../${INSTALL_DIR}" \
-    -DBUILD_SHARED_LIBS=off \
     -DCMAKE_BUILD_TYPE=Release \
     -DLLVM_ENABLE_ASSERTIONS=ON \
     -DAIE_ENABLE_BINDINGS_PYTHON=ON \
+    "-DAIE_RUNTIME_TARGETS=x86_64;aarch64" \
+    -DAIE_RUNTIME_TEST_TARGET=aarch64 \
```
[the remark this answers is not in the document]
erwei-xilinx: This script builds mlir-aie targetting X86, and the cross-build-mlir-aie.sh builds mlir-aie targeting aarch64. There is no file named build-mlir-aie-pcie.sh.

utils/cross-build-mlir-aie.sh
keryell: If it was useful someone would have commented the use case, right? ;-)
So we are free to remove it? :-)
So, please add a lot of comments stating the hypothesis and whatever for the newcomers for  a more inclusive world.

utils/build-mlir-aie-local.sh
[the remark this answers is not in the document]
fifield: sorry, I misread the name. My comment applies to build-mlir-air.sh vs build-mlir-air-pcie.h. It seems like those can be combined.

Why are we hard coding setting the test target to aarch64?

utils/cross-build-mlir-aie.sh
fifield: Checking return codes is common good practice. It should not need a comment `// check for error`

utils/build-mlir-aie-local.sh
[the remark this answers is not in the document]
erwei-xilinx: This line is copied over from mlir-aie's x86 compilation scrip: https://github.com/Xilinx/mlir-aie/blob/08aff39136b2c797d0f01915f11de4a3ff1edc57/utils/build-mlir-aie.sh#L50. My knowledge of mlir-aie's build flow is limited, and I don't know for sure why they test against aarch64 rather than x86...

utils/cross-build-mlir-aie.sh
fifield: > Hi Jeff, thanks for your responses. This script, which cross compiles mlir-aie was trying to be conform with mlir-aie's ARM cross compile script: https://github.com/Xilinx/mlir-aie/blob/main/utils/build-mlir-aie-cross.sh
> 
> * The toolchain file should not be hard coded, or it should at least use the air aarch64 toolchain file.
>   -- I think if mlir-aie's build process depends on an air toolchain file, then there seems to be a circular dependency where air's aarch64 toolchain -> mlir-aie repo -> mlir-air. Then when mlir-aie bumps, should air developers have to maintain aarch64 toolchain file accordingly too? I wonder if it will be possible to use aie's own build process, using its own toolchain file? Is there a fundamental difference between air's aarch64 toolchain file and aie's clang crosscomp pync toolchain file, that we in AIR must maintain?

Just don't hard code it please so that I can use my own.

erwei-xilinx: > > Hi Jeff, thanks for your responses. This script, which cross compiles mlir-aie was trying to be conform with mlir-aie's ARM cross compile script: https://github.com/Xilinx/mlir-aie/blob/main/utils/build-mlir-aie-cross.sh
> > 
> > * The toolchain file should not be hard coded, or it should at least use the air aarch64 toolchain file.
> >   -- I think if mlir-aie's build process depends on an air toolchain file, then there seems to be a circular dependency where air's aarch64 toolchain -> mlir-aie repo -> mlir-air. Then when mlir-aie bumps, should air developers have to maintain aarch64 toolchain file accordingly too? I wonder if it will be possible to use aie's own build process, using its own toolchain file? Is there a fundamental difference between air's aarch64 toolchain file and aie's clang crosscomp pync toolchain file, that we in AIR must maintain?
> 
> Just don't hard code it please so that I can use my own.

Reverted to use your own toolchain.

PR #260, opened 2023-07-24
Adding the AIR version of the weather stencil to the AIR repo

eddierichter-amd:
Adding AIR versions of the 1 and 4 bblock designs to example directory.

### DIFF
--- a/examples/weather_stencil_air/hdiff_base.py
+++ b/examples/weather_stencil_air/hdiff_base.py
@@ -1,0 +1,45 @@
+import numpy as np
+
+ROW1 = 9
+COL = 256
+
+inp1 = np.zeros((ROW1*COL,))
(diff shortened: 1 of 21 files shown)


examples/weather_stencil_air/hdiff_base.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
```
jgmelber, line 1: License header?

jgmelber (changes requested):
I think we need to also have the AMD license header on these files?

fifield:
Maybe I missed something, but I did not see any AIR design here?  I only found mlir-aie dialect...

jgmelber:
It's just the airrt data movement version of the stencil design using HSA packets

PR #263, opened 2023-07-31
Channel to ObjectFIFO Broadcast

abisca:
This PR adds a lowering from channels with a broadcast_shape attribute to objectFifos that broadcast to multiple consumers.

It also adds support in the SpecializeChannelBundle pattern in the AIRToAIE conversion pass to change the broadcast_shape based on the initial channel bundle size.

Finally, it also removes the memory space from memrefs during the channel to objectFifo lowering.


mlir/test/Conversion/AIRToAIE/specialize_channel_broadcast.mlir
```diff
@@ -0,0 +1,24 @@
...
+// CHECK:   %[[VAL_3:.*]] = AIE.tile(2, 2)
+// CHECK-COUNT-6:    air.channel @{{.*}}[1, 3] {broadcast_shape = [2, 1]}
+// CHECK: }
+   
+AIE.device(xcvc1902) {
+    %0 = AIE.tile(1, 1)
+    %1 = AIE.tile(2, 1)
+    %2 = AIE.tile(1, 2)
+    %3 = AIE.tile(2, 2)
+    air.channel @channel_0 [2, 3] {broadcast_shape = [4, 3]}
```
erwei-xilinx, line 23: Our current air.channel broadcast notation follows Numpy's broadcast format: https://numpy.org/doc/stable/user/basics.broadcasting.html#general-broadcasting-rules So under that rule, this broadcast is not legal, because how dimension 0 is broadcasted from 2 to 4 is undefined. 

[1,3] -> [4,3] and [4,1] -> [4,3] are examples of legal broadcast shapes.

mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
```diff
@@ -1171,6 +1171,40 @@ void WaitAllOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    if (bundle_size.size() != broadcast_shape.size())
+      return emitOpError("bundle size should match broadcast_shape");
+    int diffDims = 0;
+    for (int i = 0; i < (int)bundle_size.size(); i++)
+      if (dyn_cast<IntegerAttr>(bundle_size[i]).getInt() !=
+          dyn_cast<IntegerAttr>(broadcast_shape[i]).getInt())
+        diffDims++;
+    if (diffDims > 1)
+      return emitOpError("bundle sizes and broadcast_shape should only differ "
+                         "along one dimension");
```
erwei-xilinx, line 1187: Thanks for adding a verifier to check for broadcast shape's legality. For this condition, I recommend modifying it to follow Numpy broadcast's rule which says if dimension n is different between input and output, then one of the two values must be 1: https://numpy.org/doc/stable/user/basics.broadcasting.html#general-broadcasting-rules

abisca:
Thank you Erwei for the review, I will set the PR to draft while I add the changes!

erwei-xilinx (approved):

PR #280, opened 2023-08-18
catch up to triton's llvm commit (b126ee6)

makslevental:
This is a bump to https://github.com/llvm/llvm-project/commit/b126ee6, which is not actually where Triton is today but [where my fork is](https://github.com/makslevental/triton/blob/mlir_bindings_2/python/setup.py#L92).

There's a lot that had to be changed because the last bump was in January. Primarily it's `affine` being moved to its own namespace and the removal of `llvm::Optional` in favor of `std::optional`. 

In addition, transform interfaces changed so the transform ops had to be moved around and ported to the new interface.

Depends on https://github.com/Xilinx/mlir-aie/pull/590

### DIFF
--- a/mlir/include/air/Dialect/AIR/AIR.td
+++ b/mlir/include/air/Dialect/AIR/AIR.td
@@ -121,35 +121,35 @@ def air_SegmentOp : air_Op<"segment", [air_AsyncOpInterface,
     }
     static StringRef getColOffsetAttrName() { return "x_loc"; }
     static StringRef getRowOffsetAttrName() { return "y_loc"; }
-    mlir::Optional<uint64_t> getColOffset() {
+    std::optional<uint64_t> getColOffset() {
       auto name = getColOffsetAttrName();
       if (auto a = (*this)->getAttrOfType<IntegerAttr>(name)) {
         return a.getInt();
       }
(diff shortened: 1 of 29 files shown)


fifield:
I think I've included most of this in https://github.com/Xilinx/mlir-air/pull/280, although I fixed up affine differently.

keryell:
I guess this will break Polygeist if it has not evolved recently?

makslevental:
> I guess this will break Polygeist if it has not evolved recently?

Polygeist is on 16: https://github.com/llvm/Polygeist/commit/dceabaac80a5367ba342c5f18b311b38f402498b. Current cut is 18.dev.

keryell:
> > I guess this will break Polygeist if it has not evolved recently?
> 
> Polygeist is on 16: [llvm/Polygeist@dceabaa](https://github.com/llvm/Polygeist/commit/dceabaac80a5367ba342c5f18b311b38f402498b). Current tip is 18.

Yes, I have vainly spent a day already to port Polygeist to a recent version of LLVM. :-(
On the other hand there is the Polygeist version of Intel in their monorepo https://github.com/intel/llvm/tree/sycl-mlir/polygeist which works with a recent version of Polygeist if we really need it with MLIR-AIR.

PR #283, opened 2023-08-22
[NO MERGE] build and test wheels

makslevental:
This PR correct a few CMake mistakes and enables building wheels that can co-exist with other python wheels/distributions. Prior to this, the C extension could not be loaded concurrently (same process) with any other MLIR C extensions.


.github/workflows/buildAndTest.yml
```diff
@@ -127,6 +127,25 @@ jobs:
...
+          cp -R build_release/install/python/air python/wheel
+          
+          pushd python/wheel
+          
+          pip install wheel
+          pip wheel . -v
+          pip install air-*whl
+          python ../test/dialect/air_ops.py
+          
+          popd
```
keryell, line 147: Why this `pushd`/`popd`? At the end you exit anyway, so why not `cd`?

python/wheel/.gitignore
```diff
@@ -0,0 +1 @@
+air
```
keryell, line 1: Missing EOL.

.github/workflows/buildAndTest.yml
```diff
@@ -127,6 +127,25 @@ jobs:
...
 
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      # Build the repo test target in release mode to build and test.
+      - name: Build and test Python wheel
+        run: |
+          cp -R build_release/install/python/air python/wheel
```
keryell, line 138:
```suggestion
          cp --archive build_release/install/python/air python/wheel
```
I tend to prefer this by keeping more stuff, including links for example. But of course it depends on the use case.

.github/workflows/buildAndTest.yml
makslevental: it's the same thing as the EOL you suggested in the gitignore (or a trailing comma in a list) - in case the next person comes along and adds a job without noticing.

keryell: I do not consider EOL with any semantics but mostly syntax.
It is just required by the C standard which is so praised in hardware companies. ;-) https://www.open-std.org/JTC1/SC22/WG14/www/docs/n3096.pdf page 10:
```
A source file that is
not empty shall end in a new-line character, which shall not be immediately preceded by a
backslash character before any such splicing takes place.
```
For the `pushd`/`popd` `bash`-only aspect, this is a good idea but I wood prefer with a C++ RAII or Python context manager semantics, such as using more general `(` and `)` in `sh`:
```
          (
            pushd python/wheel
            pip install wheel
            pip wheel . -v
            pip install air-*whl
            python ../test/dialect/air_ops.py
          )
```
because if anything breaks in the middle it is still safe.

PR #286, opened 2023-08-24
format everything

makslevental:


stephenneuendorffer (changes requested):
Nack this...  Generally wholesale reformatting like this just causes churn... better to just let the CI handle the changes naturally over time as files are touched.

makslevental:
> Nack this... Generally wholesale reformatting like this just causes churn... better to just let the CI handle the changes naturally over time as files are touched.

mlir/include/air/Util/Outliner.h
```diff
@@ -20,18 +20,15 @@ namespace air {
 struct AIROutliner {
 
 public:
-  AIROutliner()
-  {
-  }
+  AIROutliner() {}
```
makslevental, line 23: @stephenneuendorffer 

> Nack this... Generally wholesale reformatting like this just causes churn... better to just let the CI handle the changes naturally over time as files are touched.

The CI is having a hard time catching much since there's no `.clang-format` at the root of the repo (which might be why you have formatting fails like this hanging around).

PR #293, opened 2023-09-04
Use correct macros for C API

makslevental:
This PR uses the correct upstream CMake macros for building the C API (as well as the correct C export macros).

image

### DIFF
--- a/mlir/lib/CAPI/CMakeLists.txt
+++ b/mlir/lib/CAPI/CMakeLists.txt
@@ -2,13 +2,13 @@
 # Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-add_mlir_public_c_api_library(AIRCAPI
-Dialects.cpp
-Registration.cpp
-Runner.cpp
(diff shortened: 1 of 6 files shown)


mlir/lib/CAPI/CMakeLists.txt
```diff
@@ -1,14 +1,13 @@
-# Copyright (C) 2022, Xilinx Inc. All rights reserved.
-# Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
-# SPDX-License-Identifier: MIT
+# Copyright (C) 2022, Xilinx Inc. All rights reserved. Copyright (C) 2022,
+# Advanced Micro Devices, Inc. All rights reserved. SPDX-License-Identifier: MIT
```
fifield, line 2: re-formatting error

fifield (changes requested):

mlir/lib/CAPI/CMakeLists.txt
makslevental: fixed

fifield (approved):

PR #294, opened 2023-09-04
Abide by standards for STL containers

makslevental:
This PR refactors `class runnerNode` and `class dependencyGraph` to be compliant with the standard for STL containers. In particular `std::deque` is not required to compile/support/work [with incomplete types](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4510.html). This lack of compliance prevents compiling on both Apple's clang and [MSVC](https://github.com/microsoft/STL/issues/108#issuecomment-532508302). 

It's not clear to me why `deque` was being used as no `push_front`/`pop_front` uses appear. Note a first pass/attempt at rectifying the issue used `std::shared_ptr` but this led to segfaults [here](https://github.com/makslevental/mlir-air/blob/2f2762c6f8d317970880fe1d82f9f5a652679a06/mlir/lib/Util/Runner/RunnerNode.cpp#L207-L211) probably due to the recursive call chain with [resetGraphBetweenTwoVertices](https://github.com/makslevental/mlir-air/blob/2f2762c6f8d317970880fe1d82f9f5a652679a06/mlir/lib/Util/Runner/RunnerNode.cpp#L188).

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ include(ExternalProject)
 find_package(MLIR REQUIRED CONFIG)
 find_package(AIE REQUIRED)
 find_package(LibXAIE)
+find_package(Boost REQUIRED)
 
 message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
 message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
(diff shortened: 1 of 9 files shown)


erwei-xilinx (approved):
Thanks for working on this. The use of deque was due to a legacy function call to runnerNode objects which was later removed. It is safe to replace deque with vector now.

fifield (approved):

PR #297, opened 2023-09-04
Use upstream `add_mlir_library` CMake functions

makslevental:
This PR uses the upstream CMake functions for `add_mlir_library`.

I'm not sure what the intent was of having the local `add_air_library` but it does not probably add to any export sets and thus the generated `AirConfig.cmake` is not usable for downstream builds (such as python bindings built independently of core). A first pass/attempt/solution added the correct CMake to `add_air_library` for that export set but it didn't work. Thus, in addition this patch adds a "re-export" of the MLIR export target set.

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,22 @@ foreach(target ${AIR_RUNTIME_TARGETS})
 	)
 endforeach()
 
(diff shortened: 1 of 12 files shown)


fifield (approved):
I don't recall the reason for `add_air_library`, but happy to remove it if the mlir version works better.

PR #305, opened 2023-09-26
Improve build scripts and documentation

keryell:


utils/clone-llvm.sh
```diff
@@ -14,14 +14,40 @@
...
 
-git clone --depth 1 https://github.com/llvm/llvm-project.git llvm
-pushd llvm
-git fetch --depth=1 origin $commithash
-git checkout $commithash -b $branch
+here=$PWD
+
+# Use --llvm-worktree <directory-of-local-LLVM-repo> to reuse some existing
+# local LLVM git repository
+if [ x"$1" == x--llvm-worktree ]; then
```
fifield, line 26: Personally when I have an existing llvm repo I simply don't run this script (actually I never run this script...). I wonder if this just adds complexity for a personal git preference?

fifield:
As a general comment it would be great if the documentation had more focus on how drive cmake and documenting the cmake options instead of being centered around a bunch of bespoke scripts. [Example](https://github.com/llvm/torch-mlir/blob/main/docs/development.md#cmake-build). That way people can catch their own fish.

keryell:
That is good feedback. Since these scripts were used in the CI I focused on evolving them but yes, we can avoid using them at all too when we are not in a clean slate situation.

PR #323, opened 2023-11-10
Iree fixes

stephenneuendorffer:


mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -2579,12 +2579,12 @@ class AIRToAIEPass : public AIRToAIEBase<AIRToAIEPass> {
     if (clTestPatterns.find("to-aie-mlir") != std::string::npos) {
       std::vector<std::pair<AIE::DeviceOp, air::HerdOp>> aie_modules;
       std::map<AIE::TileOp, air::HerdOp> tileToHerdMap;
-      AIRToAIEOptions options = {.col_offset = clColOffset,
-                                 .row_offset = clRowOffset,
-                                 .emit_while = clEmitWhileLoop,
-                                 .emit_herd_lock = clEmitHerdLock,
-                                 .generate_shim_dma = clGenerateShimDMA,
-                                 .device = *device};
+      AIRToAIEOptions options = {/*.col_offset = */ clColOffset,
```
keryell, line 2582: Just compile with C++20 since this is not up-streamed?

mlir/lib/Conversion/AIRLoweringPass.cpp
```diff
@@ -1300,8 +1300,10 @@ class AIRLoweringPass : public air::AIRLoweringBase<AIRLoweringPass> {
         merge_candidate_loop = true;
       else if (auto scf_par = dyn_cast<scf::ParallelOp>(scf_loop)) {
         merge_candidate_loop = false;
-        for (auto child_scf_for_loop : scf_par.getBody()->getOps<scf::ForOp>())
+        for (auto child_scf_for_loop : scf_par.getBody()->getOps<scf::ForOp>()) {
```
keryell, line 1303: Just curious about what this code does.
I guess it is not just testing if `scf_par.getBody()->getOps<scf::ForOp>()` is non empty?

stephenneuendorffer: yeah, there's a bound of nonsense code here.

mlir/lib/Conversion/AIRToAIEPass.cpp
stephenneuendorffer: Iree isn't compiling with c++20.. just looking for the simple fix here.

keryell: That is my point. When you compile IREE yourself, just add `-std=c++20`.
For all our SYCL work I had `-std=c++23` because I do not care about being compatible with old-compilers upstream.

PR #340, opened 2023-11-30
 Remove remaining use of boost graph library

newling:
This patch removes the final use of boost graph library. 

The only remaining use of boost is program options in 1 example.

### DIFF
--- a/mlir/lib/Transform/AIRDependency.cpp
+++ b/mlir/lib/Transform/AIRDependency.cpp
@@ -8,6 +8,7 @@
 #include "air/Transform/AIRDependency.h"
 #include "air/Dialect/AIR/AIRDialect.h"
 #include "air/Util/Dependency.h"
+#include "air/Util/DirectedAdjacencyMap.h"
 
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Thanks for working on this, @newling . I think both assumptions should hold true, although in the previous boost graph impl they were never checked.

erwei-xilinx:
The part which throws error is here: https://github.com/Xilinx/mlir-air/blob/c99190a1ad455b13f49c04107033aad94b7a4e46/mlir/lib/Util/Dependency.cpp#L1606 The air-runner expects a one-to-one mapping between the air.hierarchy ops (herd, segment and launch) and graphs. This error suggests that there is a mismatch.

newling:
Thanks @erwei-xilinx 

I can't figure out what logic I have changed, so will decompose this patch step-by-step (expect some small PRs tomorrow!)

newling:
I still need to implement the sub-graph logic. Will see if I can before cleaning up and making more PRs.

newling:
image

The only remaining use of boost is program options in 

```examples/air_to_ipu/test.cpp```

PR #341, opened 2023-11-30
Minor simplification to the call to transitive closure reduction

newling:
Not required, but simplifies APIs a bit by pushing variables not used globally into their local use positions. 

Part of de-boosting task (although not strictly necessary -- we can keep the current API if preferred)


erwei-xilinx (approved):
Looks great to me. Thanks!

PR #342, opened 2023-11-30
Ensure there are no parellel edges in (boost) graph 

newling:
Make sure there are no parallel (duplicate) edges in use:
1) use setS as the edge storage type instead of vecS
2) do not iterate over edges between two nodes

Technically the second item above has no effect if the first one is used, but I think it adds clarity (see add_edge method documentation in https://www.boost.org/doc/libs/1_81_0/libs/graph/doc/adjacency_list.html)

### DIFF
--- a/mlir/include/air/Util/Dependency.h
+++ b/mlir/include/air/Util/Dependency.h
@@ -138,7 +138,7 @@ struct dependencyNodeEntry {
 };
 
 // Boost dependency graph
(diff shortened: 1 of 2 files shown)


erwei-xilinx (approved):
Thanks

newling:
I merged without updating the commit message :-( 
Apologies.

PR #344, opened 2023-12-01
Remove all parts of the code which use boost sub-graphs

newling:
I implemented custom sub-graph logic which undoubtedly had bugs, but when I ran the tests they all passed. Which made me wonder if the sub-graph stuff was even tested. Turns out it isn't. So my question is: can I just remove it?


erwei-xilinx (approved):

newling:
Thanks Erwei. 

It might be good to go through the project and remove unused code at some point (might be able to use something like https://github.com/Xilinx/mlir-aie/pull/780 from @makslevental in mlir-aie)

PR #345, opened 2023-12-01
[FAILS] Store edges as a set instead of a vector 

newling:
The equivalent change worked fine for the graph class in Dependency.h (see https://github.com/Xilinx/mlir-air/pull/342) 

but with this one (AIRDependency.cpp) the change makes 5 tests fail. 

This PR changes the storage of graph edges from vector (where the same edge can be stored multiple times) to set (where each edge is unique). 

Previous discussion (see https://github.com/Xilinx/mlir-air/pull/340) suggests that it should be valid to assume edges are unique. 

So I'm not sure what to do here, I think the first thing is to understand what is failing. I don't really understand what the test (see CI lit test failure) is doing -- is the change in IR invalid or just noise? 
/
@erwei-xilinx  -- if you have any insights, please let me know!

### DIFF
--- a/mlir/lib/Transform/AIRDependency.cpp
+++ b/mlir/lib/Transform/AIRDependency.cpp
@@ -78,7 +78,7 @@ struct executeNode {
   unsigned operationId;
 };
 
-typedef boost::adjacency_list<boost::vecS, boost::vecS, boost::bidirectionalS,
+typedef boost::adjacency_list<boost::setS, boost::vecS, boost::bidirectionalS,
                               executeNode>
     Graph;
 typedef boost::graph_traits<Graph>::in_edge_iterator in_edge_iterator;


erwei-xilinx:
Hmm looking at the CI log, specifically test `Transform/AIRDependency/air_partition_dma_memcpy_nd.mlir`, it seems to suggest that there is now simply a reordering in dep list, which could be still valid dep list but not what CI previously expected...

newling:
Ok. I see 2 options:

1) I can rewrite the tests so that they pass with setS (but fail with vecS, so still brittle). This will pave the way to removing boost, but keep the test tied to the graph implementation. 

2) Someone familiar with what is being tested (I'm still in the dark here, so this might be tricky for me) can make the tests more robust (so they pass with vecS and setS), thus decoupling the tests from the graph implementation. 

I'm happy to push on with (1), although if (2) is feasible that would be a better long-term solution. 

I'm not familiar enough with the code to say whether 2 is a priority (let alone feasible). 

@erwei-xilinx please let me know what you think!

erwei-xilinx:
I think 1 would make sense, and I do not see a way to have filecheck pass with multiple patterns (and within the same line) for option 2. But perhaps people with more experience with filecheck can comment.

newling:
Is is sometimes enough to 

1) use CHECK-DAG instead of CHECK https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-dag-directive

2) CHECK only what you need 

when it is a case of the unimportant relative orders of ops. I'd need to better understand the passes to say if that's enough here though.

newling:
I'll try and push on with 1 though (rewrite the tests to pass with setS)

erwei-xilinx:
> 1. use CHECK-DAG instead of CHECK https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-dag-directive
> 2. CHECK only what you need

Yeah I know CHECK-DAG can check without any assumptions on line ordering, but I don't know if it would work if rearranged components are within the same line.

And the contents in dep list are indeed what we really need to check for, it's just that its ordering doesn't matter. I think it's quite tricky...

newling:
> > 1. use CHECK-DAG instead of CHECK https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-dag-directive
> > 2. CHECK only what you need
> 
> Yeah I know CHECK-DAG can check without any assumptions on line ordering, but I don't know if it would work if rearranged components are within the same line.
> 
> And the contents in dep list are indeed what we really need to check for, it's just that its ordering doesn't matter. I think it's quite tricky...

Ok, I hadn't actually looked closely enough as to what the diff to to the IR was, I didn't realise it was a permutation of values in a single line. 

How about this approach: https://github.com/Xilinx/mlir-air/pull/350 ? Let me know if that's overkill, and I'll just update the tests (i.e. option 1 above).

PR #349, opened 2023-12-04
Fix for more recent LLVM

newling:
This allows air to build with llvm commit on main branch:
hash: c45a66ecd4cb8f351298ca987d6086cf02b77bfb 
Date:  Fri Dec 1 17:22:12 2023 -0800

'const StringLiteral name' design copied from spriv dialect.


erwei-xilinx (approved):
Thanks for working on this!

PR #350, opened 2023-12-04
Canonicalize printing of async dependencies

newling:
With the change, the order in which the async dependencies (mlir::Values) are stored does not effect the order in which they are printed.

This makes it easier to change the implementation of the underlying dependency analysis (currently using boost) without effecting the printed IR used for testing. 

Alternatives considered:

1) just update the tests whenever the underlying algorithm changes
2) make the storage of async dependencies natively sorted 

1 seems fine, and I can still take this approach, but brittle tests might slow future development down
2 seems good, although I can't figure out how to do this (might take me more time to check this option out).

### DIFF
--- a/mlir/include/air/Dialect/AIR/AIR.td
+++ b/mlir/include/air/Dialect/AIR/AIR.td
@@ -576,9 +576,9 @@ def air_ExecuteOp : air_Op<"execute", [SingleBlockImplicitTerminator<"ExecuteTer
     the region must be executed sequentially.
   }];
 
-  // TODO(jn) async_dependencies are printed in sorted order for ops which have
-  // custom assembly formats. ExecuteOp should be made to have a custom print
-  // and parse too, so that it follows this pattern.
+  // Note: async_dependencies are printed in sorted order for ops which have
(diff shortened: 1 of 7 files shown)


erwei-xilinx (approved):
Thanks. I think this is a good idea.

stephenneuendorffer (commented):
Sorting is a good thing.  Another option would be CHECK-SAME to catch reorderings on the same line.

newling:
> Sorting is a good thing. Another option would be CHECK-SAME to catch reorderings on the same line.

I don't think CHECK-SAME can be used for permutations on a single line. Like, I'm pretty sure this fails:

// CHECK: func.func
// CHECK-SAME: beta
// CHECK-SAME: alpha
  func.func @alpha_beta() {
  ...
  
  but passes if the beta and alpha CHECK-SAME lines are permuted. 
  
Thanks for the thumbs-up. I'm mostly happy with this approach, but still think it would be better if the underlying storage was sorted.. but I don't want to go through all the build methods and check that the vector gets sorted before unique-d, and I'm not familiar enough with custom storage to invest time exploring that.

PR #351, opened 2023-12-05
A minimal graph class to replace boost equivalent

newling:
A user of mlir-air (i.e. IREE) has prioritized removing the boost dependency from mlir-air. 

A (WIP) PR which completely removes boost from mlir-air is at https://github.com/Xilinx/mlir-air/pull/340 

The PR 340 is probably too large, I'd like to break it up. This PR introduces the first part of the decomposition: a graph class which replaces boost's graph class.  

Subsequent PR(s) will use this new class to replace boost. 

This PR also introduces C++ unit tests, I couldn't find any pre-existing equivalent in mlir-air. I can add more tests, if requested.


erwei-xilinx (approved):
Thanks!

PR #358, opened 2023-12-06
Remove boost from cmake and CI

newling:
Final planned patch in boost removal.


newling:
@makslevental @erwei-xilinx boost finally goes from air with this patch. I've added suggestions for a boost_options replacement for the one example that uses (in docs).

makslevental:
Congrats @newling!

erwei-xilinx:
Thanks for working on this @newling!

PR #361, opened 2023-12-11
AIRRt-To-IPU: Unroll inner-dim repeat pattern per ipu dma instruction

erwei-xilinx:
IPU instruction forbids `stepsize=0` in DMA BDs, meaning that only the outermost dimension can perform repeat, using `repeat_count`, while inner dimension `wraps` cannot. Therefore, those repeat patterns were unrolled to get things working.


jgmelber (approved):
Test case?

erwei-xilinx:
> Test case?

Added.

PR #362, opened 2023-12-13
Bump llvm and mlir-aie

erwei-xilinx:


utils/clone-llvm.sh
```diff
@@ -14,7 +14,7 @@
 #
 ##===----------------------------------------------------------------------===##
 
-export commithash=6577f95f9fb04fb7def7a5ae78bcf5f2b4f21b6e
```
makslevental, line 17: FWIW - you could replace all of the clones and build with 

* `pip download mlir -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/mlir-distro`
* `pip download mlir-aie -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels` 

You can pin with e.g. `==18.0.0.2023120309+9584f583` too.

utils/build-llvm-local.sh
```diff
@@ -53,8 +59,17 @@ cmake ../llvm \
     -DLLVM_OPTIMIZED_TABLEGEN=OFF \
     -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
     -Dpybind11_DIR=${PYTHON_ROOT}/pybind11/share/cmake/pybind11 \
-    -DLLVM_DISTRIBUTION_COMPONENTS="cmake-exports;not;count;FileCheck;MLIRPythonModules;mlir-cpu-runner;mlir-linalg-ods-yaml-gen;mlir-opt;mlir-reduce;mlir-tblgen;mlir-translate;mlir-headers;mlir-cmake-exports" \
-    |& tee cmake.log
+    -DLLVM_DISTRIBUTION_COMPONENTS=cmake-exports;not;count;FileCheck;MLIRPythonModules;mlir-cpu-runner;mlir-linalg-ods-yaml-gen;mlir-opt;mlir-reduce;mlir-tblgen;mlir-translate;mlir-headers;mlir-cmake-exports"
```
makslevental, line 62: Just curious: what does this do?

makslevental: Oh cool it's a way to install FileCheck and stuff. Do you need all those others though?

utils/build-llvm-local.sh
```diff
@@ -27,20 +27,26 @@ PYTHON_ROOT=`pip3 show pybind11 | grep Location | awk '{print $2}'`
 
 mkdir -p $LLVM_DIR/$BUILD_DIR
 mkdir -p $LLVM_DIR/$INSTALL_DIR
+LLVM_ENABLE_RTTI=${LLVM_ENABLE_RTTI:OFF}
```
makslevental, line 30: I definitely wouldn't do this - did someone (iree folks?) ask you to turn off RTTI?

utils/clone-llvm.sh
makslevental: Note, this won't work if you decide you want to build without RTTI here.

utils/build-llvm-local.sh
erwei-xilinx: > Why disable by default? Did someone (iree folks?) ask you to turn off RTTI?

It is indeed related to interfacing with IREE, where disabling RTTI would help with delivering the end-to-end flow simpler and quicker.

utils/build-llvm-local.sh
erwei-xilinx: > Oh cool it's a way to install FileCheck and stuff. Do you need all those others though?

Those were inherited from the AIR build script and I haven't experimented by disabling some of them.

utils/clone-llvm.sh
erwei-xilinx: > Note, this won't work if you decide you want to build without RTTI here.

I would love to use the wheels as part of the build flow instead of this. In the short term having RTTI off is a requirement though...

makslevental: i can turn build `no-rtti` wheels for you guys.

erwei-xilinx: That would be super cool. Once the no-rtti wheels are ready I'll create a new PR to replace this bit.

Issue #374, opened 2024-01-04
Need to replace DimTuple


spaceotter:
I'm trying to update the iree-amd-aie project with the latest version of mlir-aie. That seems to have removed or renamed the DimTuple attr which used to be defined here https://github.com/Xilinx/mlir-aie/blob/c0341aa3d525827a21f25b7423b18f4359a34cf3/include/aie/Dialect/AIE/IR/AIEAttrs.td but is still used in this project here:  
https://github.com/Xilinx/mlir-air/blob/3ae51efcb11e9fb5504fa5851b54c0b87ea0f704/mlir/lib/Conversion/AIRToAIEPass.cpp#L2288
https://github.com/Xilinx/mlir-air/blob/3ae51efcb11e9fb5504fa5851b54c0b87ea0f704/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp#L175

erwei-xilinx:
DimTupleAttr has been removed in https://github.com/Xilinx/mlir-air/pull/383 which has been merged.

PR #370, opened 2023-12-18
Add a template workflow example for building mlir-air with wheels

erwei-xilinx:

PR #375, opened 2024-01-04
Update for newer mlir-aie

spaceotter:
Fixes https://github.com/Xilinx/mlir-air/issues/374

### DIFF
--- a/utils/clone-llvm.sh
+++ b/utils/clone-llvm.sh
@@ -14,7 +14,7 @@
 #
 ##===----------------------------------------------------------------------===##
 
-export commithash=d36b483f4f1109f53399ef82fda32f2c04d4ef44
+export commithash=5cfc7b3342ce4de0bbe182b38baa8a71fc83f8f8
 branch=air
 
 git clone --depth 1 https://github.com/llvm/llvm-project.git llvm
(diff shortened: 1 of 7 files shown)


stephenneuendorffer (approved):
Thanks!

stephenneuendorffer:
The LLVM bump will fail because scf::ReduceOp has changed slightly.

newling:
The build failure -- I don't think it's related to this PR. 

I was seeing it on my local build, and couldn't figure out what the problem was. I ended up building with wheels (@erwei-xilinx introduced this option in https://github.com/Xilinx/mlir-air/pull/370 and it worked for me)

image

utils/clone-llvm.sh
```diff
@@ -14,7 +14,7 @@
 #
 ##===----------------------------------------------------------------------===##
 
-export commithash=d36b483f4f1109f53399ef82fda32f2c04d4ef44
+export commithash=5cfc7b3342ce4de0bbe182b38baa8a71fc83f8f8
 branch=air
```
newling, line 18: These llvm and aie hashes also appear in the script build-mlir-air-using-wheels.sh

https://github.com/Xilinx/mlir-air/blob/3ae51efcb11e9fb5504fa5851b54c0b87ea0f704/utils/build-mlir-air-using-wheels.sh#L36

That script was added recently in https://github.com/Xilinx/mlir-air/pull/370 , this is first bump since it was added.

spaceotter: There's no mlir-aie wheel for 652d0f72.

PR #377, opened 2024-01-08
Fixing build-mlir-air.sh to point to hsa.h and adding some clarifications to the documentation

eddierichter-amd:


erwei-xilinx (approved):
Looks great to me. Thanks.

PR #378, opened 2024-01-09
Fix build-mlir-air.sh script

newling:
@eddierichter-amd what are the 2 HSA directories, can it be 1 directory?

### DIFF
--- a/utils/build-mlir-air.sh
+++ b/utils/build-mlir-air.sh
@@ -31,9 +31,9 @@
 #
 ##===----------------------------------------------------------------------===##
 
-if [ "$#" -lt 4 ]; then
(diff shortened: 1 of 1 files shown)


eddierichter-amd (approved):

eddierichter-amd:
Thanks for catching my typo! 

As per the two directories, it is the hsa runtime `${ROCM_ROOT}/lib/cmake/hsa-runtime64/` and the hsa kmt `${ROCM_ROOT}/lib/cmake/hsakmt/` Yeah I think just passing in ${ROCM_ROOT} and the script filling in the gaps is a good idea.

newling:
Thanks for info @eddierichter-amd . Closing this because of your most recent PR.

PR #379, opened 2024-01-09
bump the LLVM used by mlir-air

newling:
This focuses on the LLVM bump. Mostly related to the SCF ReduceOp changes in https://github.com/llvm/llvm-project/pull/75314

As a follow-up we can bump the mlir-aie used. I just first want to see if this PR actually works (I haven't built and tested succesfully locally).

### DIFF
--- a/mlir/include/air/Conversion/PassDetail.h
+++ b/mlir/include/air/Conversion/PassDetail.h
@@ -10,6 +10,8 @@
 #define AIR_CONVERSION_PASSDETAIL_H
 
 #include "aie/Dialect/AIEX/IR/AIEXDialect.h"
+#include "mlir/Pass/PassOptions.h"
+#include "mlir/Pass/Pass.h"
 
 namespace xilinx {
 namespace air {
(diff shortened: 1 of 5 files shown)


newling:
In [utils/clone-llvm.sh](https://github.com/Xilinx/mlir-air/pull/379/files#diff-f10634e37347e39187b7d370c82c420c23fc2b794487dbbe17079ee8b7124d69) the branch used is the 'air' branch. Why is this? When I bump, can I use main?

fifield:
I suggest bumping mlir-aie at the same time and using the LLVM hash here: https://github.com/Xilinx/mlir-aie/blob/main/utils/clone-llvm.sh#L16.  The llvm version for mlir-aie and mlir-air should match.

fifield:
> In [utils/clone-llvm.sh](https://github.com/Xilinx/mlir-air/pull/379/files#diff-f10634e37347e39187b7d370c82c420c23fc2b794487dbbe17079ee8b7124d69) the branch used is the 'air' branch. Why is this? When I bump, can I use main?

It's naming the local branch `air` so the name is not too important here.

PR #380, opened 2024-01-09
Fixing the compiler only flow

eddierichter-amd:
This PR includes CMake changes to allow building just the AIR compiler without any runtime dependencies. Before this, the compiler only flow (`build-mlir-air.sh`) required the runtime headers, which became more cumbersome when our compiler runtime required the ROCm runtime (ROCr) as a dependency.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -65,7 +65,7 @@ declare_mlir_python_extension(AirPythonExtensions.MLIR
 
 # Only building this if we are building the runtime, as it requires
 # the headers from the runtime
-if(AIR_RUNTIME_TEST_TARGET_VAL)
+if(AIR_RUNTIME_TARGETS)
   message("Building AirHostModule.cpp because we are building the runtime")
   declare_mlir_python_extension(AirPythonExtensions.AIRRt
     MODULE_NAME _airRt
@@ -123,7 +123,7 @@ include_directories(
 )
 
 # Only include this if we are building the runtime
-if(AIR_RUNTIME_TEST_TARGET_VAL)
+if(AIR_RUNTIME_TARGETS)
   include_directories(
     ${hsa-runtime64_DIR}/../../../include
   )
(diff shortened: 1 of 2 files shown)


python/CMakeLists.txt
```diff
@@ -63,17 +63,22 @@ declare_mlir_python_extension(AirPythonExtensions.MLIR
...
-    AirHostModule.cpp
-  EMBED_CAPI_LINK_LIBS
-    AIRCAPI
-  PRIVATE_LINK_LIBS
-    LLVMSupport
-)
+# Only building this if we are building the runtime, as it requires
+# the headers from the runtime
+if(AIR_RUNTIME_TEST_TARGET_VAL)
+  message("Building AirHostModule.cpp because we are building the runtime")
```
newling, line 69:
```suggestion
  message(STATUS "Building python extension because we are building the runtime")
```

newling (approved):
Thanks!

python/CMakeLists.txt
```diff
@@ -63,17 +63,22 @@ declare_mlir_python_extension(AirPythonExtensions.MLIR
...
-  SOURCES
-    AirHostModule.cpp
-  EMBED_CAPI_LINK_LIBS
-    AIRCAPI
-  PRIVATE_LINK_LIBS
-    LLVMSupport
-)
+# Only building this if we are building the runtime, as it requires
+# the headers from the runtime
+if(AIR_RUNTIME_TEST_TARGET_VAL)
```
fifield, line 68: It's going to work the same either way but `AIR_RUNTIME_TARGETS` is what determines if the runtime is built.

fifield (approved):

python/CMakeLists.txt
eddierichter-amd: Thanks for pointing that out. Addressed in [24ea273](https://github.com/Xilinx/mlir-air/pull/380/commits/24ea2732edf24120d953affbbe585025173f59e4).

fifield:
It's unfortunate that the aircpu library is not built in the "compiler only" mode, since it is for cpu execution of air programs without any device or runtime. But we can address that in a different PR.

PR #389, opened 2024-01-18
Canonicalize wrap and stride lists, to avoid BD dim limitation when possible

erwei-xilinx:


mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
```diff
@@ -1411,6 +1475,8 @@ struct AIRSpecializeChannelWrapAndStrideInScfFor
         rewriter, for_op.getOperation(), channel_ops[0].getOperation(), offsets,
         wraps, strides, channel_ops[0].getMemref());
 
+    (void)AIRCanonicalizeWrapAndStrideList(rewriter, offsets, wraps, strides);
```
fifield, line 1478: This causes your previously working end to end example to segfault.

erwei-xilinx: This method itself isn't the problem. It creates the scenario where the # dims in `offsets` are higher than the rank of memref type. This patch should fix the issue with calculating offset: https://github.com/Xilinx/mlir-air/pull/394

PR #393, opened 2024-01-22
Add -air-fuse-channels pass to aircc.py

erwei-xilinx:
- In a previous patch, the DMA channel sharing capability of -air-to-aie pass has been taken out and reimplemented with a new pass `-air-fuse-channels`.
- If aircc.py doesn't contain `-air-fuse-channels`, it will not properly handle cases where more than two memcpys go into one single tile.

### DIFF
--- a/python/air/compiler/aircc/main.py
+++ b/python/air/compiler/aircc/main.py
@@ -116,6 +116,7 @@ def run(mlir_module, args=None):
     air_placed = opts.tmpdir+'/placed.'+air_mlir_filename
     pass_pipeline = ','.join([
       'air-pipeline-to-affine{lowering-type=getput}',
+      'air-fuse-channels',
       'canonicalize', 'cse',
       'func.func(air-renumber-dma)',
       'func.func(convert-linalg-to-loops)',


fifield:
> * In a previous patch, the DMA channel sharing capability of -air-to-aie pass has been taken out and reimplemented with a new pass `-air-fuse-channels`.
> * If aircc.py doesn't contain `-air-fuse-channels`, it will not properly handle cases where more than two memcpys go into one single tile.

which tests were breaking without this patch?

erwei-xilinx:
> which tests were breaking without this patch?

48 and 49 which were flagged with XFAIL.

fifield:
I am getting mismatch in tests 44 and 46 on vck5000, so I tried applying this patch. After applying this patch aircc.py started to hang.

erwei-xilinx:
> I am getting mismatch in tests 44 and 46 on vck5000, so I tried applying this patch. After applying this patch aircc.py started to hang.

Ok let me look into this.

fifield:
> > I am getting mismatch in tests 44 and 46 on vck5000, so I tried applying this patch. After applying this patch aircc.py started to hang.
> 
> Ok let me look into this.

The mismatch appears to be peano related. No mismatch with chess.

erwei-xilinx:
> After applying this patch aircc.py started to hang.

This issue is fixed with this patch: https://github.com/Xilinx/mlir-air/pull/459

Tests 44 and 46 do not have air.channels, so `-air-fuse-channels` requires `-air-dma-to-channel` to work. Without `-air-fuse-channels` and `-air-dma-to-channel`, we currently do not have a way to directly fuse `air.dma` ops to share DMA channels.

fifield:
> > After applying this patch aircc.py started to hang.
> 
> This issue is fixed with this patch: #459
> 
> Tests 44 and 46 do not have air.channels, so `-air-fuse-channels` requires `-air-dma-to-channel` to work. Without `-air-fuse-channels` and `-air-dma-to-channel`, we currently do not have a way to directly fuse `air.dma` ops to share DMA channels.

`air-dma-to-channel` will get added to aircc in https://github.com/Xilinx/mlir-air/pull/454.  I am still cleaning up some issues in the vck5000 tests, but the plan is to make it safe to always run `air-dma-to-channel` with that PR.

erwei-xilinx:
Air-dma-to-channel now part of the experimental_passes in aircc.py.

PR #395, opened 2024-01-23
Fold a chain of memref.subview/expand_shape/transpose ops to wraps and strides in air.dma

erwei-xilinx:
- Add a pass at the end of -air-copy-to-dma to fold a chain of memref data rearrangement ops to wraps and strides.
- As a continuation of the lowering of IREE pack/unpack through AIR: https://github.com/nod-ai/iree-amd-aie/pull/61

### DIFF
--- a/mlir/lib/Conversion/ConvertToAIRPass.cpp
+++ b/mlir/lib/Conversion/ConvertToAIRPass.cpp
@@ -2318,6 +2318,267 @@ private:
   // bool generateSegment;
 };
 
+/// Build a strided memref type by applying `permutationMap` tp `memRefType`.
+static MemRefType inferTransposeResultType(MemRefType memRefType,
+                                           AffineMap permutationMap) {
+  auto rank = memRefType.getRank();
+  auto originalSizes = memRefType.getShape();
+  auto [originalStrides, offset] = getStridesAndOffset(memRefType);
+  assert(originalStrides.size() == static_cast<unsigned>(rank));
+
+  // Compute permuted sizes and strides.
+  SmallVector<int64_t> sizes(rank, 0);
+  SmallVector<int64_t> strides(rank, 1);
+  for (const auto &en : llvm::enumerate(permutationMap.getResults())) {
+    unsigned position = cast<AffineDimExpr>(en.value()).getPosition();
+    sizes[en.index()] = originalSizes[position];
+    strides[en.index()] = originalStrides[position];
+  }
+
+  return MemRefType::Builder(memRefType)
+      .setShape(sizes)
+      .setLayout(
+          StridedLayoutAttr::get(memRefType.getContext(), offset, strides));
+}
+
+static void extractStridesFromMemrefType(MemRefType memrefTy,
+                                         OpBuilder &builder,
+                                         SmallVector<Value, 4> &strides) {
+  // get the strides and offsets from the memref type
+  int64_t offset;
+  SmallVector<int64_t, 4> layout_strides;
+  auto successStrides = getStridesAndOffset(memrefTy, layout_strides, offset);
+  if (failed(successStrides)) {
+    llvm::outs() << "Failed to get strides\n";
+    return;
+  }
+
+  for (auto s : layout_strides)
+    strides.push_back(
+        builder.create<arith::ConstantIndexOp>(builder.getUnknownLoc(), s));
+}
+
+static void extractSizesFromMemrefType(MemRefType memrefTy, OpBuilder &builder,
+                                       SmallVector<Value, 4> &sizes) {
+  for (auto s : memrefTy.getShape())
+    sizes.push_back(
+        builder.create<arith::ConstantIndexOp>(builder.getUnknownLoc(), s));
+}
+
+static void extractOffsetsFromSubview(memref::SubViewOp subview,
+                                      OpBuilder &builder,
+                                      SmallVector<Value, 4> &offsets) {
+  auto subview_offsets = subview.getOffsets().begin();
+  auto static_offsets = subview.getStaticOffsets();
+  auto loc = subview.getLoc();
+
+  for (auto o : static_offsets) {
+    if (o >= 0)
+      offsets.push_back(builder.create<arith::ConstantIndexOp>(loc, o));
+    else
+      offsets.push_back(*subview_offsets++);
+  }
+}
+
+static LogicalResult canonicalizeAIRDmaOperands(OpBuilder builder,
+                                                SmallVector<Value, 4> &offsets,
+                                                SmallVector<Value, 4> &sizes,
+                                                SmallVector<Value, 4> &strides,
+                                                MemRefType memref) {
+  // Increase vector sizes up to memref size
+  int max_dim_size =
+      std::max(std::max(offsets.size(), sizes.size()), strides.size());
+  if (max_dim_size && offsets.size() < memref.getRank()) {
+    for (unsigned i = offsets.size(); i < memref.getRank(); i++) {
+      offsets.insert(offsets.begin(), builder.create<arith::ConstantIndexOp>(
+                                          builder.getUnknownLoc(), 0));
+    }
+  }
+  if (max_dim_size && sizes.size() < memref.getRank()) {
+    for (unsigned i = sizes.size(); i < memref.getRank(); i++) {
+      sizes.insert(sizes.begin(), builder.create<arith::ConstantIndexOp>(
+                                      builder.getUnknownLoc(), 1));
+    }
+  }
+  int memref_size = 1;
+  for (auto size : memref.getShape())
+    memref_size *= size;
+  if (max_dim_size && strides.size() < memref.getRank()) {
+    for (unsigned i = strides.size(); i < memref.getRank(); i++) {
+      strides.insert(strides.begin(),
+                     builder.create<arith::ConstantIndexOp>(
+                         builder.getUnknownLoc(), memref_size));
+    }
+  }
+
+  // Reduce highest dimensions if more than memref size
+  while (strides.size() > memref.getRank() && getConstantIntValue(strides[0]) &&
+         *getConstantIntValue(strides[0]) == memref_size) {
+    strides.erase(strides.begin());
+  }
+  while (sizes.size() > memref.getRank() && getConstantIntValue(sizes[0]) &&
+         *getConstantIntValue(sizes[0]) == 1) {
+    sizes.erase(sizes.begin());
+  }
+  while (offsets.size() > std::min(sizes.size(), strides.size()) &&
+         getConstantIntValue(offsets[0]) &&
+         *getConstantIntValue(offsets[0]) == 0) {
+    offsets.erase(offsets.begin());
+  }
+
+  if (offsets.size() != sizes.size() || sizes.size() != strides.size())
+    return failure();
+
+  return success();
+}
+
+static LogicalResult CondenseMemrefDataReorderingToAIRDma(
+    air::DmaMemcpyNdOp dmaOp, std::vector<Operation *> src_ancestor_memref_ops,
+    std::vector<Operation *> dst_ancestor_memref_ops) {
+  OpBuilder rewriter(dmaOp);
+  auto src = dmaOp.getSrcMemref();
+  auto dst = dmaOp.getDstMemref();
+  auto loc = dmaOp->getLoc();
+
+  // It must already be a memref
+  auto src_type = src.getType().dyn_cast<MemRefType>();
+  auto dst_type = dst.getType().dyn_cast<MemRefType>();
+  if (!src_type)
+    return failure();
+  if (!(src_type.hasStaticShape() || dst_type.hasStaticShape()))
+    return failure();
+
+  // Revert the vector of memref ops, as it was built with push_back.
+  std::reverse(src_ancestor_memref_ops.begin(), src_ancestor_memref_ops.end());
+  std::reverse(dst_ancestor_memref_ops.begin(), dst_ancestor_memref_ops.end());
+
+  SmallVector<Value, 4> src_offsets, dst_offsets;
+  SmallVector<Value, 4> src_strides, dst_strides;
+  SmallVector<Value, 4> src_sizes, dst_sizes;
+  SmallVector<Value, 4> empty;
+
+  MemRefType src_memref_ty;
+  if (!src_ancestor_memref_ops.empty()) {
+    if (auto subviewOp =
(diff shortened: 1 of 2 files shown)


mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+
+static void extractOffsetsFromSubview(memref::SubViewOp subview,
+                                      OpBuilder &builder,
+                                      SmallVector<Value, 4> &offsets) {
+  auto subview_offsets = subview.getOffsets().begin();
+  auto static_offsets = subview.getStaticOffsets();
+  auto loc = subview.getLoc();
+
+  for (auto o : static_offsets) {
+    if (o >= 0)
```
newling, line 2376: Are you checking for the ShapedType::kDynamic value here? https://mlir.llvm.org/docs/Dialects/MemRef/#memrefsubview-memrefsubviewop
Please add a comment in the code :)

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+      offsets.push_back(builder.create<arith::ConstantIndexOp>(loc, o));
```
newling, line 2377: nit: If you wanted to have a location for all the ConstantIndexOps created in this PR (which is almost certainly not necessary..) you could consider passing them in as arguments (and use FusedLoc if there are multiple source locations).

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+  }
+}
+
+static LogicalResult canonicalizeAIRDmaOperands(OpBuilder builder,
+                                                SmallVector<Value, 4> &offsets,
+                                                SmallVector<Value, 4> &sizes,
+                                                SmallVector<Value, 4> &strides,
+                                                MemRefType memref) {
+  // Increase vector sizes up to memref size
+  int max_dim_size =
```
newling, line 2389: Why is the case where they're all empty handled differently, and when does this happen? Please add comment in code.

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+    }
+  }
+  int memref_size = 1;
+  for (auto size : memref.getShape())
+    memref_size *= size;
+  if (max_dim_size && strides.size() < memref.getRank()) {
+    for (unsigned i = strides.size(); i < memref.getRank(); i++) {
+      strides.insert(strides.begin(),
+                     builder.create<arith::ConstantIndexOp>(
+                         builder.getUnknownLoc(), memref_size));
```
newling, line 2410: So all dimensions have a stride which is the full size of the tensor? I would have imagined some kind of dimension specific stride (product of all following dimension size). 

Does memref have a static shape (are all elements of memref.getShape() positive)? I'm just trying to connect this to the check for strictly positive values in the static_offsets earlier.

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+    offsets.erase(offsets.begin());
+  }
+
+  if (offsets.size() != sizes.size() || sizes.size() != strides.size())
+    return failure();
+
+  return success();
+}
+
+static LogicalResult CondenseMemrefDataReorderingToAIRDma(
```
newling, line 2435: nit: choice of Condense vs condense? (upper case for first letter of function)

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2389,6 +2650,87 @@ struct CopyToDmaPass : public air::impl::CopyToDmaBase<CopyToDmaPass> {
...
+        src_condense |= isa<memref::SubViewOp>(src_defop);
+      }
+      bool dst_condense = false;
+      if (auto dst_defop = dmaOp.getDstMemref().getDefiningOp()) {
+        dst_condense |= isa<memref::TransposeOp>(dst_defop);
+        dst_condense |= isa<memref::ExpandShapeOp>(dst_defop);
+        dst_condense |= isa<memref::SubViewOp>(dst_defop);
+      }
+      if (src_condense || dst_condense) {
+        std::tuple<air::DmaMemcpyNdOp, std::vector<Operation *>,
```
newling, line 2676: nit: Please add a comment on what the 3 fields of the tuple are.

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2389,6 +2650,87 @@ struct CopyToDmaPass : public air::impl::CopyToDmaBase<CopyToDmaPass> {
...
+                   std::vector<Operation *>>
+            log_entry;
+        std::get<0>(log_entry) = dmaOp;
+        if (src_condense) {
```
newling, line 2680: nit: I think the number of lines can be reduced by not duplicating the code for src and dst cases (and less code is generally good :))

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+
+  if (failed(canonicalizeAIRDmaOperands(rewriter, src_offsets, src_sizes,
+                                        src_strides,
+                                        src.getType().cast<MemRefType>())) ||
+      failed(canonicalizeAIRDmaOperands(rewriter, dst_offsets, dst_sizes,
+                                        dst_strides,
+                                        dst.getType().cast<MemRefType>()))) {
+    assert(false);
+  }
+  auto new_dma = rewriter.create<xilinx::air::DmaMemcpyNdOp>(
```
newling, line 2570: Can you please explain what the new source and destination tensors are? 

I would have expected them to be different to the old DmaMemcpyNdOp (as you've unwound through view operations). 

i.e. it seems to me like you've updated the DmMemcpy attributes to reflect the preceding view operations, but the view operations are still being applied (so effectively applied twice now). I'm probably missing something

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2389,6 +2650,87 @@ struct CopyToDmaPass : public air::impl::CopyToDmaBase<CopyToDmaPass> {
...
+                   std::vector<Operation *>>
+            log_entry;
+        std::get<0>(log_entry) = dmaOp;
+        if (src_condense) {
+          Operation *ancestor = dmaOp.getSrcMemref().getDefiningOp();
+          bool exit = false;
+          while (ancestor && !exit) {
+            if (auto transpose_anc = dyn_cast<memref::TransposeOp>(ancestor)) {
+              std::get<1>(log_entry).push_back(ancestor);
+              ancestor = transpose_anc.getIn().getDefiningOp();
```
newling, line 2686: What if ancestor has more than one consumer -- just thinking ahead here, in case the ancestor is removed, which seems to be what happens at the end of this function.

mlir/test/Conversion/ConvertToAIR/condense_memref_ops_to_air_memcpy.mlir
```diff
@@ -0,0 +1,92 @@
+//===- condense_memref_ops_to_air_memcpy.mlir ------------------*- MLIR -*-===//
+//
+// Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: air-opt %s -air-copy-to-dma -canonicalize -cse | FileCheck %s
+
```
newling, line 9: Would it be possible to split this into smaller tests (check the canonicalization of a single dma_memcpy at a time) ?

newling (commented):
Thanks @erwei-xilinx 

Please feel free to ignore my 'nit' comments. I reviewed this quite thoroughly, which is possibly not appropriate for this PR.

mlir/test/Conversion/ConvertToAIR/condense_memref_ops_to_air_memcpy.mlir
```diff
@@ -0,0 +1,92 @@
+//===- condense_memref_ops_to_air_memcpy.mlir ------------------*- MLIR -*-===//
+//
+// Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
```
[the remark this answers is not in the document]
erwei-xilinx: Thanks for the comment. 

`-canonicalize` in this test removes duplicates of `arith.constant` and could make the tests a lot shorter. Checking the entries of wraps and strides is needed in this test, and those are SSA values returned from `arith.constant` ops. If without `-canonicalize` then the FileCheck checks shall go over 100 lines...

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+
+static void extractStridesFromMemrefType(MemRefType memrefTy,
+                                         OpBuilder &builder,
+                                         SmallVector<Value, 4> &strides) {
+  // get the strides and offsets from the memref type
+  int64_t offset;
+  SmallVector<int64_t, 4> layout_strides;
+  auto successStrides = getStridesAndOffset(memrefTy, layout_strides, offset);
+  if (failed(successStrides)) {
+    llvm::outs() << "Failed to get strides\n";
```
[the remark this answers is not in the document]
erwei-xilinx: You are right I forgot to change it back after debugging. Fixed in an incoming commit. Thanks.

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -2318,6 +2318,267 @@ class ScfForallToLaunchConversion : public OpRewritePattern<scf::ForallOp> {
...
+    strides[en.index()] = originalStrides[position];
+  }
+
+  return MemRefType::Builder(memRefType)
+      .setShape(sizes)
+      .setLayout(
+          StridedLayoutAttr::get(memRefType.getContext(), offset, strides));
+}
+
+static void extractStridesFromMemrefType(MemRefType memrefTy,
```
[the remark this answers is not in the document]
erwei-xilinx: Make sense. Will update in an incoming commit.

mlir/lib/Conversion/ConvertToAIRPass.cpp
erwei-xilinx: Added a comment.

mlir/lib/Conversion/ConvertToAIRPass.cpp
erwei-xilinx: At the moment AIR assumes that all strides (one per dimension) are static and that the memref has a static shape. This is to be revised later when an application drives the need for dynamic shapes.

mlir/lib/Conversion/ConvertToAIRPass.cpp
erwei-xilinx: Typo. Thanks for pointing out.

mlir/lib/Conversion/ConvertToAIRPass.cpp
erwei-xilinx: Comment added.

mlir/lib/Conversion/ConvertToAIRPass.cpp
erwei-xilinx: The original memcpy op has src and dst memrefs being those returned at the end of chains of memref ops. The src and dst memrefs of the new memcpy op have been changed to the starting memrefs of each chain, where the chain of memref ops are condensed into sizes and strides list and applied at the two sides of the new memcpy. The chain of memref ops are then removed.

mlir/lib/Conversion/ConvertToAIRPass.cpp
erwei-xilinx: This is a great point. I changed it to not delete the chain of memref ops, and instead rely on the `-canonicalize` to remove the ops for me. Thanks for pointing it out.

mlir/lib/Conversion/ConvertToAIRPass.cpp
newling: Ok good. I can see that now (I somehow missed the use of dst_ancestor_memref_ops[0] and src_ancestor_memref_ops[0] when I looked earlier).

mlir/test/Conversion/ConvertToAIR/condense_memref_ops_to_air_memcpy.mlir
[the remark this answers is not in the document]
newling: Ok, I understand. I wonder, if you ran constant subexpression elimination then you wouldn't need to check every line because the arith.constants would all be unique.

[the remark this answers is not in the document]
erwei-xilinx: No because the air.launch air.segment and air.herd ops have `isolateFromAbove` interface, so the constant ops are not really unique across those ops...

mlir/lib/Conversion/ConvertToAIRPass.cpp
newling: Ok. And as for setting all strides to the number of elements in memref, is that because these are 'leading' dimensions?

erwei-xilinx: I remember I did that to simply get around a constraint in a downstream AIR pass which complains when the # stride dimensions < memref rank... But I think one bugfix that I did yesterday might have lifted that constraint... Might need to double check.

Anyway, I don't think there's a deep reason for doing that, apart from getting around some engineering constraints...

newling (approved):
Remaining comments are all stylistic and optional. Logic makes sense to me.

PR #399, opened 2024-01-25
Channel to object FIFO: 1-to-1 L2 support

abisca:
This PR introduces 1-to-1 L2 support in the channel to object FIFO lowering. Specifically, it leverages the existing MemTile allocator to allocate L2 buffers to MemTiles. That information is stored in a map and forwarded to the object FIFO lowering pass which then uses it to generate objectFifos with MemTiles and LinkOps, where applicable (i.e., if the L2 result from a get from one channel is put into another channel). The allocated L2 buffers are erased as object FIFOs allocate their own.

Two examples are part of this PR: one with channel-to-objectFifo lowering of 1-to-1 channels with L3-L2-L1 data movement, and the other with channels using broadcast and L3-L2-L1 data movement.

Support for join and distribute patterns will be part of a subsequent PR.


.gitignore
```diff
@@ -17,6 +17,7 @@ _x
 vck5000_aie_base_graph_hw.ltx
 vck5000_aie_base_graph_hw.xsa.link_summary
 Work
+sandbox
```
erwei-xilinx, line 20: Would be great to remove this, as virtualenv name doesn't have to be called "sandbox".

abisca: Ah yes, good point. I will remove it!

jgmelber (approved):
LGTM

mlir/test/Conversion/AIRToAIE/air_channel_to_objectfifo_L1toL2.mlir
```diff
@@ -0,0 +1,74 @@
+//===- air_channel_to_objectfifo_L1toL2.mlir --------------------*- MLIR -*-===//
+//
+// Copyright (C) 2022, Advanced Micro Devices, Inc.
```
fifield, line 3: 2024

mlir/test/Conversion/AIRToAIE/air_channel_to_objectfifo_L1toL2.mlir
```diff
@@ -0,0 +1,74 @@
...
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: air-opt %s -air-place-herds='num-rows=2 num-cols=2 row-anchor=3 col-anchor=5' --air-to-aie='emit-while-loop=false row-offset=3 col-offset=5 use-objectfifo=true device=xcve2802' --canonicalize | FileCheck %s
```
fifield, line 8: Can this command line be simplified to only `air-to-aie`?  `emit-while-loop=false` is already the default

mlir/test/Conversion/AIRToAIE/air_channel_to_objectfifo_L2_broadcast.mlir
```diff
@@ -0,0 +1,83 @@
+//===- air_channel_to_objectfifo_L2_broadcast.mlir --------------------*- MLIR -*-===//
+//
+// Copyright (C) 2022, Advanced Micro Devices, Inc.
```
fifield, line 3: 2024

mlir/test/Conversion/AIRToAIE/air_channel_to_objectfifo_L2_broadcast.mlir
```diff
@@ -0,0 +1,83 @@
...
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: air-opt %s -air-place-herds='num-rows=2 num-cols=2 row-anchor=3 col-anchor=5' --air-to-aie='emit-while-loop=false row-offset=3 col-offset=5 use-objectfifo=true device=xcve2802' --canonicalize | FileCheck %s
```
fifield, line 8: Can this command line be simplified to only `air-to-aie`?

fifield (approved):

mlir/test/Conversion/AIRToAIE/air_channel_to_objectfifo_L2_broadcast.mlir
AndraBisca: Noted, thank you Jeff!

mlir/test/Conversion/AIRToAIE/air_channel_to_objectfifo_L1toL2.mlir
AndraBisca: Good point. I'll take a look and simplify.

PR #409, opened 2024-02-05
Affine loop tiling pass

erwei-xilinx:
- Intended to tile airrt shim dma ops, to generate ipu.sync op at the end of each tile to reconfigure shim dma bds.
- Added a registration template to assist with IREE integration.


fifield:
This reminds me of the functionality in mlir/lib/Transform/AffineLoopOptPass.cpp, which nobody has used or touched in _years_.

PR #411, opened 2024-02-07
non-const static variable -> class member for a thread safe pass

newling:
This PR ensures (for my build setup) that compile pipeline 2 with multithreading does not segfault (I ran ~10 times)

### DIFF
--- a/mlir/lib/Transform/AIRDependency.cpp
+++ b/mlir/lib/Transform/AIRDependency.cpp
@@ -92,7 +92,7 @@ static uint64_t WaitAllOpID;
 static uint64_t ChannelOpID;
 
 namespace {
-  std::mutex mtx;
+std::mutex mtx;
 }
 
 class AIRDependency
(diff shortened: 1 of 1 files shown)


nirvedhmeshram (commented):
I think the problem is that this pass is changing state outside its own module, so I am curious what that state is and if we can make it not do that, we would ideally want to allow multiple threads to use the pass together if we can fix the root issue.

newling:
> I think the problem is that this pass is changing state outside its own module, so I am curious what that state is and if we can make it not do that, we would ideally want to allow multiple threads to use the pass together if we can fix the root issue.

I think the problem is that the pass has (C++) state, which gets read/written by 2 threads.

nirvedhmeshram:
> > I think the problem is that this pass is changing state outside its own module, so I am curious what that state is and if we can make it not do that, we would ideally want to allow multiple threads to use the pass together if we can fix the root issue.
> 
> I think the problem is that the pass class has (C++) state, which gets read/written by 2 threads.

I think C++ state is fairly standard in passes? Just giving a random example from IREE (choose a module pass as well)
https://github.com/openxla/iree/blob/main/compiler/src/iree/compiler/Codegen/LLVMCPU/ConvertToLLVM.cpp

newling:
From  https://mlir.llvm.org/docs/PassManagement : passes "Must not maintain mutable pass state across invocations of runOnOperation". A single pass object runs on multiple operations at the same time, makes sense that of the class has mutable state then bad things might happen.

nirvedhmeshram (approved):
Thanks, LGTM!

erwei-xilinx (approved):
Thanks for working on this, LGTM!

PR #415, opened 2024-02-08
Add a pattern to hoist air.herd out of parent perfectly nested loops

erwei-xilinx:
- In support for K-dimension scaling of GEMMs.


fifield:
> I'm curious of the use case here vs. permuting the loops differently during tiling (before mapping loop to herd) ?

I guess it it because you are doing a multi-level tiling, and you want to push the reduction loop from the next level up into the herd. I think you can often get away with `emit-while-loop=true` if the herd is perfectly nested within the loop?

PR #416, opened 2024-02-08
Repair herd to ipu lowering and add e2e test

fifield:
* restore ability to lower `air.herd` to ipu without `air.launch` or `air.segment`
* add eltwise mul e2e test: one core; simple tiling from L3 to L1; no `air.launch` or `air.segment`


erwei-xilinx (approved):
LGTM!

PR #410, opened 2024-02-06
Thread sanitizer (tsan) fixes

newling:
See issue https://github.com/nod-ai/iree-amd-aie/issues/130 for reproducer

Note that this PR does NOT fix https://github.com/nod-ai/iree-amd-aie/issues/130 just the issues detected with tsan

Steps taken:

1) build iree-amd-aie with tsan (thread sanitizer) enabled. This requires setting IREE_ENABLE_TSAN:BOOL=ON,  IREE_BYTECODE_MODULE_ENABLE_TSAN:BOOL=ON, and IREE_BYTECODE_MODULE_FORCE_LLVM_SYSTEM_LINKER:BOOL=ON . At least that worked for me. 

2) run the reproducer (see "compile pipeline 2" in https://github.com/nod-ai/iree-amd-aie/issues/130 )

3) observe the warnings that tsan throws up and fix them. In total there were 3 warnings, which this PR fixes. With this done, tsan produces no more warnings with "compile pipeline 2". Good, but the UB remains. 

Warnings were:

```
SUMMARY: ThreadSanitizer: data race /proj/gdba/jamesn/workspace/iree-amd-aie/third_party/mlir-air/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp:124:29 in xilinx::air::generateBufferNameInStringStream(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, unsigned long&, mlir::StringAttr, int, int)
```

```
SUMMARY: ThreadSanitizer: heap-use-after-free /proj/gdba/jamesn/workspace/iree-amd-aie/third_party/mlir-air/mlir/lib/Conversion/ConvertToAIRPass.cpp:1140:19 in (anonymous namespace)::AIRDmaToAIRChannelConversion::matchAndRewrite(xilinx::air::DmaMemcpyNdOp, mlir::PatternRewriter&) const
```

and another data race warning.

PR #417, opened 2024-02-08
Remove static BufferId

newling:
This fixes the tsan errors in the "correct" way, and undoes the previous changes in https://github.com/Xilinx/mlir-air/pull/410

The code in this PR is not pretty, I can try and 

1) be consistent with camelCase vs CamelCase
2) remove the "const uint64_t dummy" hacks, which were required to keep the 'getBuffer' APIs aligned as they're used in a template function. I can use template SFINAE to make this better (maybe...)
3) Pass the parameter BufferId later in the function signature (I've stolen position #0 to make it easier to deal with methods with default parameters) 

The thing is that each instance of the AIRToAIEPass must have its own version of BufferID, otherwise there are multithreading issues.


newling:
@nirvedh I can't add you as an air reviewer, but this is the "revert" of the mutex approach

erwei-xilinx (approved):
LGTM!

PR #424, opened 2024-02-13
update mlir-aie

fifield:
update mlir-aie to 677369a
update llvm to b45de48

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -2340,13 +2340,14 @@ public:
         end_bb = new Block();
         mem.getBody().push_back(end_bb);
         auto b = OpBuilder::atBlockBegin(channel_head);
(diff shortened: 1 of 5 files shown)


fifield:
https://github.com/llvm/llvm-project/pull/80590 makes it mandatory for a symbol op to have a symbol table parent op. This breaks `air.channel` defined inside `func.func`, `air.launch` or `air.segment`.

fifield:
The workaround here is to make `air.channel` NOT a symbol. Just fixing the tests to define `air.channel` in the right places seems to break `air-dependency`.  Making `air.segment` and `air.launch` into symbol table ops so they can nest `air.channel` seems to additionally break `air-fuse-channels`. 
@erwei-xilinx FYI

fifield:
> The workaround here is to make `air.channel` NOT a symbol. 

I've reverted this to make channels symbols again, but still get failure in air dependency pass, in test mlir/test/Transform/AIRDependency/air_channel.mlir

PR #425, opened 2024-02-14
Support for <32bit data types in airrt-to-ipu

fifield:
The `aiex.ipu.dma_memcpy_nd` ops require memrefs with 32-bit element types. This patch inserts `unrealized_conversion_cast` for any dma copies with memrefs with 8 or 16-bit element types during lowering in airrt-to-ipu pass. The types of the control function arguments are also updated which results in the casts getting optimized away.


mlir/lib/Conversion/AIRRtToIpuPass.cpp
```diff
@@ -957,16 +1031,21 @@ struct AIRRtToIpuPass : public impl::AIRRtToIpuBase<AIRRtToIpuPass> {
...
+      // if the memref is the result of a cast of an arg, return
+      if (auto cast = dyn_cast_or_null<UnrealizedConversionCastOp>(memref.getDefiningOp()))
+        if (std::find(args.begin(), args.end(), cast.getOperand(0)) != args.end())
+          return;
+        else
+          memref = cast.getOperand(0);
+      // push back if unique
+      if (std::find(memrefs.begin(), memrefs.end(), dma.getMemref()) == memrefs.end()) {
+        memrefs.push_back(dma.getMemref());
+        memrefTypes.push_back(dma.getMemref().getType());
```
fifield, line 1048: This fixes a utility function that doesn't need to exist. Since we rely on the `func.func` to be in a certain form (e.g. with output parameters), the `func.func` to be lowered really should be in that form coming into the pass, instead of trying to patch it on the fly. We have at least [one test with broken IR](https://github.com/Xilinx/mlir-air/blob/main/mlir/test/Conversion/AIRRtToIpu/airrt_to_ipu.mlir#L387) depending on this fixup.

fifield:
my understanding of the offset, sizes, and strides of the `aiex.ipu.dma_memcpy_nd`[1] is that everything except the highest 3 strides have units "number of 32-bit words".

fifield:
> > my understanding of the offset, sizes, and strides of the `aiex.ipu.dma_memcpy_nd`[1] is that everything except the highest 3 strides have units "number of 32-bit words".
> > [1] I don't think this is documented
> 
> Correct. Though IIRC the strides are all in 32-bit words. The lengths/offsets are multiples of the lower dimensions.

Right, I wrote strides but meant sizes. I think I've gotten offsets wrong now that I look at the dma-to-ipu lowering in mlir-aie. We interpret them as multiples of element size to make the base offset.

fifield:
> I think I've gotten offsets wrong now that I look at the dma-to-ipu lowering in mlir-aie. We interpret them as multiples of element size to make the base offset.

Except that bitwidth is always 32, so we _do_ need to translate to multiples of 32-bit words in airrt-to-ipu

PR #427, opened 2024-02-16
A tsan warning fix

newling:
band[0] was modified or removed before used maybe? Not 100% sure but this fixes the warning.


erwei-xilinx (approved):
Thanks for working on this! LGTM.

PR #437, opened 2024-02-20
Add e2e example with pack-based (2k,2k,2k) GEMM

erwei-xilinx:


fifield:
Can you add e2e tests in addition to examples since we don't regress examples?  Tests can be tested, examples rot.

erwei-xilinx:
> Can you add e2e tests in addition to examples since we don't regress examples? Tests can be tested, examples rot.

Great point - PR created. The test may be a bit slow.

PR #438, opened 2024-02-20
Add a test corresponding to the air-to-ipu pack-based gemm example

erwei-xilinx:


fifield (approved):
I ran it and it printed `PASS!`

erwei-xilinx:
> I ran it and it printed `PASS!`

Awesome, thanks for checking, @fifield !

PR #440, opened 2024-02-20
Add e2e test of external function linking for ipu target

fifield:
Add a test for calling external function from `air.herd` for the ipu xrt target.

issues / work in progress:
* It seems to require `-air-dma-to-channel -air-dependency`, in that order, which is not the usual order.
* ~~The external function declaration gets duplicated during lowering, leading to error. Fixing this by hand gives `PASS!`~~

### DIFF
--- a/test/xrt/05_extern_func/run.lit
+++ b/test/xrt/05_extern_func/run.lit
@@ -8,6 +8,6 @@
 // REQUIRES: ryzen_ai, valid_xchess_license
 
 // RUN: xchesscc_wrapper aie2 -I %aietools/include -c %S/chess/beefmaker_kernel.cc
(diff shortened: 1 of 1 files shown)


fifield:
* ~~The external function declaration gets duplicated during lowering, leading to error. Fixing this by hand gives `PASS!`~~ 

Workaround by adding `-symbol-dce`

fifield:
> * It seems to require `-air-dma-to-channel -air-dependency`, in that order, which is not the usual order.

@erwei-xilinx fyi, It doesn't seem to generate the shim allocation metadata correctly without this ordering or without the air-dependency. It also fails without a canonicalization between the dma-to-channel and air-dependency passes:
```
error: 'scf.parallel' op currently only supporting input scf::ParallelOp with no reductions
```

erwei-xilinx:
> @erwei-xilinx fyi, It doesn't seem to generate the shim allocation metadata correctly without this ordering or without the air-dependency. It also fails without a canonicalization between the dma-to-channel and air-dependency passes:

Thanks for identifying the issues. Let me take a look.

PR #446, opened 2024-02-21
Add e2e test with bfloat16

fifield:
Add an end to end test of eltwise add of two 128x128xbf16 tensors. The output type is f32.

### DIFF
--- a/test/xrt/06_add_shim_bf16/gen.py
+++ b/test/xrt/06_add_shim_bf16/gen.py
@@ -1,0 +1,118 @@
+# gen.py -*- Python -*-
+#
+# Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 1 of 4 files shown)


test/xrt/06_add_shim_bf16/gen.py
```diff
@@ -0,0 +1,118 @@
...
+    mlir_module = generate_add_module([128,128], BF16Type.get())
+
+    print("\nTiled AIR Module:\n\n", mlir_module)
+    # with open("add.air.mlir", "w") as f:
+    #     f.write(str(mlir_module))
+
+    pipeline = "builtin.module(" + ",".join([
+        "func.func(air-lower-herd-parallel)",
+        #"air-dependency",
+        #"air-dependency-schedule-opt",
```
fifield, line 72: note: air-dependency doesn't seem to work here.

PR #447, opened 2024-02-22
Fix to enable lowering of 16-bit types 

newling:
Before this PR, the IR created in airrt-to-ipu was like

```
           func.func @matmul_8x32_16xi32__dispatch_0_matmul_8x32x16_i16xi16xi32(%arg0: memref<64xi32>, %arg1: memref<256xi32>, %arg2: memref<8x32xi32>) {
             %0 = hal.interface.binding.subspan set(0) binding(0) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : memref<8x16xi16>
             memref.assume_alignment %0, 64 : memref<8x16xi16>
             %1 = hal.interface.binding.subspan set(0) binding(1) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : memref<16x32xi16>
             memref.assume_alignment %1, 64 : memref<16x32xi16>
             %2 = hal.interface.binding.subspan set(0) binding(2) type(storage_buffer) alignment(64) offset(%c0) : memref<8x32xi32>
             memref.assume_alignment %arg2, 64 : memref<8x32xi32>
             %3 = builtin.unrealized_conversion_cast %0 : memref<8x16xi16> to memref<64xi32>
             aiex.ipu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 0][4, 1, 8, 8][0, 0, 8]) {id = 0 : i64, metadata = @airMemcpyId4} : memref<64xi32>
```

The `hal.interface.binding.subspan` ops are then never dead-code-eliminated, I assume because the `memref.assume_alignment` cannot be removed (it might throw?)

The IR we want the pass to generate should have %arg1 and %arg2 being the operands to `hal.interface.binding.subspan` ops. i.e. we want: 

```
           func.func @matmul_8x32_16xi32__dispatch_0_matmul_8x32x16_i16xi16xi32(%arg0: memref<64xi32>, %arg1: memref<256xi32>, %arg2: memref<8x32xi32>) {
             %0 = hal.interface.binding.subspan set(0) binding(0) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : memref<8x16xi16>
             memref.assume_alignment %arg0, 64 : memref<64xi32>
             %1 = hal.interface.binding.subspan set(0) binding(1) type(storage_buffer) alignment(64) offset(%c0) flags(ReadOnly) : memref<16x32xi16>
             memref.assume_alignment %arg1, 64 : memref<256xi32>
             %2 = hal.interface.binding.subspan set(0) binding(2) type(storage_buffer) alignment(64) offset(%c0) : memref<8x32xi32>
             memref.assume_alignment %arg2, 64 : memref<8x32xi32>
             %3 = builtin.unrealized_conversion_cast %0 : memref<8x16xi16> to memref<64xi32>
             aiex.ipu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 0][4, 1, 8, 8][0, 0, 8]) {id = 0 : i64, metadata = @airMemcpyId4} : memref<64xi32>
```

In this case the `hal.interface.binding.subspan` ops can be removed in a subsequent canonicalization pass. 

@fifield this is my goal, and the "peep-hole" pattern I've added achieves this, although you might have a neater way of doing this.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToIpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToIpuPass.cpp
@@ -61,6 +61,7 @@ struct RelocateAssumeAlignmentOp
   matchAndRewrite(memref::AssumeAlignmentOp assumeOp,
                   mlir::PatternRewriter &rewriter) const override {
 
+
     auto producerOp = assumeOp.getOperand().getDefiningOp();
     if (!producerOp)
       return rewriter.notifyMatchFailure(assumeOp,
--- a/mlir/lib/Conversion/AIRRtToIpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToIpuPass.cpp
@@ -17,6 +17,7 @@
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/LoopUtils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/Utils/Utils.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/DialectConversion.h"
@@ -26,6 +27,7 @@
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
+#include <cstddef>
 
 #define DEBUG_TYPE "airrt-to-ipu-pass"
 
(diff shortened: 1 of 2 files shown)


fifield:
Is this be a limitation of the lowering I added in #425? What's supposed to happen is that both the function args and the air dma operands are converted to i32 types and an unrealized conversion cast is inserted in both places, then the casts get canonicalized away. I'm not sure what your input IR looks like, maybe I made some wrong assumptions.

newling:
> Is this a limitation of the lowering I added in #425? What's supposed to happen is that both the function args and the air dma operands are converted to i32 types and an unrealized conversion cast is inserted in both places, then the casts get canonicalized away. I'm not sure what your input IR looks like, maybe I made some wrong assumptions.

I think so. 
My understanding of what is happening (before this PR) is that you go from 


%0 = hal.load
memref.assume %0
air.dma %0 

to 

%0 = hal.load
memref.assume %0
%1 = conversion_cast %0
air.dma %1

to 

%0 = hal.load
memref.assume %0 
%1 = conversion_cast %arg0
air.dma %1


The hal.load is not DCE'd because it's result is consumed by the memref.assume op. 

The pattern added here transfers the memref.assume op consume the conversion_cast's output rather. This means that the hal.load can be DCE'd. 

It definitely helps lowering my bfloat16 IREE matmul!

fifield:
If #449 doesn't help then I'd like to at least capture the problem in a test that doesn't depend on hal dialect

fifield (approved):

PR #449, opened 2024-02-22
Run CastFunctionArgs after BufferMemrefToFuncArgs

fifield:
Run `CastFunctionArgs` after `BufferMemrefToFuncArgs` because `CastFunctionArgs` assumes the IR is in the form produced by the legalization in `BufferMemrefToFuncArgs`.


fifield:
@newling I tried to capture the issue from #447 in the unit test added here

newling:
@fifield  This unfortunately doesn't fix the issue fixed in https://github.com/Xilinx/mlir-air/pull/447

I can create a lit test for #447

fifield:
> @fifield This unfortunately doesn't fix the issue fixed in #447
> 
> I will create a lit test for #447

Thanks for trying it. If this PR didn't make your problem worse I'd like merge it anyway. The pass was already trying to support the test added in this PR, it just didn't work for 16bit because of the ordering of things inside the pass.

newling (approved):
Looks good, I think the test I will add for my edge case will look very similar

PR #460, opened 2024-02-27
Bump the mlir-aie version that is used 

newling:
aie's buffer op now has an optional initial value.


newling:
#461 bumps llvm too, so closing #460 (will let @fifield land #461)

PR #461, opened 2024-02-27
update mlir-aie

fifield:
update llvm to 641d160ad
update mlir-aie to 2903588


fifield:
ha! pushed 1 min behind https://github.com/Xilinx/mlir-air/pull/460. @newling I think I updated one more file here.

newling:
I thought someone else might doing this! I see you've bumped llvm too, so I'll close #460

PR #464, opened 2024-02-28
Tweak airrt segment lowering

fifield:
This is part of a rabbit hole trying to make `air-dma-to-channel` safe to run on a broader set of code, in particular the existing airhost tests, so that it can be included in standard pipelines like aircc.py

This updates the air.segment to airrt lowering by,

* Don't delete segments if they don't contain dma ops. The idea is to preserve control flow around herd load operations. e.g. a herd may occur in a loop inside a segment.
* Don't "Move herd load to before the first ctrl loop"  This was (I think) a workaround of the fact that channel get/put ops, after being hoisted out of herd, will happen before herd or segment load, which is an error.
* Instead of above, make sure segment load is emitted at the start of its region. This implies that air.segment is required.
* replace some `replaceAllUsesWith`/`erase` with rewriter `clone`/`eraseOp`

### DIFF
--- a/mlir/lib/Conversion/AIRLoweringPass.cpp
+++ b/mlir/lib/Conversion/AIRLoweringPass.cpp
@@ -16,6 +16,7 @@
 
 #include "aie/Dialect/AIE/IR/AIEDialect.h"
 
+#include "mlir/Conversion/ReconcileUnrealizedCasts/ReconcileUnrealizedCasts.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/IR/AffineValueMap.h"
 #include "mlir/Dialect/Affine/LoopUtils.h"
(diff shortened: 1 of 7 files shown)


mlir/lib/Conversion/AIRLoweringPass.cpp
```diff
@@ -583,28 +611,32 @@ class AIRChannelPutToAIRRtConversion
...
+        if (o.getType().isa<xilinx::airrt::EventType>())
+          deps.push_back(o);
+      if (deps.size())
+        rewriter.replaceOpWithNewOp<xilinx::airrt::WaitAllOp>(
+            op, xilinx::airrt::EventType::get(ctx), deps);
+      return success();
+    }
+
+    rewriter.eraseOp(op);
+    return success();
```
erwei-xilinx, line 639: Duplicate `return success();`

erwei-xilinx (approved):
LGTM.

PR #466, opened 2024-02-29
Add openssl-dev to buildAndTestAieTools workflow

fifield:
Fix workflow failure after mlir-aie update


fifield:
https://github.com/Xilinx/mlir-air/actions/runs/8100115494/job/22137374575

PR #468, opened 2024-03-01
Collapse multi-dimensional offsets to 1d for bf16 func arguments

erwei-xilinx:
... when func argument memref type gets collapsed.

This PR inteprets offset(i), 0 <= i < 3, as "offset after this many dim i-1 strides". Offset(3), i.e. lowest-dimension offset, is still counted as "number of 32-bit offsets".

Without this PR, IPU instructions on multi-dimensional, bf16 memrefs are failing at MLIR-AIE due to higher-dimension offsets (higher than memref rank, i.e. 1) being ignored.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToIpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToIpuPass.cpp
@@ -138,17 +138,17 @@ struct DmaToIpuPattern : public OpConversionPattern<DmaMemcpyNdOp> {
     SmallVector<Value> offsets;
     SmallVector<int64_t> staticOffsets;
     if (auto const_int = getConstantIntValue(adaptor.getOffset3()))
-      staticOffsets.push_back(*const_int / div);
+      staticOffsets.push_back(*const_int);
     else
-      offsets.push_back(divOp(adaptor.getOffset3()));
+      offsets.push_back(adaptor.getOffset3());
     if (auto const_int = getConstantIntValue(adaptor.getOffset2()))
-      staticOffsets.push_back(*const_int / div);
+      staticOffsets.push_back(*const_int);
     else
-      offsets.push_back(divOp(adaptor.getOffset2()));
+      offsets.push_back(adaptor.getOffset2());
     if (auto const_int = getConstantIntValue(adaptor.getOffset1()))
-      staticOffsets.push_back(*const_int / div);
+      staticOffsets.push_back(*const_int);
     else
-      offsets.push_back(divOp(adaptor.getOffset1()));
+      offsets.push_back(adaptor.getOffset1());
     if (auto const_int = getConstantIntValue(adaptor.getOffset0()))
       staticOffsets.push_back(*const_int / div);
     else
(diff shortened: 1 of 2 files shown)


mlir/lib/Conversion/AIRRtToIpuPass.cpp
```diff
@@ -344,6 +344,34 @@ static LogicalResult CastFunctionArgs(func::FuncOp funcOp,
     rewriter.setInsertionPointToStart(&entry);
     auto cast = rewriter.create<UnrealizedConversionCastOp>(
         rewriter.getUnknownLoc(), memrefTy, entry.getArgument(i));
+    // With memref shape collapsed to 1d, the multi-dimensional offset also
+    // needs to be collapsed.
+    SmallVector<Operation *> users;
+    for (auto user : entry.getArgument(i + 1).getUsers()) {
+      if (auto cast = dyn_cast<UnrealizedConversionCastOp>(user)) {
```
fifield, line 351: `cast` shadows the variable from line 345. best to pick another name.

mlir/lib/Conversion/AIRRtToIpuPass.cpp
```diff
@@ -344,6 +344,34 @@ static LogicalResult CastFunctionArgs(func::FuncOp funcOp,
...
+                        *getConstantIntValue(dmaUser.getMixedStrides()[i]);
+        rewriter.setInsertionPoint(dmaUser);
+        const std::vector<int64_t> newStaticOffsets = {0, 0, 0, oneDOffset};
+        rewriter.create<AIEX::IpuDmaMemcpyNdOp>(
+            rewriter.getUnknownLoc(), dmaUser.getX(), dmaUser.getY(),
+            dmaUser.getMemref(), SmallVector<Value>{}, dmaUser.getSizes(),
+            dmaUser.getStrides(), ArrayRef(newStaticOffsets),
+            dmaUser.getStaticSizes(), dmaUser.getStaticStrides(),
+            dmaUser.getMetadata(), dmaUser.getId());
+        dmaUser->erase();
```
fifield, line 372: This pattern already modifies the IR directly, but should use `rewriter.eraseOp` here instead of `erase`

fifield (changes requested):
Thanks for the tricky bugfix.

mlir/lib/Conversion/AIRRtToIpuPass.cpp
erwei-xilinx: Done.

mlir/lib/Conversion/AIRRtToIpuPass.cpp
erwei-xilinx: Done.

erwei-xilinx:
@fifield Thanks for the comments. Updated.

mlir/lib/Conversion/AIRRtToIpuPass.cpp
newling: Variable `i` is also shadowing another `i` too.

newling (commented):
LGTM

fifield (approved):

mlir/lib/Conversion/AIRRtToIpuPass.cpp
erwei-xilinx: Thanks for spotting this, @newling ! Done.

PR #473, opened 2024-03-06
[AIR][AMD-AIE] Fix `link_with` relaying for nested func.call ops

Abhishek-Varma:
-- This commit adds a fix to relay `link_with` information from
   nested func.call ops within scf.parallel op while creating
   air.herd op.

Signed-off-by: Abhishek Varma <abhvarma@amd.com>


Abhishek-Varma:
Hi @fifield @erwei-xilinx - could this be reviewed/merged?

The only major thing it adds as a fix is the "walk" of ALL ops within scf.parallel - the earlier version was very shortsighted and only was hunting for the immediate child `func.call` op.

PR #477, opened 2024-03-06
Remove unused lit config option vitis_root

fifield:
The check for `config.vitis_root` was preventing lit tests tagged with `REQUIRES: valid_xchess_license` from running in the "Build and Test with AIE tools" action. The check is not used anywhere else, and is pointless, so the fix it to remove it.

### DIFF
--- a/.github/workflows/buildAndTestAieTools.yml
+++ b/.github/workflows/buildAndTestAieTools.yml
@@ -155,6 +155,8 @@ jobs:
               export LIT_FILTER="${{ inputs.LIT_FILTER }}"
             fi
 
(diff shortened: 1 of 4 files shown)


.github/workflows/buildAndTestAieTools.yml
```diff
@@ -155,6 +155,8 @@ jobs:
               export LIT_FILTER="${{ inputs.LIT_FILTER }}"
             fi
 
+            export LIT_OPTS="-sv --timeout 300 -j1 --time-tests --order=random"
```
fifield, line 158: Needed this because the runner ran out of memory without it. At least I think that's what happened...it kept getting killed partway through.

PR #479, opened 2024-03-07
Fixup a bug when removing redundant dimensions; enable clearing of wr…

erwei-xilinx:
…ap-and-stride lists if default access pattern

### DIFF
--- a/mlir/lib/Util/Util.cpp
+++ b/mlir/lib/Util/Util.cpp
@@ -809,6 +809,7 @@ LogicalResult air::canonicalizeWrapAndStrideList(OpBuilder builder,
                                                  SmallVector<Value> &strides,
                                                  int memref_volume) {
 
+  bool listsHaveChanged = false;
   // Match offsets size with sizes and strides
   int max_dim_size =
       std::max(std::max(offsets.size(), sizes.size()), strides.size());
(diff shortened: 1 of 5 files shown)


mlir/lib/Util/Util.cpp
```diff
@@ -833,34 +834,43 @@ LogicalResult air::canonicalizeWrapAndStrideList(OpBuilder builder,
...
+      if (*getConstantIntValue(strides[0]) == 1 && *getConstantIntValue(sizes[0]) == memref_volume){
+        offsets.erase(offsets.begin());
+        sizes.erase(sizes.begin());
+        strides.erase(strides.begin());
+      }
+    }
   }
 
-  if (unit_dims.empty() && redundant_dims.empty()) {
+  if (unit_dims.empty() && !hasRedudantDims) {
```
fifield, line 873: Is it possible to erase the vectors in the loop above but not trigger these conditions? i.e. a false `failure()`?

mlir/lib/Util/Util.cpp
```diff
@@ -833,34 +834,43 @@ LogicalResult air::canonicalizeWrapAndStrideList(OpBuilder builder,
...
   }
 
-  SmallVector<int> redundant_dims;
-  if (!sizes.empty())
-    for (int i = sizes.size() - 1; i >= 1; i--) {
-      if (getConstantIntValue(sizes[i]) && getConstantIntValue(sizes[i - 1]) &&
+  bool hasRedudantDims = false;
+  if (!sizes.empty()){
+    int i = sizes.size() - 1;
+    while (i >= 1) {
```
fifield, line 840: can this use logic similar to line 823?
```c++
// skip if size of 'sizes' is zero or one
for (int i = sizes.size() - 1; i >=1; i--)
```

erwei-xilinx: Done.

fifield (approved):

PR #481, opened 2024-03-08
GEMM (256x1024x256, bf16) E2E example, with external vectorized kernel, in CI

erwei-xilinx:


test/xrt/08_gemm_extern_vec/run.lit
```diff
@@ -0,0 +1,10 @@
+// (c) Copyright 2024 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+// REQUIRES: ryzen_ai, valid_xchess_license
+
+// RUN: xchesscc_wrapper aie2 -I %aietools/include -c %S/mm.cc -o mm.o
+// RUN: %python %S/aie.py
+// RUN: %python aiecc.py --no-aiesim --aie-generate-cdo --aie-generate-ipu --no-compile-host --xclbin-name=aie.xclbin --ipu-insts-name=insts.txt aie.mlir
+// RUN: g++-13 %S/test.cpp -o test.exe -std=c++23 -Wall %xrt_flags -lrt -lstdc++ -lboost_program_options -lboost_filesystem
```
jsetoain, line 9: Do we want to have the version of `g++` hardcoded there?

fifield: It's somewhat necessary at the moment to ensure c++23 for bf16

fifield (approved):
Thanks, this is hopefully a good check that we don't break code coming out of iree. We're close to having a version without the embedded IR (i.e. generated from upstream+air tiling), I will post a draft PR later.

mlir/include/air/Conversion/Passes.td
```diff
@@ -140,6 +140,11 @@ def AIRLinalgToFunc : Pass<"air-linalg-to-func", "ModuleOp"> {
                 "function calls";
   let constructor = "xilinx::air::createAIRLinalgToFuncPass()";
   let dependentDialects = ["func::FuncDialect", "memref::MemRefDialect"];
+  let options = [
+    Option<"clLinkWith", "link-with", "std::string",
+          /*default=*/"\"\"",
+           "Compiled kernel object file to link with.">
```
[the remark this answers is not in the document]
erwei-xilinx: Thanks for the comment. Done.

mlir/lib/Conversion/ConvertToAIRPass.cpp
```diff
@@ -470,12 +470,16 @@ Value lookupOrDefaultRange(Value v, IRMapping &remap) {
...
-  Operation *output = nullptr;
+Operation *getCoreComputeOpFromExecuteOp(Operation *op) {
+  // We assume all linalg ops (except for linalg.copy) and func.call ops do
+  // computations only and do not participate in data movement.
   if (auto exec = dyn_cast<air::ExecuteOp>(op)) {
-    exec.walk([&](linalg::LinalgOp linalg_op) { output = linalg_op; });
+    if (isa<linalg::LinalgOp>(exec.getChildOp()))
+      return exec.getChildOp();
+    else if (isa<func::CallOp>(exec.getChildOp()))
+      return exec.getChildOp();
```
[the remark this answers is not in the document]
erwei-xilinx: I think this is a great point. For now we do asynchronous analysis on every single operation, so that the `air::ExecuteOp` currently never contains more than one operation in its body. But I think the definition for execute op is not restricted to single operation. So perhaps we should revisit the `getChildOp` method to either enforce to single child only, or do a `getChildOps` instead.

test/xrt/08_gemm_extern_vec/run.lit
jsetoain: Oh, I see. I'm so used to clang nowadays that I thought g++ 13 was an old version. Still, I don't think we're too far from g++ 14 release, and this test should pass with both. Ideally, support for c++23 should be checked, and the test enabled or disable accordingly, for good measure.

You can get a list of supported C++ standards with something like:

```sh
gcc -v --help 2> /dev/null | grep "std=" | grep Conform | cut -d"=" -f 2 | cut -d" " -f 1 | grep c++
```

Or:
```python
def list_gcc_supported_standards():
    supported = set()
    p = subprocess.run(['g++', '-v', '--help'], capture_output=True, encoding='utf8')
    for line in p.stdout.split('\n'):
        m = re.match(r'  -std=(c\+\+[\w]+)', line)
        if m:
            supported.add(m.group(1))
    return supported
```
In python.

PR #483, opened 2024-03-08
Update run.lit

fifield:
Add  `--xchesscc --xbridge` to aiecc command line. Otherwise test will fail to build if chess is not default compiler


erwei-xilinx (approved):
LGTM!

PR #485, opened 2024-03-09
Bugfix on affine.map propagation

erwei-xilinx:
Ensure that broadcast specialization only propagates through single-operand affine.apply ops.


nirvedhmeshram:
@erwei-xilinx I think we should land this, since its a correct bug fix that helps the iree-amd-aie project.

erwei-xilinx:
> @erwei-xilinx I think we should land this, since its a correct bug fix that helps the iree-amd-aie project.

Agreed. PR landed.

PR #491, opened 2024-03-12
Update testing for single python file e2e

fifield:
* Update lit.cfg.py with XRT testing environment variable
* Update test 02 to be a single file, testing vector mul of multiple data types. Remove newly redundant test 03.


fifield:
This crashes chess with any bfloat16 data types.

PR #497, opened 2024-03-14
Fixes for Windows build

spaceotter:
MSVC makes long 32 bits, which creates some confusion in these calls to std::max.


erwei-xilinx (approved):
LGTM!

PR #500, opened 2024-03-14
support 2d specialization

fifield:
The code patched here was crashing on the affine map in the modified test for two reasons: It's 2d, and it was using dims not symbols. The PR updates the pass to support the 2d affine map, but changes the test to use symbols instead of fixing the pass to support dims.


erwei-xilinx (approved):
Thanks for working on this.

PR #508, opened 2024-03-19
update mlir-aie

fifield:
update mlir-aie to e0514b13
update llvm to 60dda1f


fifield:
This seems to require the rewriter fixes in #464, where splice operations are replaced with clone.

PR #512, opened 2024-03-22
Update mlir aie

fifield:


fifield:
@erwei-xilinx please review the air-to-aie changes related to https://github.com/Xilinx/mlir-aie/pull/1111, in case there are any changes I missed. The mlir and xrt tests are passing for me.

mlir/test/Conversion/AIRToAIE/air_shimcpy_to_aie2.mlir
```diff
@@ -735,7 +735,7 @@ func.func @func11(%arg0: memref<128xbf16>, %arg1: memref<128xbf16>) {
         air.execute_terminator %alloc : memref<32x256xbf16, 1>
       }
       %2 = scf.parallel (%arg4) = (%c0) to (%c2_0) step (%c1) init (%async_token) -> !air.async.token {
-        %4 = air.channel.put async [%async_token]  @channel_1[%arg4, %c0] (%results[%c0, %c0, %c0] [%c8, %c32, %c256] [%c32, %c256, %c1]) {id = 4 : i32} : (memref<32x256xbf16, 1>)
```
fifield, line 738: This code was generating out of bounds bd access, which is now checked by mlir-aie verifier, because of the extra 8 in the size `[%c8, %c32, %c256]`. The channel get later on in the herd also generates an out of bounds offset, but it is not caught by the verifier.

erwei-xilinx: Make sense! Thanks for identifying that error in the test. Perhaps if we change the channel put op to be `%results[%c0, %c0, %c0] [%c8, %c32, %c32] [%c32, %c256, %c1]` then it should avoid the issue.

erwei-xilinx (approved):
The changes in `-air-to-aie` look good to me.

mlir/test/Conversion/AIRToAIE/air_shimcpy_to_aie2.mlir
fifield: You're right, I fixed it in https://github.com/Xilinx/mlir-air/pull/512/commits/6f3e7c8c2465ce0a48065f5e242d94a616bdd1a8

PR #528, opened 2024-04-05
Initial commit of MLIR-AIR documentation to CI test 09

erwei-xilinx:


fifield:
This is good stuff, can some of it be used to update the [pass](https://github.com/Xilinx/mlir-air/blob/main/mlir/include/air/Conversion/Passes.td) [documentation ](https://github.com/Xilinx/mlir-air/blob/main/mlir/include/air/Transform/Passes.td) that gets [auto](https://xilinx.github.io/mlir-air/AIRTransformPasses.html)-[generated](https://xilinx.github.io/mlir-air/AIRConversionPasses.html)?

erwei-xilinx:
> This is good stuff, can some of it be used to update the [pass](https://github.com/Xilinx/mlir-air/blob/main/mlir/include/air/Conversion/Passes.td) [documentation ](https://github.com/Xilinx/mlir-air/blob/main/mlir/include/air/Transform/Passes.td) that gets [auto](https://xilinx.github.io/mlir-air/AIRTransformPasses.html)-[generated](https://xilinx.github.io/mlir-air/AIRConversionPasses.html)?

Yeah definitely. This doc is more specific to the E2E example, where the explanation there is more tailored to this specific GEMM case. I'm thinking of putting together a separate set of doc to be the pass-specific auto-generated doc.

fifield (approved):

PR #535, opened 2024-04-12
Placing herds with shared herd name to the same location in segment

erwei-xilinx:


fifield:
This seems to exploit a hole in the dialect where we allow multiple herds to be declared with the same symbol name. And it assumes the bodies are identical? It might be worth examining how to outline the common herd body somewhere else, then refer to it by name. Maybe with something similar to `gpu.launch_func` to launch a `func` as an `air.herd`?

erwei-xilinx:
> This seems to exploit a hole in the dialect where we allow multiple herds to be declared with the same symbol name. And it assumes the bodies are identical? It might be worth examining how to outline the common herd body somewhere else, then refer to it by name. Maybe with something similar to `gpu.launch_func` to launch a `func` as an `air.herd`?

Agreed. We should definitely revisit this, because this is currently misusing the sym_names. The intent here is to label multiple air.herds which will then be outlined to the same set of AIE cores, where each core runs the herd bodies in sequence.

Would the launch_func have to launch air.herd with identical bodies? Because the air.herds to be outlined to shared group of AIE cores may not have identical bodies.

PR #537, opened 2024-04-12
Outlining multiple herds to the same aie.cores

erwei-xilinx:


fifield:
Sounds like a useful feature, what will the user interface look like?

erwei-xilinx:
> Sounds like a useful feature, what will the user interface look like?

The interface shall get generated by the -air-place-herd-on-segment pass, after this PR lands https://github.com/Xilinx/mlir-air/pull/535. I will create more unit tests for these PRs to demo the user interface, but basically it is from a multiple of air herds that get placed onto the same anchor coordinates + having the same shapes

PR #544, opened 2024-04-23
Fixup AIRRtToIpuPass wrap tiling, if wrap is larger than 1023 but not divisible by 512

erwei-xilinx:


mlir/lib/Conversion/AIRRtToIpuPass.cpp
```diff
@@ -531,6 +531,19 @@ bool violatesAIE2WrapLimit(airrt::DmaMemcpyNdOp dma) {
   return false;
 }
 
+// A naive implementation to find largest factor, smaller than a given int, for
+// a given integer.
+int getLargestFactorSmallerThan(int inputInt, int smallerThanInt = 0) {
```
newling, line 536: @erwei-xilinx here is iree-amd-aie's efficient implementation of this: https://github.com/nod-ai/iree-amd-aie/blob/da8025312112c10201005212e4d2718798f47dda/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEUtils.cpp#L334

PR #546, opened 2024-04-23
Fixup a bug at AIRToAIEPass AIE.Core lock allocation

erwei-xilinx:
Enforce core dma ping-pong to only happen where different L1 buffers were accessed by the same channel.

### DIFF
--- a/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
+++ b/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
@@ -269,32 +269,38 @@ std::pair<int64_t, int64_t> air::getLockValuePair(AIE::AIEArch arch,
   if (!air_chan)
     return getLockValuePair(arch, buffer_memref);
 
-  // Infer semaphore lock values using air.channel
-  int read_counter = 0;
-  int write_counter = 0;
+  // Infer semaphore lock values using air.channel. This method enables
+  // ping-pong compute-communication overlap.
+  SmallVector<Value> unique_write_buffers;
+  SmallVector<Value> unique_read_buffers;
   for (auto get : getChannelGetOpThroughSymbol(air_chan)) {
     if (isa<AIE::ExternalBufferOp>(buffer_memref.getDefiningOp())) {
       // Shim DMA locks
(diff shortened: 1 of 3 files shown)


mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
```diff
@@ -1009,6 +1015,12 @@ template <typename T> int air::foundInVector(T item, std::vector<T> vec) {
   return index;
 }
 
+template <typename T>
+void air::push_back_if_unique(SmallVector<T> &vec, T entry) {
+  if (std::find(vec.begin(), vec.end(), entry) == vec.end())
+    vec.push_back(entry);
+}
+
```
fifield, line 1023: aka `llvm::SmallSet` ?

erwei-xilinx: Make sense. Will replace them all. Thanks.

erwei-xilinx: Hmm looks like Value cannot be used as key to SmallSet, because `mlir::Value` isn't invocable by `std::less`:
```
error: static assertion failed due to requirement 'std::__is_invocable<std::less<mlir::Value> &, const mlir::Value &, const mlir::Value &>{}': comparison object must be invocable with two arguments of key type
```

fifield: > Hmm looks like Value cannot be used as key to SmallSet, because `mlir::Value` isn't invocable by `std::less`:
> 
> ```
> error: static assertion failed due to requirement 'std::__is_invocable<std::less<mlir::Value> &, const mlir::Value &, const mlir::Value &>{}': comparison object must be invocable with two arguments of key type
> ```

doh. Sorry for the distraction.

erwei-xilinx: Not at all, as the `push_back_if_unique` has also been used elsewhere on `Operation *`, which works with `SmallSet`. I have a PR to come very soon.

PR #550, opened 2024-04-24
AIRRtToNpuPass SHIM DMA BD optimization

erwei-xilinx:
- When tiling any wrap>1023 into two wraps, make the largest int factor the outer wrap, so that the new stride is smaller and less likely to go beyond 1M.
- Code quality.
- Avoid using `SymbolTable::getSymbolUses` method, which turns out to be quite slow with big IRs.


newling (approved):
Thanks!

PR #551, opened 2024-04-24
Improve compile time (AIRRtToNpuPass)

newling:
This reduces total end-to-end compile time from 7 seconds to 0.5 seconds on one model. At some point I had a model which was spending 2+ minutes in this air pass, I am optimistic that this optimization reduces that too (I can't recall the matmul dimensions). 

I am currently trying to build mlir-air outside of IREE to check locally that all the tests pass, but I'm also just throwing this to CI now to check.


erwei-xilinx:
Thanks for the cache implementation, which made the compilation go faster! Changes have been merged into https://github.com/Xilinx/mlir-air/pull/550 Maybe we can close this one now.

PR #552, opened 2024-04-25
try upgrading ubuntu

fifield:
replace ubuntu 20 with ubuntu 24


fifield:
nope, there isn't an ubuntu 24 runner type provided by github.

fifield:
> nope, there isn't an ubuntu 24 runner type provided by github.

https://github.com/actions/runner-images

keryell:
Yes, a beta only mid-May https://github.com/actions/runner-images/issues/9691 :-(
You could use a docker image, but I am unsure it is worth the effort for 2 weeks.

PR #561, opened 2024-05-01
Add NPU board test showing GEMM + BIAS ADD fusion

erwei-xilinx:
Requires mlir-aie bump https://github.com/Xilinx/mlir-air/pull/560


test/xrt/11_gemm_bias_fusion/test.cpp
fifield, line 1: needs copyright

erwei-xilinx: Thanks for spotting that! Fixed

PR #562, opened 2024-05-01
Fixup bug caused by tiling factor calculation

erwei-xilinx:
- The previous logic mistakenly took every dimension of an `scf.parallel` as a new access pattern; the entire `scf.parallel` is one access. This bug causes any 4x2-shaped `air.herds` to hang at runtime.
- Removed an obsolete assertion check.


newling (approved):
Thanks!

PR #564, opened 2024-05-06
bump mlir-aie

fifield:
update aie to b8d512fa
update llvm to 1cde1240

mlir snippets using `memref.expand_shape` updated to add `output_shape`


erwei-xilinx (approved):
Thanks! LGTM, as long as the xrt board tests are still passing.

PR #573, opened 2024-05-10
New lock allocator for bf16 element wise fusion

erwei-xilinx:


mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -2588,11 +2606,38 @@ class AIRToAIEPass : public air::impl::AIRToAIEBase<AIRToAIEPass> {
...
+        auto prevLockUse = dyn_cast<AIE::UseLockOp>(t->getPrevNode());
+        mergePrevLockUse(prevLockUse, lockRelValue);
+      }
+      auto br = dyn_cast<cf::BranchOp>(readInRAW->getBlock()->getTerminator());
+      if (br)
+        builder.setInsertionPointToStart(br.getDest());
+      else
+        builder.setInsertionPointToStart(readInRAW->getBlock());
+      builder.create<AIE::UseLockOp>(builder.getUnknownLoc(), relLockOp,
+                                     AIE::LockAction::Release, lockRelValue);
```
fifield, line 2640: This is a bit convoluted with all the conditionals and the calls to getReadInRAW and mergePrefLockUse. Do you have an example of the IR you're trying to match and replace here? Does the merge work for AIE1 locks? Or if the get and put are in the same block?

erwei-xilinx: This is a prototype of a new lock allocator. The goal here is to have the elementwise fusion example (the new board test 12) work on npu, where the challenge is that the same accumulator buffer has (1) dada copied in,(2) compute on and (3) data copied out.

Although the new board test seems to pass with the pr, it seems to be too complicated and not compatible with a number of existing tests, or AIE1. So I think I'll close the PR and find another solution.

PR #580, opened 2024-05-15
Don't build llvm when building docs

fifield:


fifield:
https://github.com/Xilinx/mlir-air/actions/runs/9104105570

PR #588, opened 2024-05-31
Docs and small updates for xrt build infrastructure

hunhoffe:
This PR:
- Adds python packages that are needed to run the lit tests in ```test/xrt``` to the requirements file
- Updates the ```.gitignore``` file to ignore xrt and python virtual environment ```sandbox``` artifacts
- Update build documentation to include some information on the building for the Ryzen device


utils/requirements.txt
```diff
@@ -1,3 +1,7 @@
+PyYAML>=5.3.1, <=6.0.1
+aiofiles
+filelock==3.13.1
+lit
 joblib
 ipykernel
 psutil
```
fifield, line 7: I've wondered, should we just say to use `mlir-aie/python/requirements.txt` ?

hunhoffe: I think this makes sense, although I see a con in that it would be an additional undocumented dependency between the two repos. It's possible it could cause some annoyance with maintenance if changes in dependencies/versions in mlir-aie aren't tested for their effects in mlir-air.

If you want me to change it, let me know!

fifield (approved):

PR #589, opened 2024-06-03
Propagate function symbol name as base for air.segment symbol name

nirvedhmeshram:
The current air.segment symbol naming logic makes sure that we don't end up with the same symbol name for air.segments in a given module. But the symbols can still be same across different modules. This can be a problem if we want to merge functions from multiple modules into one. In this PR we use the symbol name of the function enclosing the air.segment op as the base name. This way such conflicts wont happen assuming the function names are unique.


erwei-xilinx (approved):
Looks good to me. Thanks.

PR #592, opened 2024-06-05
trace generation

Yu-Zhewen:
Generate trace for all cores and memtiles using the new `trace-offset` and `trace-size` options, synchronized through event broadcast.


Yu-Zhewen:
@erwei-xilinx Hi, I think this PR is ready for review and please let me know if the example is also working on your side. Thanks!

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -3200,6 +3253,8 @@ class AIRToAIEPass : public air::impl::AIRToAIEBase<AIRToAIEPass> {
       lowerAIRMemcpyOp<air::DmaMemcpyNdOp>(device, shimDmaAlloc, options);
 
       // lowerPipelineGetPut(device, tileToHerdMap);
+      if (options.trace_size > 0)
+        createTracePacketFlow(device);
```
fifield, line 3257: Can you also add this to [runTestPatterns](https://github.com/Xilinx/mlir-air/blob/main/mlir/lib/Conversion/AIRToAIEPass.cpp#L3016), along with a unit test in `mlir/test/Conversion/AIRToAIE`? The run line in the unit test might look something like this:
```mlir
// RUN: air-opt %s -air-to-aie='test-patterns=insert-trace-packet-flow trace_offset=65536 trace_size=32768'
```

mlir/lib/Conversion/AIRRtToNpuPass.cpp
```diff
@@ -1251,6 +1258,145 @@ struct AIRRtToNpuPass : public impl::AIRRtToNpuBase<AIRRtToNpuPass> {
     }
   }
 
+  void insertNpuWrite32ForTrace(ModuleOp module, int64_t trace_size,
```
fifield, line 1261: This needs a unit test in `mlir/test/Conversion/AIRRtToNpu`

fifield (commented):
It might be handy to add a flag to aircc too.

erwei-xilinx:
Thanks for working on this, @Yu-Zhewen! I can indeed successfully generate trace file with the updated `xrt/01_air_to_npu` test. I wonder if this test can act as a working-out-of-the-box reference design for air-generated tracing, by setting the default trace size in aie.py and run.lit to be `32768` instead of `0`? Otherwise, if the user follow the doc and change aie.py but not the runtime command, they would get segfault.

Also, could you maybe add some instructions on which parse.py to use to generate trace file for visualization?

mlir/lib/Conversion/AIRToAIEPass.cpp
erwei-xilinx: Agreed. Adding more CI tests would be very helpful, so that for example future code development wouldn't silently break this feature.

mlir/lib/Conversion/AIRRtToNpuPass.cpp
Yu-Zhewen: added [mlir/test/Conversion/AIRRtToNpu/generate_trace_write32.mlir](https://github.com/Xilinx/mlir-air/pull/592/files#diff-bc62766234896565f7f875c3129091accef48afa0440e3893454e610d4b6639c)

fifield (approved):

erwei-xilinx (approved):
Looks good to me, thanks.

PR #594, opened 2024-06-05
[WIP] Add Matrix Scalar Add Programming Example

hunhoffe:
This PR adds a programming example of matrix scalar addition using channels and tiling. It includes 5 different ways to do tiling in this context.

This PR also includes some changes to the python API with the goal of providing a clean interface:
* Wrap `AllocOp` and `DeallocOp` to clean up `AllocOp` interface, and then wrap `DeallocOp` to maintain a symmetric interface
* Add default values for the `ChannelGet`/`ChannelPut` operations, as well as automatic conversion from python ints where it seems useful.


programming_examples/matrix_scalar_add/README.md
```diff
@@ -0,0 +1,78 @@
...
+This example ([matrix_scalar_add_channel.py](matrix_scalar_add_channel.py)) uses channel *sizes*, *offsets*, and *strides* to explicitly loop over the image worth of data to read, increment, and write to one tile at a time. The entirety of the work is done by one launch which manages one segment which manages one herd which manages one core. This is the only example that explicitly uses only one core to do all of the processing.
+```bash
+make clean && make channel
+```
+
+#### ```herd```: Tiling using a herd managing multiple cores
+
+This example ([matrix_scalar_add_herd.py](matrix_scalar_add_herd.py)) uses herds and the internal logic of channels to process all the tiles. Here, there is one launch and one segment and one herd, but a herd of a larger size: one core per tile to process in the image.
+
+TODO: is this example correct? Do we need to specify the offsets, etc. to make sure we aren't doing extra work?
```
hunhoffe, line 34: I'm not sure how much to rely on implicit behavior in this example

programming_examples/matrix_scalar_add/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```bash
+make clean && make herd
+```
+
+#### ```segment```: Tiling using multiple segments
+
+This example ([matrix_scalar_add_segment.py](matrix_scalar_add_segment.py))  uses segments and the internal logic of channels to process all the tiles. Here, there is one launch and several segments: one segment per tile. Each segment consists of a single herd and each of those herds manages a single core which processes one tile. The structure of this example is very similar to the herd example.
+
+TODO: is this example correct? Do we need to specify the offsets, etc. to make sure we aren't doing extra work?
```
hunhoffe, line 44: I'm not sure how much to rely on implicit behavior in this example

programming_examples/matrix_scalar_add/README.md
```diff
@@ -0,0 +1,78 @@
...
+This example ([matrix_scalar_add_launch.py](matrix_scalar_add_launch.py)) uses launches and channel sizes, strides, and offsets. Each launch only reads/writes a single tile from the input/output. Each launch then manages a single segment, herd, and core, and that core processes the single tile of data.
+
+```bash
+make clean && make launch
+```
+
+#### ```implicit```: Tiling using channels and data sizes
+
+This example ([matrix_scalar_add_implicit.py](matrix_scalar_add_implicit.py)) demonstrates the power of the channel abstraction. Here, we provide no explicit instructions on how to iterate through the tiles or index into the input/output data. We simply specify the input/output channels with input/output of type ```i32 x IMAGE_SIZE```. Then, we use the channels on the core to only read/write ```i32 x TILE_SIZE```. mlir-air is able to infer by the difference between data types that we want to read/write all the data, so logic is added without the writer explicitly signalling for it to happen.
+
```
hunhoffe, line 61: I'm still not sure if the fact that this example works seems like a feature or a bug

programming_examples/matrix_scalar_add/matrix_scalar_add_herd.py
```diff
@@ -0,0 +1,95 @@
...
+                        element_type=T.i32(),
+                        memory_space=mem_space,
+                    )
+
+                    # We must allocate a buffer of the tile size for the input/output
+                    tile_in = Alloc(tile_type)
+                    tile_out = Alloc(tile_type)
+
+                    # Input a tile
+                    ChannelGet("ChanIn", tile_in)
```
hunhoffe, line 68: @jgmelber I'm not sure how much to rely on implicit behavior in this example. What I have here seems to work, but should I instead explicitly specify how to walk through the data using the strides/sizes/offsets in the channel? What is (or should be) considered best practice?

programming_examples/matrix_scalar_add/matrix_scalar_add_implicit.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (C) 2024, Advanced Micro Devices, Inc.
```
hunhoffe, line 1: This example still feels a bit like magic to me. I can see that it somehow works - all the data in the image is updated, not just one tile - but it's opaque to me at what level and how some sort of looping behavior eventually is introduced.

I looked at the mlir generated by `python matrix_scalar_add_implicit.py`, and I get:
```mlir
module {
  air.channel @ChanIn []
  air.channel @ChanOut []
  func.func @addone(%arg0: memref<32x16xi32>, %arg1: memref<32x16xi32>) {
    air.launch () in () args(%arg2=%arg0, %arg3=%arg1) : memref<32x16xi32>, memref<32x16xi32> {
      air.channel.put  @ChanIn[] (%arg2[] [] []) : (memref<32x16xi32>)
      air.channel.get  @ChanOut[] (%arg3[] [] []) : (memref<32x16xi32>)
      air.segment @seg  {
        %c1 = arith.constant 1 : index
        %c1_0 = arith.constant 1 : index
        air.herd @addherd  tile (%arg4, %arg5) in (%arg6=%c1, %arg7=%c1_0) {
          %alloc = memref.alloc() : memref<16x8xi32, 2 : i32>
          %alloc_1 = memref.alloc() : memref<16x8xi32, 2 : i32>
          air.channel.get  @ChanIn[] (%alloc[] [] []) : (memref<16x8xi32, 2 : i32>)
          %c8 = arith.constant 8 : index
          %c1_2 = arith.constant 1 : index
          scf.for %arg8 = %c0 to %c8 step %c1_2 {
            %c0_3 = arith.constant 0 : index
            %c16 = arith.constant 16 : index
            %c1_4 = arith.constant 1 : index
            scf.for %arg9 = %c0_3 to %c16 step %c1_4 {
              %0 = memref.load %alloc[%arg9, %arg8] : memref<16x8xi32, 2 : i32>
              %c1_i32 = arith.constant 1 : i32
              %1 = arith.addi %0, %c1_i32 : i32
              memref.store %1, %alloc_1[%arg9, %arg8] : memref<16x8xi32, 2 : i32>
            }
          }
          air.channel.put  @ChanOut[] (%alloc_1[] [] []) : (memref<16x8xi32, 2 : i32>)
          memref.dealloc %alloc : memref<16x8xi32, 2 : i32>
          memref.dealloc %alloc_1 : memref<16x8xi32, 2 : i32>
          air.herd_terminator
        }
        air.segment_terminator
      }
      air.launch_terminator
    }
    return
  }
}
```
I don't see the loop yet here either. I could go deeper, but as a sample "python AIR developer" I'm trying to see if I can use the python without being an expert in mlir.

I guess my question here is philosophical:
* Do we want to present this behavior as 'trust the compiler, you don't need to know what it's doing'
* Or do we want to say, 'you can do this, but the behavior is ambiguous so it's best to be specific at the python level instead'? (a which point maybe I'd remove this example from the PR entirely).

programming_examples/matrix_scalar_add/README.md
```diff
@@ -0,0 +1,78 @@
+# Matrix Scalar Addition
```
hunhoffe, line 1: I discuss tiling at different layers of abstraction but I don't discuss temporal vs spacial at all here. @jgmelber can you help me understand what spacial/temporal settings for tiling I could add as examples or write about here in the readme?

programming_examples/matrix_scalar_add/matrix_scalar_add_implicit.py
fifield: There is a `while(1)` loop generated at the top level of every `aie.core` because of the `emit-while-loop` option getting set to `true` here: https://github.com/Xilinx/mlir-air/blob/main/python/air/compiler/aircc/main.py#L399 (if you are compiling via aircc, which you are if you're using the xrt wrapper). I agree it's confusing and I had it set to `false` until a few weeks ago.

The reason that I set it to `true` is for benchmarking and I assume it's the same reason there's a `for i = 0...int_max` in many of the mlir-aie examples. Our usual benchmarking host code will typically do some number of warmup iterations of the kernel followed by some number of timed iterations. The problem I had was that xrt did not reconfigure the aie array between calls of the same kernel. So, if I programmed the aie cores to do exactly one invocation worth of tiles, then each core would happily do so for the first invocation, halt, then everything would hang the second time around.

hunhoffe: Ah! Thanks for the explanation.

Would it make sense to disable ```emit-while-loop``` again and then add some sort of `--benchmarking` option in aircc that enables it on demand?

fifield: The option should be exposed to the aircc user, but there are use cases other than benchmarking.

fifield:
> * Wrap `AllocOp` and `DeallocOp` to clean up `AllocOp` interface, and then wrap `DeallocOp` to maintain a symmetric interface
> * Add default values for the `ChannelGet`/`ChannelPut` operations, as well as automatic conversion from python ints where it seems useful.

It would be nice to make these their own PR(s).

> ```python 
> from air.dialects.air import Alloc

I'm a bit hesitant to wrap (i.e. diverge from) upstream operators to save a couple characters of typing. `Alloc` is not an air dialect op. Do we already do this to `AllocOp` in mlir-aie? Or maybe everything is objfifo there...

hunhoffe:
> > * Wrap `AllocOp` and `DeallocOp` to clean up `AllocOp` interface, and then wrap `DeallocOp` to maintain a symmetric interface
> > * Add default values for the `ChannelGet`/`ChannelPut` operations, as well as automatic conversion from python ints where it seems useful.
> 
> It would be nice to make these their own PR(s).
> 
> > ```python
> > from air.dialects.air import Alloc
> > ```
> 
> I'm a bit hesitant to wrap (i.e. diverge from) upstream operators to save a couple characters of typing. `Alloc` is not an air dialect op. Do we already do this to `AllocOp` in mlir-aie? Or maybe everything is objfifo there...

I believe the aie uses AllocOp internally but users at the python level do not use it (at least, I can't find an example of direct use in the examples in the mlir-aie repo).

My thinking was that it's best not to expose options/interfaces that are basically wholly undocumented in air because if you leave them open, the only way to know how to use AllocOp/DeallocOp in this context is to copy/paste from an existing example and hope for the best - which doesn't exactly build programmer confidence.

However, I'm happy to remove/table the AllocOp/DeallocOp changes. And I'm also happy to put the ChannelGet/ChannelPut changes in a separate PR as well for further discussion.

fifield:
> 
> My thinking was that it's best not to expose options/interfaces that are basically wholly undocumented in air because if you leave them open, the only way to know how to use AllocOp/DeallocOp in this context is to copy/paste from an existing example and hope for the best - which doesn't exactly build programmer confidence.
> 
 
Upstream dialects and bindings are documented upstream, e.g. https://mlir.llvm.org/docs/Dialects/MemRef/, https://mlir.llvm.org/docs/Bindings/Python. Maybe usability and documentation improvements can be proposed there?

hunhoffe:
Added separate PR for toggling the aircc `emit-while-loop` option: https://github.com/Xilinx/mlir-air/pull/613

hunhoffe:
> > My thinking was that it's best not to expose options/interfaces that are basically wholly undocumented in air because if you leave them open, the only way to know how to use AllocOp/DeallocOp in this context is to copy/paste from an existing example and hope for the best - which doesn't exactly build programmer confidence.
> 
> Upstream dialects and bindings are documented upstream, e.g. https://mlir.llvm.org/docs/Dialects/MemRef/, https://mlir.llvm.org/docs/Bindings/Python. Maybe usability and documentation improvements can be proposed there?

Okay, yes, that makes sense! I believe I was thinking about this more as a Python programmer and less like an MLIR programmer. Since we import this operations through `air.dialects...` it makes me think it's an air specific python packages and not a representation of a library that exists outside of MLIR-AIR. But I believe this import structure is an artifact of the MLIR build structure (that the dialect python code is generated uniquely for this repo), so it's probably fine as-is.

hunhoffe:
Closing this PR in favor of:
* aircc option for `emit-while-true-loop`: https://github.com/Xilinx/mlir-air/pull/613
* matrix scalar add example: https://github.com/Xilinx/mlir-air/pull/621
* Improvements to ChannelGet and ChannelPut python wrappers: https://github.com/Xilinx/mlir-air/pull/622

PR #590, opened 2024-06-04
Add Python Programming Example

hunhoffe:
Add a python programming example that is checked end-to-end by the CI workflow. This example includes multiple workflows for how to run the code (defined using AIR python bindings) on an NPU.

To support some of the additional passes used in test/xrt (including those for dependencies), I added a new argument, ```--experimental```, to ```aircc.py```, which will run those additional passes. It works on my example but I suspect that this will need to be changed/updated as more examples are made but I believe what I have is a reasonable starting point.

To simplify the example, I also added a wrapper around the python ```dma_memcpy_nd``` function to supply default values and allow the caller to use python integers.

PR #595, opened 2024-06-06
delete old examples directory

fifield:
`examples` is replaced by `programming_examples` directory (#590).  The new folder is tested in the CI and known to work which is better than the old `examples` directory. The best way I can think of to clean up `examples` is to delete it. If anyone still wants one of the old examples we can resurrect it and put it in the new area.


erwei-xilinx:
I agree with removing the air-runner and air-dependency examples as they are obsolete. Might also worth checking with Gagan and Eddie on the weather stencil examples.

jgmelber (approved):
LGTM. We will try to capture a list of relevant `examples` we want to redevelop as `programming_examples`.

- Channels examples
- Attention head (airbert)
- matmul gelu (and more activations)
- MLP
- matmul
- conv
- weather stencil

@hunhoffe @denolf @singagan @abisca

jgmelber:
> I agree with removing the air-runner and air-dependency examples as they are obsolete. Might also worth checking with Gagan and @eddierichter-amd on the weather stencil examples.

Agreed, I think the interesting _AIR_ parts from the weather stencil are captured here: https://github.com/Xilinx/ROCm-air-platforms/tree/main/examples/sparta-weather-stencil

eddierichter-amd:
> > I agree with removing the air-runner and air-dependency examples as they are obsolete. Might also worth checking with Gagan and @eddierichter-amd on the weather stencil examples.
> 
> Agreed, I think the interesting _AIR_ parts from the weather stencil are captured here: https://github.com/Xilinx/ROCm-air-platforms/tree/main/examples/sparta-weather-stencil

Agreed, removing the weather stencil in this repo is fine with me.

PR #600, opened 2024-06-10
Implement python decorator to abstract module InsertionPoint

hunhoffe:
This PR introduces a new python decorator, ```module_builder```, which may be used to help create a module of code using the AIR python bindings. The purpose of this decorator is to help simplify the air python bindings for the benefit of the user.

As an example, I changed ```programming_examples/shim_dma_2d/shim_dma_2d.py``` to use this new decorator, as well as any tests that seem to create a module using python bindings.

It turns out there is already an ```@module``` decorator which is builtin with the MLIR Python bindings. However, I couldn't find much documentation on it. When I finally got it to work (and I'm not 100% sure this is how it's meant to be used, but it's my best attempt), the end result seemed a bit unintuitive so I think it's worth defining and using a new decorator.

#### Using new ```@module_builder``` decorator

```python
@module_builder
def build_module():
    memrefTyInOut = MemRefType.get(IMAGE_SIZE, T.i32())

    # We will send the image worth of data in and out
    @FuncOp.from_py_func(memrefTyInOut, memrefTyInOut)
    def copy(arg0, arg1):
        ...

if __name__ == "__main__":
    module = build_module()
    print(module)
```

#### Using builtin ```@module``` decorator
```python
def build_module(loc):
    @module(loc=loc)
    def my_module():
        memrefTyInOut = MemRefType.get(IMAGE_SIZE, T.i32())

        # We will send the image worth of data in and out
        @FuncOp.from_py_func(memrefTyInOut, memrefTyInOut)
        def copy(arg0, arg1):
            ...
    return my_module

if __name__ == "__main__":
    with Context() as ctx, Location.unknown() as loc:
        my_module = build_module(loc)
    print(my_module)
```

Two things I don't like about using the ```@module``` decorator are that:
* `loc` is not a positional argument, so I need to be able to say ```loc=<some_var>``` at the time of function definition. To ensure there is a `loc` variable in scope, I had to keep an extra layer of function definitions (`build_module()` AND `my_module()`, instead of just one)
* At the end of the `build_module()` function, you must `return my_module` (no explicit invocation) instead of `return my_module()`. This seems unintuitive.


python/air/dialects/_air_ops_ext.py
```diff
@@ -227,6 +229,18 @@ def __init__(
 dma_memcpy_nd = DmaMemcpyNd
 
 
+def air_module(module_function):
+    @functools.wraps(module_function)
+    def air_module_wrapper(*args, **kwargs):
+        with Context() as ctx, Location.unknown():
```
hunhoffe, line 235: @jgmelber I am unsure if it makes sense to encapsulate the context within this decorator or not. In the spirit of removing boilerplate, it did seem to make sense for the programming examples. However, this then causes problems with some of the structure of ```tests/xrt/01``` though ```tests/xrt/01``` which perform additional logic/transforms within the context (if I am reading those tests correctly).

It's possible I could make an additional decorator or parameterize this decorator to take as an argument an additional optional parameter, a function pointer which could be applied to the module within the context.
What do you think?

python/air/dialects/_air_ops_ext.py
```diff
@@ -227,6 +229,18 @@ def __init__(
 dma_memcpy_nd = DmaMemcpyNd
 
 
+def air_module(module_function):
```
AndraBisca, line 232: I think the name `air_module` may be misleading as the ModuleOp comes from the `BuiltIn` MLIR dialect. You can just name it `module`.

AndraBisca: Ah the builtin dialect already has a `module` python binding as well.

python/air/dialects/_air_ops_ext.py
jgmelber: @erwei-xilinx @fifield would have the most informed opinion on this

python/air/dialects/_air_ops_ext.py
hunhoffe: I'm very open to other suggestions, but I think I'll try to rename it ```module_builder``` unless there are more suggestions!

hunhoffe: @AndraBisca Thanks also for pointing out the ```module``` decorator! I updated the PR description to reflect my experiences with it.

TL;DR - it works, but doesn't seem very clean so I think it's still worth defining a new decorator. If you have any further thoughts (or if you suspect I am using ```@module``` incorrectly), please let me know!

python/air/dialects/_air_ops_ext.py
hunhoffe: I figured out that I could get the context manually from ```module.context``` so I went ahead and moved the decorator to all the tests I found that seemed like they could use it. The only real downside I see is that it's hard to instantiate some typing parameters (the ```dtype``` argument in some of the ```test/xrt``` tests I modified)... but that parameter was hardcoded in all the tests so it doesn't seem like it's a current issue. I think there are ways around this issue, for example, by passing a ```dtype_representor``` (maybe a ```str``` or an ```int``` instead of a ```dtype``` directly) as an argument which could then get converted into the right type within the function wrapped by ```@module_builder```

However, I'd still welcome any feedback on this overall!

AndraBisca: Thank you Erika for your thorough analysis. It seems that your `module_builder` results in a cleaner syntax overall so I think we can proceed with it.

jgmelber (approved):

PR #596, opened 2024-06-06
Update for mlir-aie switch to transaction ops

fifield:
update mlir-aie to https://github.com/Xilinx/mlir-aie/pull/1517

PR #601, opened 2024-06-10
fix trace buffer id

Yu-Zhewen:
bdID should be a reference for the correct decrement


erwei-xilinx (approved):

mlir/lib/Conversion/AIRRtToNpuPass.cpp
```diff
@@ -1371,7 +1371,7 @@ struct AIRRtToNpuPass : public impl::AIRRtToNpuBase<AIRRtToNpuPass> {
         // configure shim tile
         if (chanToIdMap.count(dstColIndex) == 0)
           chanToIdMap[dstColIndex] = 15;
-        int bdID = chanToIdMap[dstColIndex];
+        int &bdID = chanToIdMap[dstColIndex];
```
fifield, line 1374: Maybe `chanToIdMap[dstColIndex]--` would be more clear to future readers of the code?

fifield: > Maybe `chanToIdMap[dstColIndex]--` would be more clear to future readers of the code?

fixed in https://github.com/Xilinx/mlir-air/pull/596

PR #613, opened 2024-06-20
Add option to aircc to toggle emit-while-loop=true; fix bug

hunhoffe:
Add option to aircc to toggle emit-while-loop=true; fix bug for --experimental aircc flag


fifield (approved):
LGTM

PR #619, opened 2024-06-25
Move dma to channel pass from Conversion passes to Transform passes

fifield:
`air-dma-to-channel` is a transformation on air dialect, but it grew inside `air/Conversion/ConvertToAIR`. This PR moves it to `air/Transform/AIRDmaToChannel`. Hopefully NFC cut-n-paste only.


erwei-xilinx (approved):
Moving that pass to Transform passes makes sense to me. Thanks.

PR #620, opened 2024-06-26
Fixup AIRRt lowerings to support convolution examples

erwei-xilinx:
- Some existing tests for `-air-to-std` were modified because in `air.channel.put/get` we follow `memref.subview` convention for strides, where zero means repeating.

### DIFF
--- a/mlir/lib/Conversion/AIRLoweringPass.cpp
+++ b/mlir/lib/Conversion/AIRLoweringPass.cpp
@@ -453,8 +453,6 @@ AIRChannelInterfaceToAIRRtConversionImpl(OpBuilder builder,
   auto i64Ty = builder.getI64Type();
   auto zero =
       builder.create<arith::ConstantOp>(loc, i64Ty, IntegerAttr::get(i64Ty, 0));
-  auto one =
-      builder.create<arith::ConstantOp>(loc, i64Ty, IntegerAttr::get(i64Ty, 1));
   auto zero_idx = builder.create<arith::ConstantIndexOp>(loc, 0);
   auto one_idx = builder.create<arith::ConstantIndexOp>(loc, 1);
 
--- a/mlir/lib/Conversion/AIRLoweringPass.cpp
+++ b/mlir/lib/Conversion/AIRLoweringPass.cpp
@@ -455,14 +455,15 @@ AIRChannelInterfaceToAIRRtConversionImpl(OpBuilder builder,
       builder.create<arith::ConstantOp>(loc, i64Ty, IntegerAttr::get(i64Ty, 0));
   auto one =
       builder.create<arith::ConstantOp>(loc, i64Ty, IntegerAttr::get(i64Ty, 1));
+  auto zero_idx = builder.create<arith::ConstantIndexOp>(loc, 0);
+  auto one_idx = builder.create<arith::ConstantIndexOp>(loc, 1);
 
   auto idTy = IntegerType::get(ctx, 32);
   // Get op id of the internal put/get op
   if (auto id_attr = theOtherOp->getAttrOfType<IntegerAttr>("id")) {
     opers.push_back(builder.create<arith::ConstantOp>(loc, idTy, id_attr));
   } else {
-    opers.push_back(builder.create<arith::ConstantOp>(
-        loc, idTy, IntegerAttr::get(idTy, 0)));
+    opers.push_back(zero);
   }
 
   scf::ParallelOp launch = thisOp->getParentOfType<scf::ParallelOp>();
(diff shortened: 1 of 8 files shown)


mlir/test/Conversion/AIRLowering/air_channel_get_put.mlir
```diff
@@ -24,25 +24,26 @@ module {
       %c32 = arith.constant 32 : index
       %c1 = arith.constant 1 : index
       %c0 = arith.constant 0 : index
-      %0 = air.channel.put async  @channel_0[%c0, %c0] (%arg0[%c8, %c0] [%c8, %c16] [%c32, %c0]) {id = 1 : i32} : (memref<32x16xi32>)
-      %1 = air.channel.get async  @channel_1[%c0, %c0] (%arg1[%c8, %c0] [%c8, %c16] [%c32, %c0]) {id = 2 : i32} : (memref<32x16xi32>)
+      %0 = air.channel.put async  @channel_0[%c0, %c0] (%arg0[%c8, %c0] [%c8, %c16] [%c32, %c1]) {id = 1 : i32} : (memref<32x16xi32>)
+      %1 = air.channel.get async  @channel_1[%c0, %c0] (%arg1[%c8, %c0] [%c8, %c16] [%c32, %c1]) {id = 2 : i32} : (memref<32x16xi32>)
```
fifield, line 28: Are you changing the semantics of get and put in this PR? Please give a longer description of what you are changing.

mlir/lib/Conversion/AIRLoweringPass.cpp
```diff
@@ -489,45 +488,71 @@ AIRChannelInterfaceToAIRRtConversionImpl(OpBuilder builder,
...
+  // Stride field implicit last element one
+  auto lastStrideConst = getConstantIntValue(strides.back());
+  assert(lastStrideConst && "the last stride is not static");
+  // If the last dimension's stride value is not 1, then for AIE2 we use the
+  // second dimension of shim dma bd to implement the last dimension.
+  if (*lastStrideConst != 1) {
+    offsets.push_back(zero_idx);
+    wraps.push_back(one_idx);
+    strides.push_back(one_idx);
+  }
```
fifield, line 521: I don't follow the description. Are we doing something AIE2 specific?

mlir/lib/Conversion/AIRLoweringPass.cpp
```diff
@@ -489,45 +488,71 @@ AIRChannelInterfaceToAIRRtConversionImpl(OpBuilder builder,
...
+  while (strides.size() < 3) {
+    strides.insert(strides.begin(), zero_idx);
+  }
+
+  for (unsigned i = 0; i < offsets.size(); i++)
+    offsets[i] = builder.create<arith::IndexCastOp>(
+        loc, IntegerType::get(ctx, 64), offsets[i]);
+
+  // In aiex.npu ops, stride value 0 means 1; only the highest dimension stride
+  // value 0 really means repeat.
```
fifield, line 539: Are we doing something runtime (npu) specific here?

test/xrt/13_conv2d_i32/aie.py
```diff
@@ -0,0 +1,202 @@
...
+from air.compiler.util import run_transform
+import sys
+
+with air.ir.Context() as ctx, Location.unknown():
+
+    ################################################
+    ## Tiling
+    ################################################
+
+    air_tiled_ir_string = """
```
fifield, line 21: Is it possible to generate this via tiling from linalg or python bindings? And if not, why not have a conv2d.mlir input file instead of imbedding as a string?  Then one can at least look at it with syntax highlighting.

mlir/lib/Conversion/AIRRtToNpuPass.cpp
```diff
@@ -137,22 +137,32 @@ struct DmaToNpuPattern : public OpConversionPattern<DmaMemcpyNdOp> {
           .getResult();
     };
     SmallVector<Value> offsets;
-    SmallVector<int64_t> staticOffsets;
-    if (auto const_int = getConstantIntValue(adaptor.getOffset3()))
-      staticOffsets.push_back(*const_int);
-    else
+    SmallVector<int64_t>
+        staticOffsets; // Note: for static offsets we compose one single offset
+                       // at the last dimension.
```
fifield, line 142: What is the motivation for this? Should it be a canonicalization in mlir-aie instead?

mlir/test/Conversion/AIRLowering/air_channel_get_put.mlir
erwei-xilinx: Sure. Some of the original tests, where the last stride value was set to 0, was using inconsistent semantics compared with the rest of the tests. The original intent of these tests should be "accessing data in row-major, consecutive" order in the inner most stride dimension. That means the last stride value should be 1.

Those tests could get away with that inconsistency because the last stride dimension is simply thrown away, as the downstream aiex.npu.dma op only has stride dimensions from 1 to 3, where the lowest dimension is assumed to be always 1.

mlir/lib/Conversion/AIRLoweringPass.cpp
erwei-xilinx: I noticed that in this line https://github.com/Xilinx/mlir-aie/blob/d850560c77799af96c6361a79e34cb0a8e842c50/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L302 the d0_stride is always set to 0, meaning that the innermost dimension must always be accessing data in "row-major, consecutive" order. I do not know if this was intended for constraint of any specific architecture version.

mlir/lib/Conversion/AIRLoweringPass.cpp
erwei-xilinx: In this line https://github.com/Xilinx/mlir-aie/blob/d850560c77799af96c6361a79e34cb0a8e842c50/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L305 aiex.npu.dma has two schemes on assigning registers, depending on if stride[i] is zero or not. I think this is to get around with the fact that in the hardware the register field is taking "strides[i] - 1".

test/xrt/13_conv2d_i32/aie.py
erwei-xilinx: Yeah this test should definitely be possible to be converted to having input IR generated by Transform Dialect or python bindings. Either choice makes great sense to me. This board test is just a placeholder to verify the functionality of convolution from tiled IR downwards. We should definitely come back to the test and change the input IR format.

mlir/lib/Conversion/AIRRtToNpuPass.cpp
erwei-xilinx: I noticed an issue in mlir-aie, where the multi-dimensional offsets were composed to the BD's base offset using memref shape: https://github.com/Xilinx/mlir-aie/blob/d850560c77799af96c6361a79e34cb0a8e842c50/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L284

I think this would not work when the memref shape and data movement wraps do not match. We have never had this issue before because with matmul they would always match; direct code-generated conv2d starts to trigger this issue.

So to get around it I compose the offsets in mlir-air, into one single static offset at the lowest dimension.

mlir/lib/Conversion/AIRLoweringPass.cpp
fifield: It's set to zero because it is lowering `airrt.dma_memcpy_nd` which implicitly sets it to zero (or *actual - 1*). The op literally does not have a d0 stride operand (that's why there's three strides vs. four wraps/offsets in the c++ code). See here: https://github.com/Xilinx/mlir-air/blob/main/mlir/include/air/Dialect/AIRRt/AIRRtOps.td#L131. Why? Because AIE1 could only do 1d contiguous memcpy and everything else was emulated in firmware. Then npu lowering was bolted onto that.

mlir/lib/Conversion/AIRLoweringPass.cpp
fifield: I still don't understand what this code is trying to accomplish. The lowering from `air` to `airrt` shouldn't be concerned with the semantics of the `aiex.npu` op it *might* eventually lower to. The lowering should only care about the semantics of `airrt.dma_memcpy_nd`.

mlir/lib/Conversion/AIRRtToNpuPass.cpp
fifield: Why don't we fix the mlir-aie code to work in the general case instead of adding workarounds?

mlir/test/Conversion/AIRLowering/air_channel_get_put.mlir
fifield: I see, thanks. The lowering always ignores the inner stride so there was never any issue.

mlir/lib/Conversion/AIRLoweringPass.cpp
erwei-xilinx: Thanks for the information on why this is the case. I reverted this change in https://github.com/Xilinx/mlir-air/pull/626

mlir/lib/Conversion/AIRLoweringPass.cpp
erwei-xilinx: Make sense. Reverted in https://github.com/Xilinx/mlir-air/pull/626

mlir/lib/Conversion/AIRRtToNpuPass.cpp
erwei-xilinx: Make sense. Seems related: https://github.com/Xilinx/mlir-aie/issues/1578

mlir/lib/Conversion/AIRLoweringPass.cpp
fifield: https://github.com/Xilinx/mlir-aie/pull/1584 is intended to simplify the situation

PR #622, opened 2024-06-27
Cleaner ChannelGetOp and ChannelPutOp Python Wrappers

hunhoffe:
This PR seeks to improve readability of python code using ChannelGetOp and ChannelPutOp wrappers and also seeks to make it easier to use these wrappers.

Readability improvements come from forcing the callee to explicitly label `(src|dst)_(strides|sizes|offsets)` labels in their code. By requiring explicit labels, this can help a reader better understand how these mechanisms are used without having to lookup or memorize the API itself.

Usability improvement is from implementing behind-the-scenes conversion from python integers to Index values.


python/air/dialects/_air_ops_ext.py
```diff
@@ -106,30 +107,29 @@ class ChannelGet(ChannelGetOp):
...
             dst=dst,
-            dst_offsets=dst_offsets,
-            dst_sizes=dst_sizes,
-            dst_strides=dst_strides,
+            dst_offsets=dst_offsets_typed,
+            dst_sizes=dst_sizes_typed,
+            dst_strides=dst_strides_typed,
             loc=loc,
             ip=ip,
         )
```
fifield, line 135: I think we could drop the `dst_` and `src_` prefixes on the wrapper. The get and put ops define getOffset, etc. as part of the `ChannelInterface` ([here](https://github.com/Xilinx/mlir-air/blob/main/mlir/include/air/Dialect/AIR/AIROpBase.td#L184)) so the naming wouldn't be inconsistent with that. The prefixes on the operand names are probably there to avoid conflicting with the `ChannelInterface` but I'm not sure both are needed.

fifield (approved):
This is a nice improvment.

python/air/dialects/_air_ops_ext.py
hunhoffe: Makes sense to me! I removed the `src_`/`dst_` prefixes

PR #629, opened 2024-06-28
Run black python formatter on test directory

hunhoffe:
I'm not sure why the CI doesn't catch this... maybe these files predate the CI formatting job? Anyways, formatting here for consistency!


fifield (approved):
Thanks!

fifield:
> I'm not sure why the CI doesn't catch this... maybe these files predate the CI formatting job?

Yes, that is why. The CI only checks new patches and python format checking is fairly recent.

PR #631, opened 2024-06-28
Add spport for rank-reduced subview in memref shrinkage

erwei-xilinx:


mlir/lib/Util/Util.cpp
```diff
@@ -1248,6 +1248,12 @@ air::writeAccessPattern(memref::SubViewOp subview) {
     else
       std::get<1>(pattern).push_back(*subview_sizes++);
   }
+  assert(static_sizes.size() >= static_strides.size() &&
+         "only support standard and rank-reduced memref.subview");
```
fifield, line 1252: It looks like ultimately `writeAccessPattern` is called from `ShrinkMemrefSizesByAccessPattern` which could return failure + diagnostics instead of asserting (in Debug mode) or crashing in unexplained ways (in Release mode).

erwei-xilinx: Make sense. I'll PR later today to rewrite that with emitOpError.

erwei-xilinx: Fixed in https://github.com/Xilinx/mlir-air/pull/632

PR #634, opened 2024-07-01
XRTBackend API Cleanup

hunhoffe:
There are a few things about the `XRTBackend` that are a bit unintuitive:
* You can't actually set the insts file path (unused parameters)
* There are unneeded parameters (`module` in `load()`)
* There aren't safety checks for the state that is saved in the backend during `load()`

This PR cleans up a few aspects of `XRTBackend` to ensure parameters make sense and are used as one would expect.


fifield (approved):
Looks good, thanks.

Issue #625, opened 2024-06-27
Multi Core Channel Matrix Scalar Add Example Fails


hunhoffe:
From branch `debugging_matrix_scalar_add`, I am working on getting my example `multi_core_channel` working ([this](https://github.com/Xilinx/mlir-air/blob/debugging-matrix-scalar-add/programming_examples/matrix_scalar_add/multi_core_channel/multi_core_channel.py) file). The single core version works in this branch, but when I increase the herd size from 1x1 to 2x2, the example does not work any more.

To be specific, to replicate run:
```bash
cd programming_examples/matrix_scalar_add/multi_core_channel
make
```

When I inspect `programming_examples/matrix_scalar_add/multi_core_channel/build/air_project/npu.air.mlir`, I notice that it looks like all of the data is only going to one core instead of being distributed to all of the cores.

```
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 0][1, 1, 8, 16][0, 0, 32]) {id = 0 : i64, metadata = @airMemcpyId3} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg1[0, 0, 0, 0][1, 1, 8, 16][0, 0, 32]) {id = 1 : i64, metadata = @airMemcpyId4} : memref<32x16xi32>
      aiex.npu.sync {channel = 0 : i32, column = 0 : i32, column_num = 1 : i32, direction = 0 : i32, row = 0 : i32, row_num = 1 : i32}
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 16][1, 1, 8, 16][0, 0, 32]) {id = 0 : i64, metadata = @airMemcpyId3} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg1[0, 0, 0, 16][1, 1, 8, 16][0, 0, 32]) {id = 1 : i64, metadata = @airMemcpyId4} : memref<32x16xi32>
      aiex.npu.sync {channel = 0 : i32, column = 0 : i32, column_num = 1 : i32, direction = 0 : i32, row = 0 : i32, row_num = 1 : i32}
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 16, 0][1, 1, 8, 16][0, 0, 32]) {id = 0 : i64, metadata = @airMemcpyId3} : memref<32x16xi32>
[...]

Issue #624, opened 2024-06-27
Multi Core DMA Matrix Scalar Add Example Fails


hunhoffe:
From branch `debugging_matrix_scalar_add`, I am working on getting my example `multi_core_dma` working ([this](https://github.com/Xilinx/mlir-air/blob/debugging-matrix-scalar-add/programming_examples/matrix_scalar_add/multi_core_dma/multi_core_dma.py) file). The single core version works in this branch, but when I increase the herd size from 1x1 to 2x2, the example does not work any more.

To be specific, to replicate run:
```bash
cd programming_examples/matrix_scalar_add/multi_core_dma
make
```

When I inspect `programming_examples/matrix_scalar_add/multi_core_dma/build/air_project/npu.air.mlir`, I notice that it looks like all of the data is only going to one core instead of being distributed to all of the cores.

```
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 0][1, 1, 8, 16][0, 0, 32]) {id = 0 : i64, metadata = @airMemcpyId9} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 16, 0][1, 1, 8, 16][0, 0, 32]) {id = 1 : i64, metadata = @airMemcpyId9} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 16][1, 1, 8, 16][0, 0, 32]) {id = 2 : i64, metadata = @airMemcpyId9} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 16, 16][1, 1, 8, 16][0, 0, 32]) {id = 3 : i64, metadata = @airMemcpyId9} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg1[0, 0, 0, 0][1, 1, 8, 16][0, 0, 32]) {id = 4 : i64, metadata = @airMemcpyId10} : memref<32x16xi32>
      aiex.npu.dma_memcpy_nd(0, 0, %arg1[0, 0, 16, 0][1, 1, 8, 16][0, 0, 32]) {id = 5 : i64, metadata = @airMemcpyId10} : memref<32x16xi32>
[...]

PR #637, opened 2024-07-02
Refactor AIRRt symbolic linkage matching logic

erwei-xilinx:
- Fixup to enable a case where there are multiple unrolled sub-channels on the shim side, directly accessing buffers in one air.herd.

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -2226,32 +2226,56 @@ public:
   void annotateMetadataPerShimAIRChannel(air::ChannelInterface chan_o,
                                          MemRefType memref_ty,
                                          StringAttr dma_name_attr) {
-    auto originalIndices = chan_o.getIndices();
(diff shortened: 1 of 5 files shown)


fifield:
This PR changes the examples. Is that because they had bugs or is it to workaround bugs in the compiler?

hunhoffe:
My understanding is that there were bugs in the examples, and a few places where I incorrectly computed a value that just happened to have a correct value. Erwei fixed it so the values are computed correctly. I will review the other examples (the ones that work) to check to see if I'm getting accidental success elsewhere too.

My understanding is that this PR fixes one of the failing examples (multi core dma, associated with this issue: https://github.com/Xilinx/mlir-air/issues/624) but I will need to rewrite the multi core channel example to better fit the channel abstraction/implementation (changes needed described in comments here: https://github.com/Xilinx/mlir-air/issues/625).

hunhoffe (approved):
Thanks for the fix in AIRRt and also the fixes in the examples!

PR #644, opened 2024-07-02
Refactor AIRLowering pass with mlir typeConverter

erwei-xilinx:


fifield (approved):

python/air/compiler/aircc/main.py
```diff
@@ -28,6 +28,7 @@
     "air-dependency",
     "air-dependency-schedule-opt",
     "air-specialize-dma-broadcast",
+    "air-dma-to-channel",
```
fifield, line 31: doesn't it already contain `air-dma-to-channel`? Do you just need a different order?
Starting about line 375:
```python
        pass_pipeline = ",".join(
            [
                "air-insert-launch-and-segment-around-herd",
                "func.func(air-lower-herd-parallel)",
            ]
            + (
                EXPERIMENTAL_PASSES
                if "npu" in opts.device and opts.experimental_passes
                else []
            )
            + (["air-dma-to-channel"] if "npu" in opts.device else [])
            + [
                "canonicalize",
                "cse",
                "air-specialize-channel-wrap-and-stride",
                "func.func(air-renumber-dma)",
                "func.func(convert-linalg-to-loops)",
                air_place_pass,
            ]
```

erwei-xilinx: Yes we need `air-dma-to-channel` to produce us with `air.channel` ops before the following passes: 
```
    "func.func(air-loop-fusion)",
    "air-label-scf-for-to-ping-pong",
    "air-ping-pong-transform{keep-memref-dealloc=true}",
    "air-isolate-async-dma-loop-nests",
```
Those passes are for BD optimizations, and they only work on `air.channels`, because `air.dma_memcpy_nd` doesn't have explicit handles for BDs.

PR #647, opened 2024-07-03
Minor AIRLowering refactor

fifield:
This is two cleanups I've had sitting in branches for a while:
* Deduplicate get/put lowering patterns
* Use more clone + IRMapping instead of replaceuseswith + splice. This fixed some segfaults I experienced when working on #585

### DIFF
--- a/mlir/test/Conversion/AIRLowering/air_launch.mlir
+++ b/mlir/test/Conversion/AIRLowering/air_launch.mlir
@@ -8,20 +8,15 @@
 // RUN: air-opt -air-to-std %s | FileCheck %s
 
 // CHECK-LABEL:   func.func @launch_0(
-// CHECK-SAME:                        %[[VAL_0:.*]]: memref<16xf16>,
-// CHECK-SAME:                        %[[VAL_1:.*]]: memref<16xf16>) {
-// CHECK:           %[[VAL_2:.*]] = arith.constant 2 : index
-// CHECK:           %[[VAL_3:.*]] = arith.constant 4 : index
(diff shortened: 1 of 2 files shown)


mlir/test/Conversion/AIRLowering/air_launch.mlir
```diff
@@ -8,20 +8,15 @@
...
+// CHECK-SAME:               %{{.*}}: memref<16xf16>,
+// CHECK-SAME:               %{{.*}}: memref<16xf16>) {
+// CHECK:       affine.for %{{.*}} = 0 to 4 {
+// CHECK:       affine.for %{{.*}} = 0 to 2 {
+// CHECK:       affine.for %{{.*}} = 0 to 2 {
 func.func @launch_0(%arg0: memref<16xf16>, %arg1: memref<16xf16>) {
   %c2 = arith.constant 2 : index
   %c4 = arith.constant 4 : index
   air.launch (%arg2, %arg3, %arg4) in (%arg5=%c4, %arg6=%c2, %arg7=%c2) args(%arg8=%arg0, %arg9=%arg1) : memref<16xf16>, memref<16xf16> {
-    air.dma_memcpy_nd (%arg9[] [] [], %arg8[] [] []) : (memref<16xf16>, memref<16xf16>)
```
fifield, line 23: When cloning ops, the new op must be legalizable. This one isn't because it's a L3->L3 copy which currently returns `failure()` in the `dma_memcpy_nd` conversion pattern.  So just remove it since it's not important to the test.

PR #652, opened 2024-07-05
Enable `air.channel.put/get` through broadcast detection

erwei-xilinx:


hunhoffe (approved):
Great, thanks!

There are now additional examples that are not currently run with experimental passes because they use channels. Specifically, this example [here](https://github.com/Xilinx/mlir-air/blob/5730bf610bdd7c6ac49b73e9e320d745af3cb8cf/programming_examples/data_transfer_transpose/channel/run.py#L47) and this example [here](https://github.com/Xilinx/mlir-air/blob/5730bf610bdd7c6ac49b73e9e320d745af3cb8cf/programming_examples/channel_examples/common.py#L37) may be able to use experimental passes with this change.

If these additional change are out of scope for this PR, I am happy to create a separate PR with these changes.

erwei-xilinx:
> Great, thanks!
> 
> There are now additional examples that are not currently run with experimental passes because they use channels. Specifically, this example [here](https://github.com/Xilinx/mlir-air/blob/5730bf610bdd7c6ac49b73e9e320d745af3cb8cf/programming_examples/data_transfer_transpose/channel/run.py#L47) and this example [here](https://github.com/Xilinx/mlir-air/blob/5730bf610bdd7c6ac49b73e9e320d745af3cb8cf/programming_examples/channel_examples/common.py#L37) may be able to use experimental passes with this change.
> 
> If these additional change are out of scope for this PR, I am happy to create a separate PR with these changes.

The `data_transfer_transpose` test works with `experimental_passes` now. `channel_examples` failed with another error. Will come back to look into it.

PR #655, opened 2024-07-08
External Function Call and Passthrough Examples

hunhoffe:
The most important contribution of this PR is to add an example of calling an external function in the mlir-air Python bindings. This is implemented in the `programming_examples/passthrough/passthrough_kernel` example. Small changes to HerdOp were necessary to support this, as well as the addition of some helper functions similar to those in mlir-aie.

The documentation for the passthrough example is mostly taken from mlir-aie. For completeness, I also include examples doing data passthrough with data movement with DMAs and channels.


programming_examples/makefile-common
```diff
@@ -0,0 +1,20 @@
...
+CHESS_FLAGS = -P ${VITIS_AIE_INCLUDE_DIR}
+
+CHESSCCWRAP2_FLAGS = aie2 -I ${VITIS_AIETOOLS_DIR}/include 
+
+TEST_POWERSHELL := $(shell command -v powershell.exe >/dev/null 2>&1 && echo yes || echo no)
+ifeq ($(TEST_POWERSHELL),yes)
+	powershell = powershell.exe
+else
+	powershell = 
+endif
```
erwei-xilinx, line 20: (Nitpick) EoF.

erwei-xilinx (approved):

mlir/include/air/Dialect/AIR/AIR.td
```diff
@@ -174,6 +174,7 @@ def air_HerdOp : air_Op<"herd", [air_AsyncOpInterface,
                                  AffineScope,
                                  SingleBlockImplicitTerminator<"HerdTerminatorOp">]>,
                         Arguments<(ins OptionalAttr<SymbolNameAttr>:$sym_name,
+                                       OptionalAttr<StrAttr>:$link_with,
```
fifield, line 177: Was this required to make the python work? Or just a convenience?

erwei-xilinx: If the `link_with` attribute is set in `air.herd`, then when aie cores are outlined in `air-to-aie`, every core of that herd shall link with the object file. https://github.com/Xilinx/mlir-air/blob/2754dcf53e9dce76df4e1a8d5f34bcb2acd1871e/mlir/lib/Conversion/AIRToAIEPass.cpp#L259

erwei-xilinx: So my understanding is it's a convenience...

fifield: > If the `link_with` attribute is set in `air.herd`, then when aie cores are outlined in `air-to-aie`, every core of that herd shall link with the object file.
> 
> https://github.com/Xilinx/mlir-air/blob/2754dcf53e9dce76df4e1a8d5f34bcb2acd1871e/mlir/lib/Conversion/AIRToAIEPass.cpp#L259

Yeah I implemented it so I'm familiar with how it works. My question was regarding the addition of the attribute in the tablegen description, and if it is necessary to make the python work.

hunhoffe: The python did not work without this change. I'm sure there may be other ways to signal that there needs to be a `link_with` attribute, but this is one path that seemed reasonable (to me, at least).

It did occur to me (and I did try this) that maybe I could just specify a type of `link_with` attribute when I declare the external function. I tried several modification to the `external_function` helper but was unable to make this work (I think I was able to set it as an attribute correctly, but this attribute then was not percolated to the herd/whoever was calling this function).

PR #659, opened 2024-07-10
Remove libxaie path

erwei-xilinx:


fifield:
Can we just remove this script in favor of using the wheel?

erwei-xilinx:
> Can we just remove this script in favor of using the wheel?

Sure that makes sense. And with the mlir-aie wheel do we still need to pass in extra libxaie build for air? I was hoping that if we can just use the aie-rt then we could remove one dependency.

PR #661, opened 2024-07-10
Channel Example: Hierarchical - Launch to Segment to Herd and back

hunhoffe:
This PR introduces one new examples using channels. The hierarchical example sends data from launch to segment to herd explicitly using channels.


programming_examples/channel_examples/hierarchical_alloc/hierarchical_alloc.py
```diff
@@ -0,0 +1,134 @@
...
+        def launch_body(a, b):
+            image_in_l3 = AllocOp(image_type_l3, [], [])
+
+            # Do some data manipulation instead of a straight copy
+            # 0, 1, 2, 3 -> 1, 1, 2, 2
+            c0 = arith.constant(T.index(), 0)
+            c1 = arith.constant(T.index(), 1)
+            c2 = arith.constant(T.index(), 2)
+            c3 = arith.constant(T.index(), 3)
+            val1 = load(a, [c0, c1])
```
erwei-xilinx, line 61: What is executing the loads and stores? Host?

programming_examples/channel_examples/hierarchical_alloc/hierarchical_alloc.py
```diff
@@ -0,0 +1,134 @@
...
+
+                ChannelGet("ChanInL2", image_in_l2a)
+
+                # Do some data manipulation instead of a straight copy
+                # 1, 1, 2, 2 -> 2, 1, 2, 2
+                c0 = arith.constant(T.index(), 0)
+                c1 = arith.constant(T.index(), 1)
+                c2 = arith.constant(T.index(), 2)
+                c3 = arith.constant(T.index(), 3)
+                val1 = load(image_in_l2a, [c0, c0])
```
erwei-xilinx, line 86: Similarly, program inside segment and outside of herd is going to be mapped to memtile DMA BDs. Memtile doesn't really have a core, so who will be doing the load or store?

programming_examples/channel_examples/hierarchical_alloc/hierarchical_alloc.py
hunhoffe: I thought that since load/store are memory operations, it was reasonable to think that maybe they would be mapped to a DMA operation?

The main goal here is just to somehow check that the allocations are actually effective in some way. Do you have any suggestions for how to do this?

hunhoffe:
I removed the alloc example for now; I plan to come up with a more realistic design at a later time

erwei-xilinx (approved):

programming_examples/channel_examples/hierarchical_alloc/hierarchical_alloc.py
erwei-xilinx: I see. Any load/store ops intended to map to DMAs need to be explicitly represented using air.channel / air.dma_memcpy_nd; load store ops are currently intepreted as being explicitly moving data using some core.

Issue #666, opened 2024-07-12
Access memory allocated in segment in the herd


hunhoffe:
I'm trying to make an example where I allocate it in the segment (in L2 memory) and then use it as a target for the `dma_memcpy_nd` in the herd within the segment. I have written an example for that in a branch called `alloc-check-example`.

In the current form of the code, where I try to send the allocated memory to the herd through an operand/argument, I get the following error:
```
  File "mlir-air/install-xrt/python/air/dialects/_air_ops_ext.py", line 105, in <listcomp>
    operand_types = [s.type for s in sizes] * 2 + [o.type for o in operands]
AttributeError: 'AllocOp' object has no attribute 'type'
make: *** [Makefile:9: run] Error 1
```

If I don't try to send it from the segment to the herd as a herd operand, I get this error instead:
```
air._mlir_libs._site_initialize.<locals>.MLIRError: Unable to parse module assembly:
error: "-":21:11: 'air.dma_memcpy_nd' op using value defined outside the region
 note: "-":21:11: see current operation: "air.dma_memcpy_nd"(<<UNKNOWN SSA VALUE>>, %arg4, %2, %3, %4, %5, %6, %7) <{operandSegmentSizes = array<i32: 0, 1, 0, 0, 0, 1, 2, 2, 2>}> : (memref<16x8xi32, 1 : i32>, memref<32x16xi32>, index, index, index, index, index, index) -> ()
 note: "-":11:9: required by region isolation constraints
make: *** [Makefile:9: run] Error 1
```

I believe one of these two things should work, but I'm not sure which one (or both).

erwei-xilinx:
In the first case, it looks to me that it was the `memref.alloc` op that got passed as argument to the herd op. I think if we pass the ssa memref value being returned from the alloc op, then it might work? Would be really useful if you could share the example.

In the second case, yeah this is expected because air herd/segment/launch ops have trait `isolatedFromAbove`.

hunhoffe:
The example is here: https://github.com/Xilinx/mlir-air/blob/alloc-check-example/programming_examples/segment_alloc/segment_alloc.py

I'm not sure how to get a handle to/represent the ssa memref value in Python, but I can try to look into it. Let me know if you have any ideas!

erwei-xilinx:
[...]

Issue #654, opened 2024-07-08
Worker-to-Worker Channel Example


hunhoffe:
I am working on writing a worker-to-worker data transfer example for channels (as part of the grouping of examples that exercise various features of channels, https://github.com/Xilinx/mlir-air/issues/648).

Draft PR is here: https://github.com/Xilinx/mlir-air/pull/653

I am basing it off the code in the channel_size example (PR waiting to be merged here: https://github.com/Xilinx/mlir-air/pull/642)

The channel_size example works well for me. As an intermediate step to adding worker-to-worker communication to that example, I tried to have each worker send data to itself over a channel. That is the version of the code that is pushed in the draft PR  https://github.com/Xilinx/mlir-air/pull/653. The particular file of interest is [this one](https://github.com/Xilinx/mlir-air/blob/daa2d52899aee2d8f6d739f44fab1181f8d29dec/programming_examples/channel_examples/worker_to_worker/worker_to_worker.py).

When I run with this intermediate step, I get the following error:
```
Using aiecc.py from:  /scratch/ehunhoff/mlir-air/mlir-aie/install/bin/..
Running: builtin.module(air-insert-launch-and-segment-around-herd,func.func(air-lower-herd-parallel),air-dma-to-channel,canonicalize,cse,air-specialize-channel-wrap-and-stride,func.func(air-renumber-dma),func.func(convert-linalg-to-loops),air-place-herds{num-rows=6 num-cols=4 row-anchor=2 col-anchor=0})
Running: builtin.module(air-to-aie{emit-while-loop=false row-offset=2 col-offset=0 device=npu1_4col})
python3: /scratch/ehunhoff/mlir-air/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp:956: void xilinx::air::simpleDMAChannelAllocation(std::vector<MemcpyBundleAsFlow> &, ShimDMAAllocator &, MemTileDMAAllocator &, TileDMAAllocator &): Assertion `core' failed.
Aborted (core dumped)
make: *** [Makefile:7: run] Error 134
```

My question is:
* Is a worker allowed to put/get data to/from a channel to itself?
* Or is this a bug (either in my example code or the air compiler)?

erwei-xilinx:
And this assumption could affect other methods and passes in air compiler.

hunhoffe:
Ah, I see!

As a path forward, would it make sense to:
* Create a channel-to-self example with two sub-examples:
[...]

PR #668, opened 2024-07-16
Access L2 Segment Allocation in Herd with Python Example

hunhoffe:
Previously, it was not possible to send an allocation made in a segment to a herd. Now it is possible to do so. I add a programming example that exercises this capability.

This addresses issue https://github.com/Xilinx/mlir-air/issues/666


python/air/dialects/_air_ops_ext.py
```diff
@@ -48,7 +71,7 @@ def __init__(
             launch_operands=operands,
             sym_name=name,
         )
-        operand_types = [s.type for s in sizes] * 2 + [o.type for o in operands]
+        operand_types = [s.type for s in sizes] * 2 + get_region_operand_types(operands)
```
hunhoffe, line 74: I think considering not just Values but also Operations makes sense in the segment and herd since one could feasibly create the operation in the outer region. However, the launch is the outermost region in which an operation could occur (to my knowledge) so I suspect that we should only allow Values here.

Does this reasoning make sense? if so, I'll remove the change to the `Launch()` class.

programming_examples/channel_examples/worker_to_self/worker_to_self.py
```diff
@@ -23,14 +23,18 @@ def build_module():
     ChannelOp("ChanOut")
```
hunhoffe, line 23: I attempted to change this file to use the new ability to send L2 allocations made in a segment to a herd. However, this example still does not work. I'm unsure of how I've rewritten it is reasonable or not, but I've made a note in the associated issue (https://github.com/Xilinx/mlir-air/issues/654)

erwei-xilinx: The `tensor_in_l2` seems to be deallocated before it was used in `air.herd`, if the code was interpreted line by line. Maybe move the deallocation to after the `herd`.

erwei-xilinx (approved):

programming_examples/channel_examples/worker_to_self/worker_to_self.py
hunhoffe: I fixed a bug (I accidentally was using L1 for both allocation types) and made the change you suggested, as well as simplifying the example a bit. The example is still not successful, although the error changed:
```
Running: builtin.module(air-to-aie{emit-while-loop=false row-offset=2 col-offset=0 device=npu1_4col})
Segmentation fault (core dumped)
```

python/air/dialects/_air_ops_ext.py
hunhoffe: @jgmelber Do you have an opinion?

programming_examples/channel_examples/worker_to_self/worker_to_self.py
```diff
@@ -47,51 +51,43 @@ def launch_body(a, b):
             @segment(name="seg")
             def segment_body():
 
+                tensor_in_l2 = AllocOp(image_type_l2, [], [])
+                ChannelGet("ChanIn", tensor_in_l2)
```
erwei-xilinx, line 55: After `tensor_in_l2` is filled with data from getting `ChanIn`, I would expect a put to some L2->L1 channel?

programming_examples/channel_examples/worker_to_self/worker_to_self.py
```diff
@@ -47,51 +51,43 @@ def launch_body(a, b):
...
                     # We must allocate a buffer of image size for the input/output
-                    tensor_in = AllocOp(image_type, [], [])
-                    tensor_out = AllocOp(image_type, [], [])
-                    tensor_in2 = AllocOp(image_type, [], [])
-                    tensor_out2 = AllocOp(image_type, [], [])
+                    tensor_in_l1 = AllocOp(image_type_l1, [], [])
+                    tensor_out_l1 = AllocOp(image_type_l1, [], [])
 
-                    ChannelGet("ChanIn", tensor_in)
+                    ChannelPut("ToSelf", tensor_in_l2)
```
erwei-xilinx, line 70: I see. I'd sugges moving this channel put to outside herd.

python/air/dialects/_air_ops_ext.py
erwei-xilinx: I think in the future there could be other operations before or after launch.

programming_examples/channel_examples/worker_to_self/worker_to_self.py
hunhoffe: This example is called worker to self because the purpose of the example is to demonstrate (if) it is possible to put data into a channel and then fetch it from the same channel from within the same worker/compute core in a herd. So to keep this example, we need both `ChannelPut("ToSelf", some_mem)` and `ChannelGet("ToSelf", some_other_mem)` within the herd. 

The only reason I modified this example in this PR is because previously this example was blocked because I was unable to access L2 memory in the herd, and that capability is needed because channels must move data between different memory regions (in this case, L2 and L1).

python/air/dialects/_air_ops_ext.py
hunhoffe: Okay, I will keep my changes then!

programming_examples/channel_examples/worker_to_self/worker_to_self.py
hunhoffe: I have to do L2->L1 in the herd due to the purpose of this example (see my other comment).

If this not a valid operation?

programming_examples/channel_examples/worker_to_self/worker_to_self.py
erwei-xilinx: I see. This scenario is not exercised before.

programming_examples/channel_examples/worker_to_self/worker_to_self.py
erwei-xilinx: It is valid operation. I didn't understand the intend the design but I get it now.

python/air/dialects/_air_ops_ext.py
fifield: Maybe I'm missing some subtlety about the python bindings but isn't the result of an Operation always one or more Values? Can you share a code snippet where this comes up? This isn't an issue in ordinary MLIR code or using C++ APIs.

erwei-xilinx: Here is the context: https://github.com/Xilinx/mlir-air/issues/666

Issue #627, opened 2024-06-27
Multiple launches, herd with single core


hunhoffe:
I'm working on an example where I have an 2D matrix of input data, then I break it into four data tiles, and then I am attempt to process one data tile per one compute tile in a variety of ways using AIR constructs. I am sanity checking my programs by making each compute core add a unique `tile_num` to each value in the data tile they modify, so I can reassure myself that the compute tile I think is doing some work is actually the compute tile doing the work.

Anyways, I am trying to compose an example of this scenario that uses four launches, where the herd size is 1x1. My first attempt is [here](https://github.com/Xilinx/mlir-air/blob/minimal-matrix-scalar-add/programming_examples/matrix_scalar_add/multi_launch_channel/multi_launch_channel.py) where I have a while loop within the herd because I hear the kernel will be persistent across launches.

Anyways, even with that persistence, I'd like to somehow parameterize the herd with the launch indices so I can calculate a unique `tile_num` per launch. Is this something that is possible to do? If not, how do I reassure myself that one data tile is being processed per launch?

erwei-xilinx:
We do not have an example in AIR today to parametrise herd using launch induction variables, or any runtime scalar parameter passed in from the host. For now, we only have launch induction variables parametrising the SHIM DMA BDs for streaming the correct data into the herd.

Would be really cool to have an example in AIR which can lower any runtime parameters into herd as MLIR-AIE RTP.

hunhoffe:
I'm trying my hand at this to see if I can figure it out.

My example is here: https://github.com/Xilinx/mlir-air/blob/1bbc92a4b1dcaa66cbab50dc494af53bec1d8472/programming_examples/matrix_scalar_add/multi_launch_channel/multi_launch_channel.py

The initial AIR MLIR looks somewhat reasonable (to my novice eye):
```mlir
#map = affine_map<()[s0] -> (s0 * 16)>
#map1 = affine_map<()[s0, s1] -> (s0 + s1)>
module {
  air.channel @ChanIn []
  air.channel @ChanOut []
  func.func @copy(%arg0: memref<32x16xi32>, %arg1: memref<32x16xi32>) {
    %c2 = arith.constant 2 : index
    %c2_0 = arith.constant 2 : index
    air.launch (%arg2, %arg3) in (%arg4=%c2, %arg5=%c2_0) args(%arg6=%arg0, %arg7=%arg1) : memref<32x16xi32>, memref<32x16xi32> {
[...]

PR #672, opened 2024-07-18
Fixup data layout in programming examples

erwei-xilinx:
This PR makes changes to the `matrix_scalar_add` programming examples to fixup the following issues:
- The original example's `common.py` testbench expects data layout in DDR as [IMAGE_WIDTH, IMAGE_HEIGHT], but many L2 memrefs have shape [TILE_HEIGHT, TILE_WIDTH].
- The rest of the examples viewed accesses to those memrefs as single pointers. While correctness is maintained in most tests, there were many cases where data access pattern does not align with the n-d memref shape.

### DIFF
--- a/programming_examples/matrix_scalar_add/common.py
+++ b/programming_examples/matrix_scalar_add/common.py
@@ -5,13 +5,13 @@ import numpy as np
 import air.backend.xrt as xrt_backend
 import filelock
 
-IMAGE_WIDTH = 32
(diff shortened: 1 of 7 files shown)


hunhoffe (approved):
Thank you for the changes! If the multi-launch works, we should update/close this issue here: https://github.com/Xilinx/mlir-air/issues/627

PR #677, opened 2024-07-23
AIRSegmentLoopFusion: Iterative loop fusion to exhaust all optimization opportunities

erwei-xilinx:


mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
```diff
@@ -4920,9 +4922,10 @@ class AIRSegmentLoopFusion
     runPreProcPatterns(func);
     SmallVector<air::SegmentOp> segs;
     func.walk([&](air::SegmentOp op) { segs.push_back(op); });
-    for (auto seg : segs) {
-      runOnSegment(seg);
-    }
+    for (auto seg : segs)
+      while (runOnSegment(seg).succeeded()) {
+        continue;
+      }
```
fifield, line 4930: This is sort of what `applyPatternsAndFoldGreedily` is for.  Maybe `runOnSegment` can be a rewrite pattern(s)? I'm not sure what it does.

erwei-xilinx: Yeah the goal of this runOnSegment is to fuse multiple bands of perfectly nested for loops in a segment into one with shared parent for loop. I originally tried writing as a pattern for matching `scf.for` but that didn't work since we are fusing multiple into one. Maybe we can rewrite that into a pattern matching `air.segment`... Let me try something.

erwei-xilinx: `runOnSegment` is now reimplemented as rewrite pattern.

PR #678, opened 2024-07-23
Special handling of allocs and stores for reshape ops

akkothar:
Handling scenarios when reshape ops are lingering around in the code to reshape tensors/matrices. memref.reshape ops require shapes to be specified using memref.allocs and memref.store operations. The changes to the AIRDependency pass accounts for these scenarios so as to avoid  creating async execute ops for memref.allocs and memref.store ops associated with memref.reshapes ops.

### DIFF
--- a/mlir/test/Transform/AIRDependency/reshape_matmul_nd.mlir
+++ b/mlir/test/Transform/AIRDependency/reshape_matmul_nd.mlir
@@ -1,0 +1,106 @@
+//===- matmul_nd.mlir ------------------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2022, Xilinx Inc. All rights reserved.
+// Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: air-opt %s -air-dependency | FileCheck %s
+
+module attributes {torch.debug_module_name = "mmult"}  {
+  func.func @forward(%arg0: memref<64x64xi32>, %arg1: memref<64x16x4xi32>, %arg2: memref<?x?xi32>) {
+    %c0_i32 = arith.constant 0 : i32
+    %c2 = arith.constant 2 : index
+    %c0 = arith.constant 0 : index
+    %c4 = arith.constant 4 : index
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Could you please add some ci tests?

mlir/test/Transform/AIRDependency/reshape_matmul_nd.mlir
```diff
@@ -0,0 +1,106 @@
...
+    %c16 = arith.constant 16 : index
+    %c32 = arith.constant 32 : index
+    %c64 = arith.constant 64 : index
+    %c1 = arith.constant 1 : index
+    %0 = memref.alloc() : memref<64x64xi32>
+    // CHECK: = air.execute
+    // CHECK: air.execute_terminator
+    linalg.fill ins(%c0_i32 : i32) outs(%0 : memref<64x64xi32>)
+    // CHECK: = air.execute
+    %1 = memref.cast %arg2 : memref<?x?xi32> to memref<64x64xi32>
```
erwei-xilinx, line 26: Is this `memref.cast` the only code pattern that you want to test for? If so, can we reduce the test size to focus on this op?

akkothar: Actually, let me just remove memref.cast. We are testing for memref.alloc and memref.stores associated with the memref.reshape op.

mlir/test/Transform/AIRDependency/reshape_matmul_nd.mlir
```diff
@@ -0,0 +1,106 @@
...
+      %c8 = arith.constant 8 : index
+      %c32_0 = arith.constant 32 : index
+      %c0_0 = arith.constant 0 : index
+      %c4_0 = arith.constant 4 : index
+      %c64_1 = arith.constant 64 : index
+      %c1_2 = arith.constant 1 : index
+      %shape = memref.alloc() : memref<2xindex>
+      // CHECK-NOT: = air.execute
+      // CHECK-NOT: air.execute_terminator
+      memref.store %c32_0, %shape[%c0_0] : memref<2xindex>
```
erwei-xilinx, line 57: And same above, would it be possible to reduce the test to focus on testing this `memref.store`?

PR #690, opened 2024-07-30
Herd Size Default Cleanup

hunhoffe:
Cleanup function args for herd sizes argument.


hunhoffe:
This was just an exploration into herd size. I'm shelving it for now.

PR #693, opened 2024-08-01
Fix tests again

fifield:
try again on https://github.com/Xilinx/mlir-air/pull/691


eddierichter-amd (approved):
These tests work for me now. Thanks!

PR #698, opened 2024-08-05
Add support for correctly handling reshape/expand/collapse shape ops

akkothar:
Ops such as reshape, expand and collapse shape do not execute, and these ops affect the dependence graph. This pull request attempts to handle these ops correctly and make sure the dependence graph is generated correctly when these ops are lingering around in the code.

### DIFF
--- a/mlir/lib/Transform/AIRDependency.cpp
+++ b/mlir/lib/Transform/AIRDependency.cpp
@@ -221,6 +221,12 @@ public:
                 isCandidateExecute = false;
             }
           }
+          // No air execute for expand, collapse and reshape ops
+          if (dyn_cast<memref::ReshapeOp>(op) ||
+              dyn_cast<memref::ExpandShapeOp >(op) ||
+              dyn_cast<memref::CollapseShapeOp >(op)) {
+            isCandidateExecute = false;
+          }
           // No air execute for loop ops
           if (isa<mlir::LoopLikeOpInterface>(op))
             isCandidateExecute = false;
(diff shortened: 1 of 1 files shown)


mlir/lib/Util/Dependency.cpp
```diff
@@ -1926,9 +1934,18 @@ void dependencyCanonicalizer::redoDepTraceIfDepOnHier(func::FuncOp func) {
     SmallVector<Value, 1> sink_op_scalar_ins;
     SmallVector<Value, 1> sink_op_scalar_outs;
     // auto &bb = exec_op.getBody().front();
-    Operation &child_op = exec_op.getBody().front().getOperations().front();
+    // Operation &child_op = exec_op.getBody().front().getOperations().front();
```
erwei-xilinx, line 1937: Nit: Commented code.

mlir/lib/Util/Dependency.cpp
```diff
@@ -1926,9 +1934,18 @@ void dependencyCanonicalizer::redoDepTraceIfDepOnHier(func::FuncOp func) {
...
     // Operation &child_op = exec_op.getOps().begin();
-    depTracer.getPartialMemrefFromOp(&child_op, sink_op_memref_reads,
+    Operation *child_op = nullptr;
+    // Pick the first op that is not a shape altering op.
+    for (auto &op : exec_op.getBody().front().getOperations()) {
```
erwei-xilinx, line 1941: Nit: air::ExecuteOp now has a method getChildOp() which will get the child op for you.

erwei-xilinx (approved):

mlir/lib/Util/Dependency.cpp
akkothar: Yes, but this is looking for the first child that is not a shape altering op, and not just the first child op. So I think the loop and checks are still necessary.

erwei-xilinx: Ok. I think it's good to merge once CI passes.

PR #700, opened 2024-08-07
Support changing the function calls replacing linalg ops in presence of reshape ops 

akkothar:
This pull request fixes signature of calls to library functions in presence of reshape/expand/collapse ops. This is important because no other way of lowering support such ops currently exists, and this is a way of getting rid of them before generating  code for AIEs.

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -3474,7 +3474,6 @@ public:
         op, fnNameAttr.getValue(), TypeRange(),
         ValueRange(ArrayRef<Value>(libFnOperands)));
 
-    module.dump();
     return success();
   }
 
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -3425,13 +3425,34 @@ public:
     if (fnName.empty())
       return failure();
 
+    // Function to get operands of the library call that will
+    // replace the given linalg op.
+    auto getLibFnOperands = [](linalg::LinalgOp op) {
+        SmallVector<Value> operands;
+        for (auto operand : op->getOperands()) {
+            if (auto operation = operand.getDefiningOp()) {
+                if (dyn_cast<memref::ReshapeOp>(operation) || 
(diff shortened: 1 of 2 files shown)


mlir/test/Conversion/AIRToAIE/air_linalg_to_func.mlir
fifield, line 1: Can this test be made smaller? The code can be nonsense, but should be the minimal code to test the transformation.

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -3425,13 +3424,34 @@ class AIRLinalgOpToLibraryCallRewrite
...
 
+    // Function to get operands of the library call that will
+    // replace the given linalg op.
+    auto getLibFnOperands = [](linalg::LinalgOp op) {
+      SmallVector<Value> operands;
+      for (auto operand : op->getOperands()) {
+        if (auto operation = operand.getDefiningOp()) {
+          if (dyn_cast<memref::ReshapeOp>(operation) ||
+              dyn_cast<memref::ExpandShapeOp>(operation) ||
+              dyn_cast<memref::CollapseShapeOp>(operation)) {
```
fifield, line 3435: Could use `isa<X...>(operation)>` here or `isa_and_present` to combine with line 3432:
```C++
Operation *operation = operand.getDefiningOp();
if (!isa_and_present<memref::ReshapeOp, memref::ExpandShapeOp, memref::CollapseShapeOp>(operation)>)
  continue;
operands.push_back(operation->getOperand(0));
```

erwei-xilinx (approved):

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -3511,7 +3531,7 @@ FailureOr<ModuleOp> convertAIRToAIE(mlir::RewriterBase &rewriter,
                                        /* .device = */ *device};
   std::vector<std::pair<ModuleOp, xilinx::air::HerdOp>> aie_modules;
   p.walk([&](xilinx::air::HerdOp h) {
-    aie_modules.push_back({aie_module, h});
```
erwei-xilinx, line 3514: Clang format error points to this. Change this back to the original format

PR #701, opened 2024-08-07
bump mlir-aie

fifield:


erwei-xilinx:
@fifield This PR should make the test to pass: https://github.com/Xilinx/mlir-air/pull/703

PR #703, opened 2024-08-09
Fixup an issue in the input IR

erwei-xilinx:
... where the amount of data moved between src and dst of `dma_memcpy_nd` mismatches.

DDR side memref size is `vector_size` which is 4k. L1 side memref size is `lineWidthInBytes` which is 1k.


erwei-xilinx:
Cc: @hunhoffe

fifield:
Thanks, merged as part of #701.

PR #704, opened 2024-08-12
Channel Example: Sliding Window

hunhoffe:
This is an example of accessing sliding windows of data with channel operations. This example is rudimentary and not ideal because all data must be read in at once. A better example would allow more of a data streaming mechanism, but I'm not sure how best to do that with the current channel API. I think to do this properly you would need some sort of `peek()` mechanism.

If I don't have the `peek()` mechanism, if the sliding window is in an `scf.for`, I believe the only options are to put nested `if/else` (which I haven't yet ever used, but should be in `mlir-python-extras`) statements or an `scf.index_switch` (which is not in `mlir-python-extras`) with one statement per unique window in the loop... which is not a scalable way to write this sort of code.

I made this example along the way towards working on a large example, and I thought it was a useful building block, so that is why I am adding it to the programming examples (even with the sub-optimal implementation). I'm very open to suggestions for how to make it better!

### DIFF
--- a/programming_examples/channel_examples/README.md
+++ b/programming_examples/channel_examples/README.md
@@ -34,6 +34,12 @@ This example ([worker_to_worker/worker_to_worker.py](worker_to_worker/worker_to_
 
 WARNING: This example currently fails for unknown reasons.
 
+#### WIP: Sliding Window of Data
+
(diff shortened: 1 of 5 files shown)


hunhoffe:
Closing because I do not intend to finish this.

PR #711, opened 2024-08-14
Fix linalg/memref copy to dma memcpy implementation when no source or destination ancestors are found

akkothar:
This accounts for scenarios when there is no source or destination ancestor to a linalg/memref copy op is found. In these situations, it is appropriate to simply copy over the original source/destination strides, sizes and offsets.


erwei-xilinx (approved):
Thanks for adding the missing else condition. LGTM!

PR #712, opened 2024-08-14
Packet-flow routing at SHIM DMA MM2S channels, for control packets

erwei-xilinx:
This PR adds to `air-to-aie` the option to `insert-control-packet-flow`, which performs the following actions:
- It generates `aie.packet_flow` instead of `aie.flow` ops for any shim dma mm2s streaming connections. Packet-flow routing allows for sharing of shim dma channels as fanout.
- For each column used in the design, it spawns a network of packet-flow routings between the column's shim dma mm2s channels, and the CTRL port of all tiles in that column.

~~TODO: board test added in this PR shall fail for now, until mlir-aie's pathfinder for packet-flow allocation is updated.~~
~~TODO: the new board test shall pass once this PR lands in mlir-aie: https://github.com/Xilinx/mlir-aie/pull/1694~~
~~TODO: bump mlir-aie once pathfinder update has landed.~~

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -1752,6 +1752,9 @@ public:
     return pktFlow;
   }
 
+  // This method generates broadcast packet flow if found multiple flows with
+  // the same source. TODO: packet flows sharing source do not always mean
+  // broadcast.
   AIE::PacketFlowOp getPacketFlowOp(AIE::DeviceOp aie_device,
                                     mlir::Value source,
                                     xilinx::AIE::WireBundle sourceBundle,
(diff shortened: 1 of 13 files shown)


erwei-xilinx:
Cc: @Yu-Zhewen

mlir/lib/Conversion/AIRToAIEPass.cpp
```diff
@@ -3061,16 +3234,65 @@ class AIRToAIEPass : public air::impl::AIRToAIEBase<AIRToAIEPass> {
...
+        (void)createPacketFlowOp(
+            builder, flowID, srcTile, AIE::WireBundle::Trace, 0, destTile,
+            AIE::WireBundle::DMA, destChan, keep_pkt_header);
+      }
+    }
+  }
+
+  // Create an overlay of packet flow network sending control packets from shim
+  // dma mm2s channels to tiles in a column.
+  void createControlPacketFlow(AIE::DeviceOp device) {
```
fifield, line 3246: As far as I can tell this is a mlir-aie transformation. Can we put it there instead? If it was generalized a bit then it could already be used to fix [this](https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetCDODirect.cpp#L611-L617) which is just a very small control overlay. It could be inserted with a pass instead. And I can think of several other uses.

erwei-xilinx: Yeah that makes great sense to me. It could totally become an mlir-aie pass, because it was only looking at `aie.tile` ops in the IR.

This overlay being generated comes with some constraints though. One thing that I haven't figure out is, if the mlir-aie design already has circuit-switch routed shim dma mm2s channel usage, then we certainly cannot use those channels for control packets. In AIR that was easily controlled, because we generate the whole design, and we can force the design to never circuit-switch reserve any shim s2mm channels to begin with. Whereas if this one becomes an mlir-aie pass, then I haven't figured out how to make this overlay flexible to an existing aie design, while still enfocing full connectivitly to every tile's control port...

Maybe we can discuss this in greater details in our meeting on Friday?

PR #713, opened 2024-08-15
Correctly generate offsets for DMA memcpys for AIEs

akkothar:
This pull request has the following modifications:

- Generating correct offsets in situations where layouts are in some blocked format.
- Correctly splitting DMA memcpys: an extra add instruction was being generated when correcting offsets after splitting memcpys.


erwei-xilinx:
@akkothar, would it be possible if you could add your blocked GEMM e2e example as a board test in mlir-air, under test/xrt? So that in our future development we do not break the design.

akkothar:
> @akkothar, would it be possible if you could add your blocked GEMM e2e example as a board test in mlir-air, under test/xrt? So that in our future development we do not break the design.

Yeah, I am about to create separate PRs for some tests so you can decide what tests you want to add because I have multiple tests on my end.

erwei-xilinx (approved):
Thanks for writing this fix. This is really neat.

PR #716, opened 2024-08-16
Add a matmul board test with weight matrix in blocked layout; transpose at memtile ->core level

akkothar:


erwei-xilinx (approved):

erwei-xilinx:
To be modified and landed in https://github.com/Xilinx/mlir-air/pull/721

PR #717, opened 2024-08-16
Xfail timeout

fifield:
XFAIL xrt tests 04 and 12 because they take too long and trigger a timeout on recent drivers.


erwei-xilinx (approved):

fifield:
> XFAIL xrt tests 04 and 12 because they take too long and trigger a timeout on recent drivers.

Maybe it's not a timeout. When I run test 04 manually it seems to pass about 20% of the time and it takes about 1minute to run. The rest of the time it fails and takes less than 3 seconds to run.

erwei-xilinx:
Hopefully this PR can make test 04 work with the recent drivers again: https://github.com/Xilinx/mlir-air/pull/719

erwei-xilinx:
This PR aims to re-enable test 12: https://github.com/Xilinx/mlir-air/pull/829

PR #718, opened 2024-08-16
bump mlir-aie and llvm

fifield:


erwei-xilinx (approved):
Thanks for creating this PR. My PR https://github.com/Xilinx/mlir-air/pull/712, which is hitting the llvm versioning issue, shall land after this one.

Issue #707, opened 2024-08-13
`air.channel.put/get` ops are not able to lower through `air-to-aie` if memref is not produced directly by `memref.alloc`


erwei-xilinx:
Unlike `air.dma_memcpy_nd` which has `air-copy-to-dma` folding its src/dst side memref chain---consisting of `memref.subview`, `memref.transpose`, or `memref.expand_shape`---into its wrap-and-stride list, `air.channel.put/get`, if having src/dst side memref produced by any of those memref ops, will fail at `air-to-aie`. This is because `air-to-aie` assumes that any memref op chain at src/dst were already folded to wraps and strides prior to this point.

This won't affect code progressively compiled from `memref.copy` via `air.dma_memcpy_nd`; it shall affect user handwritten code.

erwei-xilinx:
Currently being handled with a slightly more informative failure: https://github.com/Xilinx/mlir-air/pull/706

Might be possible to make this work with some canonicalizer to `air.channel.put/get` to handle this scenario.

erwei-xilinx:
Update on this issue: this could be fixed by rewriting the `condenseMemrefDataReorderingToAIRDma` as a canonicalizer. Related PR: https://github.com/Xilinx/mlir-air/pull/744 Thanks, @fifield!

erwei-xilinx:
WiP: https://github.com/Xilinx/mlir-air/pull/749

PR #744, opened 2024-10-18
CopyToDma: Fold CastOp and ExpandShapeOp into air::DmaMemcpyOp

erwei-xilinx:


fifield (commented):
It seems like `condenseMemrefDataReorderingToAIRDma` is a canonicalization pattern? Should it be implemented as one so that it runs whenever the canonicalilzer runs and not just during this pass? It might also simplify the code to consider each case as a simple canonicalization, instead of as nested if's trying to do the whole thing at once.

erwei-xilinx:
> It seems like `condenseMemrefDataReorderingToAIRDma` is a canonicalization pattern? Should it be implemented as one so that it runs whenever the canonicalilzer runs and not just during this pass? It might also simplify the code to consider each case as a simple canonicalization pattern, instead of as nested if's trying to do the whole thing at once.

Totally agreed. The lack of such canonicalizer has been creating issues for `air.dma_memcpy_nd` and `air.channel.put/get`, as subsequent passes can no longer handle those `memref` ops. I updated the issue https://github.com/Xilinx/mlir-air/issues/707 to keep track of this.

PR #745, opened 2024-10-18
AIRSplitL2Memref: Disables L2 memref splitting for small designs

erwei-xilinx:
- Disables the AIRSplitL2Memref pass if none of the `air.segments` need to map to more than one memtiles.
- A new option `clNumTilesPerL2Tile` is added to pass in architecture information: the number of compute tiles with close affinity to one memtile.
- For example, a column of 4 tiles have close affinity to 1 memtile on NPU.


fifield (commented):
The existing docs say:
```
Split L2 memref into smaller buffers if it couldn’t fit with the buffer harware constraints

Transforms the input IR by splitting certain L2 memory references (memrefs) to adhere to AIE memtile-specific buffer and DMA channel constraints or optimization opportunities.
```
and the new option says:
```
"Number of compute tiles per L2 memory tile. Used to estimate if an air.segment shall allocate to multiple L2 memory tiles, and therefore requires L2 memref splitting."
```
The existing documentation is pretty vague about when splitting occurs. I can't parse it. The first sentence makes sense, but the rest sounds like a politician's answer. Can you define: "certain L2 memrefs", "memtile-specific buffer and DMA channel constraints", and "optimization opportunities", and how "compute tiles per L2 memory tile" is "Used to estimate" in the context of those other things?

erwei-xilinx:
> The existing docs say:

Thanks for the comment. Agreed that the previous doc was vague and some parts of it are obsolete. I hope that the updated doc better describes the pass.

fifield:
> > The existing docs say:
> 
> Thanks for the comment. Agreed that the previous doc was vague and some parts of it are obsolete. I hope that the updated doc better describes the pass.

It is much better, thanks!

PR #750, opened 2024-10-23
Safer `Operation` erase using `llvm::SmallSet`

erwei-xilinx:


fifield:
Why can't this op be handled by a conversion pattern like the other ops?

erwei-xilinx:
> Why can't this op be handled by a conversion pattern like the other ops?

This PR changes the way that `Operations` targetted for deletion were collected in passes, where previously they were buffered in a `SmallVector` before being deleted, posing the risk of deleting the same op multiple times. This PR changes to buffer them using `SmallSet` instead.

Hmm this PR isn't targetting any specific op conversion or transformation, so which op are we referring to?

PR #752, opened 2024-10-24
AIRSegmentLoopFusion: Fixups on `affine::DelinearizeIndexOp` and rank reduction

erwei-xilinx:


mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
```diff
@@ -46,6 +46,8 @@
 #include <string>
 #include <vector>
 
+#include <iostream>
+
```
fifield, line 50: remove

erwei-xilinx: Oops. PR updated. Thanks for checking!

PR #768, opened 2024-11-08
Generalize canonicalization of async op dependency operands

fifield:
Generalize the async dependency operand canonicalization of wait_all and hierarchical ops so that it applies to other async interface ops. Also adds a new canonicalization to remove async token dependencies which are dependencies of other dependencies.


fifield:
Probably needs some additional tests too...

erwei-xilinx (approved):

PR #772, opened 2024-11-12
AIRSpecializeChannelWrapAndStride: Hoist loop-invariant pure ops for dominance

erwei-xilinx:
Below shows the standard case that the pass has been supporting, where the `affine.apply` can get folded into the `scf.for` loop bounds:
```
    scf.for %arg4 = %c0 to %c8 step %c1 {
      %0 = affine.apply #map()[%arg4]
      %1 = air.channel.put async @channel_21[] (%arg0[%0] [%c8] [%c1]) : (memref<128x256xi32>)
    }
```
However, user-provided code could come in the following form too:
```
    scf.for %arg4 = %c0 to %c8 step %c1 {
      %0 = affine.apply #map()[%arg1]
      %1 = air.channel.put async @channel_21[] (%arg0[%0] [%c8] [%c1]) : (memref<128x256xi32>)
    }
```
Where the `affine.apply` cannot fold into the loop due to `%arg1` not being the loop induction variable. In such case, the pass should hoist the `affine.apply` out of the for loop in order to fix value domination, since `affine.apply` is a pure op that doesn't touch memory and can therefore hoist safely.

### DIFF
--- a/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
+++ b/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
@@ -2001,17 +2001,41 @@ struct AIRSpecializeChannelWrapAndStrideInScfFor
     }
     SmallVector<Value, 1> deps =
         for_op.getOperands().drop_front(for_op.getNumControlOperands());
+
+    // Hoist any pure ops that the new channel op depends on.
+    auto new_opers = llvm::concat<Value>(
+        SmallVector<Value>{channel_ops[0].getMemref()},
+        channel_ops[0].getIndices(), offsets, wraps, strides);
+    IRMapping remap;
+    auto clonedOps = cloneDefiningOpsInRegion(
+        rewriter, &for_op.getRegion(), llvm::to_vector(new_opers), remap);
+    for (auto cloned : clonedOps) {
+      clearAsyncDependenciesOfAsyncOp(cloned);
+      for (auto token : deps)
+        addAsyncDependencyIfNew(cloned, token);
+      if (auto token = getAsyncTokenFromOp(cloned))
+        deps.push_back(token);
+    }
+
+    // Create specialized air.channel.put/get.
     if (isa<air::ChannelPutOp>(channel_ops[0]))
       new_chan_op = rewriter.create<air::ChannelPutOp>(
           loc, tys, deps, channel_ops[0].getChanName(),
-          channel_ops[0].getIndices(), channel_ops[0].getMemref(), offsets,
-          wraps, strides);
+          air::lookupOrDefaultRange(channel_ops[0].getIndices(), remap),
+          air::lookupOrDefaultRange(channel_ops[0].getMemref(), remap),
+          air::lookupOrDefaultRange(offsets, remap),
+          air::lookupOrDefaultRange(wraps, remap),
+          air::lookupOrDefaultRange(strides, remap));
     else if (isa<air::ChannelGetOp>(channel_ops[0]))
       new_chan_op = rewriter.create<air::ChannelGetOp>(
           loc, tys, deps, channel_ops[0].getChanName(),
-          channel_ops[0].getIndices(), channel_ops[0].getMemref(), offsets,
-          wraps, strides);
+          air::lookupOrDefaultRange(channel_ops[0].getIndices(), remap),
+          air::lookupOrDefaultRange(channel_ops[0].getMemref(), remap),
+          air::lookupOrDefaultRange(offsets, remap),
+          air::lookupOrDefaultRange(wraps, remap),
+          air::lookupOrDefaultRange(strides, remap));
 
+    // Clear all external uses of for_op before erasing it.
     for (auto res : for_op.getResults()) {
       if (isa<air::AsyncTokenType>(res.getType())) {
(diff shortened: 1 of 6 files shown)


mlir/lib/Util/Util.cpp
```diff
@@ -1577,8 +1577,27 @@ air::evaluateConstantsInMap(AffineMap map,
...
-  // auto newmap = map.replace(getAffineSymbolExpr(0, ctx), c, 0, 1);
   output = simplifyAffineMap(newmap).getSingleConstantResult();
   return output;
 }
+
+// Extend the lookupOrDefault method to operate on a vector of values.
+Value air::lookupOrDefaultRange(Value v, IRMapping &remap) {
+  return remap.lookupOrDefault(v);
+}
+SmallVector<Value> air::lookupOrDefaultRange(SmallVector<Value> vec,
```
fifield, line 1588:
```suggestion
SmallVector<Value> air::lookupOrDefaultRange(SmallVector<Value> &vec,
```

mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
```diff
@@ -2087,13 +2137,13 @@ struct AIRUnrollAffineForIntoBDChain
...
-        else if (auto exec = dyn_cast<air::ExecuteOp>(o)) {
-          auto childOp = exec.getChildOp();
-          if (childOp && isPure(childOp))
-            continue;
-        }
+        // else if (auto exec = dyn_cast<air::ExecuteOp>(o)) {
+        //   auto childOp = exec.getChildOp();
+        //   if (childOp && isPure(childOp))
+        //     continue;
+        // }
```
fifield, line 2146:
```suggestion
```

mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
```diff
@@ -2024,6 +2048,37 @@ struct AIRSpecializeChannelWrapAndStrideInScfFor
   }
 
 private:
+  // Clone backward slices of a list of values, opers.
+  SmallVector<Operation *> cloneDefiningOpsInRegion(OpBuilder builder,
+                                                    Region *region,
+                                                    SmallVector<Value> opers,
```
fifield, line 2054:
```suggestion
                                                    SmallVector<Value> &opers,
```

mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
```diff
@@ -2024,6 +2048,37 @@ struct AIRSpecializeChannelWrapAndStrideInScfFor
...
+      if (!region->isAncestor(operandDefOp->getParentRegion()))
+        continue;
+      assert(air::isPure(
+          operandDefOp)); // Pure ops are safe to hoist out of loops.
+      // Get backward slices
+      SetVector<Operation *> operandBS;
+      getBackwardSlice(operandDefOp, &operandBS, bsOptions);
+      for (auto b : operandBS)
+        if (air::isPure(b))
+          backwardSlices.insert(b);
```
fifield, line 2075: We assert that the operands are pure, but not the operands' operands?

mlir/lib/Util/Util.cpp
```diff
@@ -1577,8 +1577,27 @@ air::evaluateConstantsInMap(AffineMap map,
...
+    output.push_back(remap.lookupOrDefault(v));
+  }
+  return output;
+}
+
+// Extend isPure method to operate on air.execute.
+bool air::isPure(Operation *op) {
+  bool result = mlir::isPure(op);
+  if (auto execOp = dyn_cast<air::ExecuteOp>(op))
+    result = mlir::isPure(execOp.getChildOp());
```
fifield, line 1601: Is air.execute always assumed to have only one op inside? This is not checked by the verifier and we should relax it to not be the case.

mlir/lib/Util/Util.cpp
erwei-xilinx: Make sense. Updated.

mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
erwei-xilinx: Oops, deleted.

mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
erwei-xilinx: Updated.

mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
erwei-xilinx: Assertion added.

mlir/lib/Util/Util.cpp
erwei-xilinx: Originally, yes. But I agree with you that we should relax this constraint. Let me create a separate PR for it.

erwei-xilinx: PR created: https://github.com/Xilinx/mlir-air/pull/773

erwei-xilinx: PR created: https://github.com/Xilinx/mlir-air/pull/773

PR #773, opened 2024-11-13
Refactor `air.execute`'s `getBody()` and `getChildOp()` methods

erwei-xilinx:
- Rename the single region in `air.execute` to `region` instead of `body`.
- Add `getBody` method to get the op's single block.
- Replace `getChildOp` with `getChildOps` which returns a list of all child ops within each `air.execute` body.

### DIFF
--- a/mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
+++ b/mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
@@ -1026,9 +1026,12 @@ uint64_t HerdOp::getNumRows() {
 //
 
 LogicalResult ExecuteOp::verify() {
-  assert(getOperation()->getNumRegions() == 1 && "ExecuteOp has zero region!");
-  assert(!getRegion().empty() && "ExecuteOp should have non-empty region");
-  assert(!getBody().empty() && "ExecuteOp should have non-empty body");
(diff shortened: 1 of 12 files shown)


mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
```diff
@@ -1027,6 +1027,7 @@ uint64_t HerdOp::getNumRows() {
 
 LogicalResult ExecuteOp::verify() {
   assert(getOperation()->getNumRegions() == 1 && "ExecuteOp has zero region!");
+  assert(!getRegion().empty() && "ExecuteOp should have non-empty region");
   assert(!getBody().empty() && "ExecuteOp should have non-empty body");
```
fifield, line 1031: These should `return 
```suggestion
  if (getOperation()->getNumRegions() != 1)
    return emitOpError("ExecuteOp has zero region!");
  if (getRegion().empty())
    return emitOpError("ExecuteOp should have non-empty region");
  if (getBody().empty())
    return emitOpError("ExecuteOp should have non-empty body");
```

erwei-xilinx: Make sense. Updated.

PR #774, opened 2024-11-13
update mlir-aie + some updates for aie2p

fifield:
This PR updates the use of mlir-aie `AIETargetModel` after https://github.com/Xilinx/mlir-aie/pull/1911. Also starts to add in some AIE2p (strix) support, although that support is not tested.


erwei-xilinx (approved):
Thanks for working on this. LGTM!

PR #781, opened 2024-11-17
AIRToAIE: Switch to `isNPU` method for NPU-specific shim dma allocations

erwei-xilinx:
Currently, `AIRToAIE` pass lowers to two forms of shim dma allocation metadata: (1) `airrt.herd_metadata {dma_allocations: ... }` form, and (2)NPU (XRT) specific `aie.shim_dma_allocation` metadata, 

The condition switching between the two should be `isNPU()`, and not necessarily `isa<AIE::AIE2TargetModel>()`.


fifield (approved):
fwiw I intend to simplify this further. Hence https://github.com/Xilinx/mlir-air/pull/779 as a (very small) start. There is no reason to have two lowering paths in air-to-aie to encode the same metadata. At the very least the npu ops should get generated in the airrt to npu lowering.

erwei-xilinx:
> fwiw I intend to simplify this further. Hence #779 as a (very small) start. There is no reason to have two lowering paths in air-to-aie to encode the same metadata. At the very least the npu ops should get generated in the airrt to npu lowering.

Totally agreed that we should unify the shim allocation encoding metadata. Do you see the unified metadata encoding format as using `aie.shim_dma_allocation` or using `airrt.herd_metadata`?

PR #784, opened 2024-11-21
Aircc.py: `--omit-ping-pong-transform` argument

erwei-xilinx:
Enable fine-grained control over whether to enable air ping-pong buffering transformation passes.

Ping-pong buffering accelerates the program but at the cost of incurring extra memory and dma buffer descriptor (BD) usage, and should be disabled if a design is memory or BD constrained.


fifield (approved):
I wouldn't mind more of this, either opting into or out of optimizations, if it could mean more "standard" pipeline(s). Like get rid of all the hand written pipelines in `test/xrt` and only call aircc's pipeline(s) with a small number of options.

PR #802, opened 2024-12-06
XRT Runner: Cast `bfloat16` results to `np.float64` before checking against reference

erwei-xilinx:
When results are in `ml_dyptes.bfloat16`, cast them to `np.float64` before checking against reference results using `np.allclose`; `np.allclose` with `ml_dtypes.bfloat16` can lead to unexpected behaviour.

### DIFF
--- a/python/air/backend/xrt_runner.py
+++ b/python/air/backend/xrt_runner.py
@@ -131,18 +131,11 @@ class XRTRunner:
                 else:
                     print(actual)
 
-            if expected.dtype in [np.float16, np.float32, np.float64]:
+            if expected.dtype in [np.float16, np.float32, np.float64, bfloat16]:
+                if expected.dtype == bfloat16:
+                    expected = expected.astype(np.float64)
+                    actual = actual.astype(np.float64)
                 if not np.allclose(actual, expected, rtol=rtol):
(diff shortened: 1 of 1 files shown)


hunhoffe (approved):
Out of curiosity, do we know why the unexpected behavior occurs? mlir-aie is also using ml_dtypes for bfloat16, so I'm curious how this issue may need to be reported/handled there too.

erwei-xilinx:
> Out of curiosity, do we know why the unexpected behavior occurs? mlir-aie is also using ml_dtypes for bfloat16, so I'm curious how this issue may need to be reported/handled there too.

I was using `allclose` to compare two numpy arrays with dtype of `bfloat16`, and `allclose` throws me an error that I was comparing an array of `numpy.float16` against an array of `numpy.dtype[bfloat16]`. I haven't investigated this any further, nor have I found any online info reporting this. But my take is that it is safer to use `allclose` with numpy native types. Below is error message.

```
numpy.exceptions.DTypePromotionError: The DTypes <class 'numpy.dtypes.Float16DType'> and <class 'numpy.dtype[bfloat16]'> do not have a common DType. For example they cannot be stored in a single array unless the dtype is `object`.
```

PR #806, opened 2024-12-10
aircc: switch to one-shot-bufferize

fifield:
I missed this in last llvm update because it's not tested in CI until push to main.


fifield:
https://github.com/Xilinx/mlir-air/actions/runs/12246935034

PR #817, opened 2024-12-17
AIRSplitL2Memref: Fixup a bug around op type check

erwei-xilinx:
...that could lead to `offsetDefOp` being nullptr while still being check with `isa<>()`.


mlir/lib/Transform/AIRMiscPasses.cpp
```diff
@@ -894,8 +894,7 @@ Value tileChannelOpByFactor(air::ChannelInterface originalChanOp, int factor,
   Value zeroIdx = builder.create<arith::ConstantIndexOp>(loc, 0);
   if (!originalChanOp.getOffsets().empty()) {
     auto offsetDefOp = originalChanOp.getOffsets()[dim].getDefiningOp();
-    if ((offsetDefOp && isa<affine::AffineApplyOp>(offsetDefOp)) ||
-        isa<air::ExecuteOp>(offsetDefOp))
+    if (offsetDefOp && isa<affine::AffineApplyOp, air::ExecuteOp>(offsetDefOp))
```
fifield, line 897: Could probably use `llvm::isa_and_present` here to avoid the `&&`

erwei-xilinx: Good point. I created this PR to apply the `isa_and_present` and `dyn_cast_if_present` in the codebase. https://github.com/Xilinx/mlir-air/pull/821 Thanks for the comment!

PR #861, opened 2025-01-17
AIRParallelToHerd/Launch: Convert `forall` to `parallel` at the start of the pass

erwei-xilinx:
... so that we do not need extra implementations of forall-to-herd/launch.


fifield (approved):
makes sense.

PR #866, opened 2025-01-20
Add `depth=-1` mode to `air-par-to-herd/segment/launch`

erwei-xilinx:
… representing converting the innermost loop body into `air.hierarchy` op.

Assigning other negative values to depth means converting all loops; it would be ambiguous to assign depth=-2 as being the second innermost body, because that would be ambiguous.

### DIFF
--- a/mlir/include/air/Conversion/Passes.td
+++ b/mlir/include/air/Conversion/Passes.td
@@ -24,8 +24,10 @@ def ParallelToHerd : Pass<"air-par-to-herd", "ModuleOp"> {
   }];
   let options = [
     Option<"clAssignDepth", "depth", "int",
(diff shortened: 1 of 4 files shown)


fifield (approved):
This seems like a useful shortcut vs. knowing the target's loop depth.  The "any other negative" is the previous behavior? i.e. the option was initialized to -1 before

erwei-xilinx:
> This seems like a useful shortcut vs. knowing the target's loop depth. The "any other negative" is the previous behavior? i.e. the option was initialized to -1 before

Yes, in the previous version, any negative value means the same, i.e. converting all parallel ops. And that used to be the default behaviour.

PR #877, opened 2025-01-23
Remove feature introduced in C++20 to avoid hard dependence to C++20

erwei-xilinx:


mlir/lib/Util/Dependency.cpp
```diff
@@ -963,12 +963,17 @@ getAllReadAccessedMemrefOperandsFromOp(Operation *op) {
         std::get<2>(entry.second) = strides;
         return entry;
       };
-  auto pushMemrefEntryToVector = []<typename T>(T entry,
-                                                SmallVector<T> &vector) {
-    if (!isa<MemRefType>(entry.first.getType()))
-      return;
-    vector.push_back(entry);
-  };
```
keryell, line 971: On the other hand you could just use some C++14 Python-esque code like
```c++
 auto pushMemrefEntryToVector = [](auto entry, auto &vector) {
    if (!isa<MemRefType>(entry.first.getType()))
      return;
    vector.push_back(entry);
  };
```

erwei-xilinx: That works! Thanks for the comment, @keryell!

PR #908, opened 2025-02-21
AIRCC: Remove experimental passes from aircc.py

erwei-xilinx:
Those passes are now default passes for npu.

Also added a number of flags to control whether some passes get called.

### DIFF
--- a/python/air/backend/xrt.py
+++ b/python/air/backend/xrt.py
@@ -46,23 +46,26 @@ class XRTBackend(AirBackend):
     def __init__(
         self,
         verbose=False,
-        experimental_passes=False,
         omit_while_true_loop=False,
         omit_pingpong=False,
+        lower_linalg_to_func=False,
+        air_loop_fusion=False,
     ):
         """Constructor for XRTBackend
 
         Args:
             verbose: verbose output
-            experimental_passes: configure aircc to run additional experimental passes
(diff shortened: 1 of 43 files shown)


python/air/backend/xrt.py
```diff
@@ -46,23 +46,28 @@ class XRTBackend(AirBackend):
     def __init__(
         self,
         verbose=False,
-        experimental_passes=False,
         omit_while_true_loop=False,
         omit_pingpong=False,
+        lower_linalg_to_func=False,
+        air_loop_fusion=False,
+        runtime_loop_tiling_sizes: list[int] = [4, 4],
```
fifield, line 53: We got rid of one option but grew some new ones :/

fifield (approved):
I'm happy to see that the pass pipeline has graduated from "experimental".

python/air/backend/xrt.py
erwei-xilinx: Yeah... But the good thing is that they now more clearly specify what "experimental" features were being used. The `lower_linalg_to_func` is likely to be necessary to say if we want micro-kernel or direct code-generated (scalar, for now) code.

I wish the `air-loop-fusion` and `runtim-loop-tiling-sizes` can go eventually...

PR #922, opened 2025-02-28
Purge assertions in codebase

erwei-xilinx:


mlir/include/air/Util/Dependency.h
```diff
@@ -364,8 +364,10 @@ class dependencyTracer {
       dep_tracing_mode = 'w';
     else if (dep_type == "WAW/WAR")
       dep_tracing_mode = 'n';
-    else
-      assert(false && "Unknown dependency type");
+    else {
+      sink_air_op->emitOpError("Unknown dependency type.");
+      return;
+    }
```
fifield, line 370: This (and the other changes in this PR) make things slightly better, but the function(s) really should propagate the error up to the caller(s).

erwei-xilinx: Yup totally agreed. It's on my todo list to change from void to FailureOr types for those functions, and provide early termination.

erwei-xilinx: This PR is intended to fix up these issues: https://github.com/Xilinx/mlir-air/pull/925

PR #926, opened 2025-03-10
Modify `air.wait_all` canonicalizer to preserve wait_all ops without any users

erwei-xilinx:
… which might be used to represent a barrier intended to sync with host.


mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
```diff
@@ -1331,6 +1341,11 @@ static LogicalResult FoldWaitAll(WaitAllOp op, PatternRewriter &rewriter) {
     return success();
   }
 
+  // If the wait_all isn't used by any other ops, then it's a barrier intended
+  // to sync with the host.
+  if (op.use_empty())
+    return success();
```
fifield, line 1347: I think `if (op.getAsyncToken())` is more exact here.

erwei-xilinx: The -air-dependency is currently making sure that all wait_all ops return async tokens. Besides, I think the intent for this wait_all is to synchronize with the host, so that a token is still generated, but it's just intended to be used by some runtime op instead of any op running on the device.

erwei-xilinx:
Closed as the other PR is a more promising solution for handling the air.wait_all synchronization. https://github.com/Xilinx/mlir-air/pull/927

PR #929, opened 2025-03-11
Herd runtime parameter support for npu

erwei-xilinx:
Moving https://github.com/Xilinx/mlir-air/pull/585/files to here.

~~TODO: Make test [test/xrt/xx_mul_rtp_1x1/run.py](https://github.com/Xilinx/mlir-air/pull/929/files#diff-ade35890b3594c1501a541e4dd53ac6cc7e3fd4de8a71a836ee234a4053eddd5) work.~~

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -226,8 +226,7 @@ struct HerdLoadToNpuPattern : public OpConversionPattern<HerdLoadOp> {
       return WalkResult::interrupt();
     });
     if (size_x < 0 || size_y < 0 || loc_x < 0 || loc_y < 0) {
-      op.emitWarning(
-            "airrt.herd_metadata missing or incomplete.");
+      op.emitWarning("airrt.herd_metadata missing or incomplete.");
       return failure();
     }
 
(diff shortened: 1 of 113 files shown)


jgmelber:
https://github.com/Xilinx/mlir-aie/pull/2160

Should help.

erwei-xilinx:
> [Xilinx/mlir-aie#2160](https://github.com/Xilinx/mlir-aie/pull/2160)
> 
> Should help.

Yeah definitely, would love to explore on lowering `airrt.herd_load` with index operand to `aiex.set_lock` to the herd from runtime sequence.

mlir/lib/Conversion/AIRRtToNpuPass.cpp
```diff
@@ -198,6 +198,65 @@ struct HerdLoadToNpuPattern : public OpConversionPattern<HerdLoadOp> {
...
+          uint32_t v = cast<IntegerAttr>(constOp.getValue()).getInt();
+          rewriter.create<AIEX::NpuWriteRTPOp>(op.getLoc(), name, i, v);
+        }
+        // FIXME: this should depend on the metadata to enable and to get the id
+        if (op.getNumOperands())
+          rewriter.create<AIEX::NpuWrite32Op>(
+              op.getLoc(), 0x0001F000, 0x1, nullptr,
+              rewriter.getI32IntegerAttr(phys_x),
+              rewriter.getI32IntegerAttr(phys_y));
+      }
```
fifield, line 258: This is the code that could be simplified by https://github.com/Xilinx/mlir-aie/pull/2160. The main challenge, however, is tracking which lock was allocated by the air-to-aie pass. The FIXME here refers to the fact that the lock is named (with a symbol name), and that name should be used to lookup which lock to set, instead of assuming lock id=0 (which is true by constuction but not obvious).

PR #936, opened 2025-03-26
AIRCC: Expose more options from `aiecc` to `aircc` and `xrtBackend`

erwei-xilinx:
… to enable air to generate multi-pdi xclbins.


fifield (commented):
do you have tests?

erwei-xilinx:
> do you have tests?

I have some tests and will land them in separate PRs.

PR #940, opened 2025-04-10
Update lit cfg python for newer xdna driver

fifield:
Changes for runner software updates:
* Copy changes from mlir-aie
* Remove hard coding of clang-15 from workflow
* Disable failing tests xrt/23 and xrt/24


erwei-xilinx:
This is great. Would love to see whether this change could also fix the current random failures in the CI Runner.

fifield:
> This is great. Would love to see whether this change could also fix the current random failures in the CI Runner.

Sadly it does not do anything to fix the errors with newer driver.

PR #974, opened 2025-05-20
Update torch e2e tests

fifield:
* Update async dialect cpu target for upstream mlir and torch-mlir changes.
* Change hw torch-mlir tests to run on xrt and ryzen ai instead of legacy runtime

A few of the torch e2e tests are now running and passing again on the runner:
```
114.58s: AIRPYTHON :: torch_mlir_e2e/mul.py
70.24s: AIRPYTHON :: torch_mlir_e2e/relu.py
43.57s: AIRPYTHON :: torch_mlir_e2e/matmul.py
```
matmul depends on a transform script for linalg to air lowering. mul and relu depend on a very simple air pass pipeline implemented in `XRTBackend.compile_from_torch_mlir` with air-linalg-codegen, par to herd, par to launch, etc.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -46,6 +46,9 @@ jobs:
           pip cache purge
           pip install --upgrade pip
           pip install lit cmake joblib
+          pip install --pre torch-mlir torchvision \
+            --index-url https://download.pytorch.org/whl/nightly/cpu \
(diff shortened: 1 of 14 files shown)


fifield:
3ead1d0 should be it's own PR but is here temporarily for testing purposes.

It is all working locally so I'm not sure why it continues to have failures in the PR.

erwei-xilinx (approved):

PR #980, opened 2025-05-22
Update tests to support both chess and peano

erwei-xilinx:


test/xrt/01_air_to_npu/run_peano.lit
```diff
@@ -0,0 +1,7 @@
+// (c) Copyright 2023 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
fifield, line 2: mlir-air is MIT license. I see the same bug in the cut-and-paste source.

test/xrt/01_air_to_npu/run_peano.lit
```diff
@@ -0,0 +1,7 @@
...
+//
+// REQUIRES: ryzen_ai_npu1, peano
+// RUN: %python %S/gen.py --trace-size 131072 --trace-offset 65536
+// RUN: g++-13 %S/test.cpp -o test.exe -std=c++23 -Wall %test_utils_flags %xrt_flags -lrt -lstdc++ -lboost_program_options -lboost_filesystem
+// RUN: %run_on_npu1% ./test.exe -x air.xclbin -k MLIR_AIE -i air.insts.bin --trace_sz 131072
```
fifield, line 7: What selects between chess and peano here? We can't select based on `REQUIRES` because some systems will have both peano and chess.

Another thing to note is that if there are two .lit files in the same directory, like there are in this change, then each needs to explicitly run in its own subdirectory, to avoid collisions. e.g.:
```C++
// RUN: mkdir -p test_0
// RUN: cd test_0
// ...
```

test/xrt/01_air_to_npu/run_peano.lit
erwei-xilinx: Thanks for catching that! This is really helpful.

I fixed all lit test files touched in this PR, but will fix other source files in a separate PR.

test/xrt/01_air_to_npu/run_peano.lit
erwei-xilinx: Good point on separating build dirs. All xrt lit tests updated.

Re selection on chess and peano, I modified the approach such that if cmake param "PEANO_INSTALL_DIR" is set when building air, then we run peano tests only, else if chess is found, then we run chess tests only.

PR #988, opened 2025-05-26
Update Ryzen AI CI workflow to switch to peano from chess

erwei-xilinx:


.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -50,68 +50,45 @@ jobs:
...
-      - name: Build and Install mlir-aie
+      - name: Get llvm-aie
         run: |
           source air-venv/bin/activate
-          pushd mlir-aie
-
-          mkdir build
-          pushd build
-
-          export PATH=$VITIS/bin:$VITIS/aietools/bin:$PATH
```
fifield, line 77: Vitis should get added back to the path. The chess tests never run...
```

-- Unable to find v++
-- Unable to find aiecompiler
-- Unable to find xchesscc
-- Unable to find aietools directory
...
-- Could NOT find AIETools (missing: AIETOOLS_DIR AIETOOLS_BINARY_DIR AIETOOLS_INCLUDE_DIR AIETOOLS_XCHESSCC AIETOOLS_XCHESS_MAKE AIETools_VERSION_MAJOR AIETools_VERSION_MINOR AIE AIE2) 
```
npu tests:
```
Total Discovered Tests: 48
  Unsupported: 21 (43.75%)
  Passed     : 27 (56.25%)
```

programming_examples/lit.cfg.py
```diff
@@ -143,46 +143,76 @@
...
+# Test to see if this mlir-air installation is built for peano. If not, then look for Chess in system.
+try:
+    result = subprocess.run(
+        [os.path.join(config.peano_tools_dir, "llc"), "-mtriple=aie", "--version"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    if re.search("Xilinx AI Engine", result.stdout.decode("utf-8")) is not None:
+        config.available_features.add("peano")
+        config.substitutions.append(("%PEANO_INSTALL_DIR", config.peano_tools_dir))
```
fifield, line 159: `%PEANO_INSTALL_DIR` is only used to find peano clang++, we already add peano_tools_dir to PATH above (so `clang++` is already supposed to be peano clang?), and also define %CLANG to be the system clang. Seems like "finding clang" needs some cleanup.

maybe just `%peano_clang = path.join(config.peano_tools_dir, 'clang')`?  (or `%PEANO_CLANG` if shouting is preferred)

programming_examples/lit.cfg.py
```diff
@@ -143,46 +143,76 @@
...
+        print("Peano found: " + os.path.join(config.peano_tools_dir, "llc"))
+        tool_dirs.append(os.path.join(config.peano_tools_dir, "bin"))
+        # Peano compiler flags
+        peano_flags = (
+            "-O2 -std=c++20 --target=aie2-none-unknown-elf -DNDEBUG -I{}".format(
```
fifield, line 164: Best not to hardcode aie2 here. And most tests already have `--target=aie2-none-unknown-elf %peano_flags`

programming_examples/matrix_multiplication/bf16/Makefile
```diff
@@ -16,27 +25,35 @@ run2x4: compile-kernel
...
+		if [ -x "$(PEANO_INSTALL_DIR)/bin/clang++" ]; then \
+			echo "Auto-selected peano backend"; \
+			$(PEANO_INSTALL_DIR)/bin/clang++ ${PEANOWRAP2_FLAGS} -DBIT_WIDTH=8 -c ${srcdir}/mm.cc -o build/mm.o; \
+		elif command -v xchesscc_wrapper >/dev/null 2>&1; then \
+			echo "Auto-selected xchesscc backend"; \
+			cd build && ${powershell} xchesscc_wrapper aie2 -c ${srcdir}/mm.cc -o mm.o; \
+		else \
+			echo "Error: no usable backend found."; \
+			exit 1; \
+		fi \
```
fifield, line 56: Is this `compile-kernel` rule the same everywhere? maybe a `makefile.common` is appropriate.

programming_examples/matrix_multiplication/bf16/run_makefile_peano.lit
```diff
@@ -0,0 +1,8 @@
+// (c) Copyright 2025 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+//
+// REQUIRES: ryzen_ai_npu1, peano
+//
+// RUN: make -f %S/Makefile clean
```
fifield, line 6: This will collide with `run_makefile_chess.lit` in the build/test directory., it is necessary for each lit test to run in its own subdirectory.
for example,
```
// RUN: mkdir -p test
// RUN: cd test
```
Where 'test' is unique for each of the lit scripts in a directory.

fifield (changes requested):
Looks good so far, but I don't see the chess tests running.

.github/workflows/buildAndTestRyzenAI.yml
fifield: > Do you suggest that we should support a super air build that can work with both chess and peano?

Yes, traditionally it has been the case that aircc can support both. Passing `--peano` to aiecc simply sets the peano install directory, so that is ok. aiecc uses the `--xchesscc, --no-xchesscc` and `--xbridge --no-xbridge` to control whether chess or peano is used for compiling and linking. aircc can select defaults for these based on how it's compiled or whatever, but aircc also includes `--xchesscc, --no-xchesscc` and `--xbridge --no-xbridge` flags which should be passed through to aiecc to override the defaults.

erwei-xilinx: Thanks for the detailed description. Now the CI lit tests are able to switch between peano and chess compiler backend, using one single AIR build. The tests are still being separated into `check-air-e2e-peano' and `check-air-e2e-chess' for convenience; with `check-air-e2e', both peano and chess tests shall run.

programming_examples/lit.cfg.py
erwei-xilinx: E.g. https://github.com/Xilinx/mlir-air/pull/988/files#diff-fc56c471bd97abc443f8e27ef7f11925f3af6a4d153d3c960330262c69ad4ad8R7

programming_examples/lit.cfg.py
erwei-xilinx: Good point. Hard-coded flag removed.

programming_examples/matrix_multiplication/bf16/run_makefile_peano.lit
erwei-xilinx: Thanks for spotting this. I missed this one when I updated all tests. This is now fixed.

fifield (approved):

PR #1008, opened 2025-06-03
AIRRunner: Add ninja test target `check-air-runner`

erwei-xilinx:


.github/workflows/buildAndTestNoRuntime.yml
```diff
@@ -110,6 +110,7 @@ jobs:
           ninja check-air-cpp
           ninja check-air-mlir
           ninja check-air-python
+          ninja check-air-runner
```
fifield: Seems like this could run no problem in all workflows. It should be lightweight?

test/CMakeLists.txt
```diff
@@ -104,6 +104,14 @@ add_lit_testsuite(check-air-e2e-chess "Running AIR E2E tests (Chess backend only
...
 set_target_properties(check-air-e2e-chess PROPERTIES FOLDER "Tests")
 
+# AIR-Runner-only test suite
+add_lit_testsuite(check-air-runner "Running AIR-Runner tests (simulation only)"
+  ${CMAKE_CURRENT_BINARY_DIR}
+  DEPENDS ${TEST_DEPENDS}
+  ARGS ${AIR_TEST_LIT_ARGS} --filter "airrunner"
+)
+set_target_properties(check-air-e2e-peano PROPERTIES FOLDER "Tests")
+
```
fifield, line 114: Not sure if this is the right spot, but other targets have this:
```suggestion

add_dependencies(check-all check-air-runner)
```
so that you can type `ninja check-all` to get everything.

.github/workflows/buildAndTestNoRuntime.yml
erwei-xilinx: It can be slow, depending on the problem size. It generates traces with any scf.for loops unrolled, so it could takes seconds for some matmul shapes, say 512x512x512.

PR #1071, opened 2025-08-14
[Spensor] Add memory analysis pass

Yuyou-Fan:
This pass only includes one change:

1. Add the memory analysis pass. Since it's just a pure analysis pass and no code generation here, there is no IR tests for this pass.

### DIFF
--- a/python/spensor/spensor/passes/memory_analysis.py
+++ b/python/spensor/spensor/passes/memory_analysis.py
@@ -1,0 +1,76 @@
+# ./python/spensor/spensor/passes/memory_analysis.py -*- Python -*-
+#
+# Copyright (C) 2025, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+from xdsl.dialects.builtin import (
+    ModuleOp,
+    StringAttr,
+)
+from xdsl.context import Context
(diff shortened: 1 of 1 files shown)


python/spensor/spensor/passes/memory_analysis.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def apply(self, ctx: Context, op: ModuleOp):
+        walker = PatternRewriteWalker(
+            GreedyRewritePatternApplier([DeclareMemoryPattern(self.memory_mapping)])
+        )
+        walker.rewrite_module(op)
+        # asserts all memory names in allowing_load/store are existing
+        for val in self.memory_mapping.values():
+            for s in val.allowing_load:
+                assert s in self.memory_mapping
```
erwei-xilinx, line 73: I wonder if we can emit error and signal pass failure, instead of assertion?

Yuyou-Fan: > I wonder if we can emit error and signal pass failure, instead of assertion?

I can replace it with some thing like 
```
raise ValueError("Cannot find" + s + "in memory mapping")
```

Do you think it would be better?

PR #1072, opened 2025-08-15
[Spensor] Add test for Spensor ops

Yuyou-Fan:
This file include two changes:
1. Add a Python script for testing Spensor operations
2. Update `python/test/lit.cfg.py` with adding possible Spensor feature

### DIFF
--- a/python/test/spensor/spensor_ops.py
+++ b/python/test/spensor/spensor_ops.py
@@ -1,0 +1,116 @@
+# ./python/test/spensor/spensor_ops.py -*- Python -*-
+#
+# Copyright (C) 2025, Xilinx Inc.
+# Copyright (C) 2025, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+# REQUIRES: spensor
+
+# RUN: %PYTHON %s | FileCheck %s
+
+from xdsl.dialects.builtin import (
+    ModuleOp,
+    IndexType,
+    TensorType,
+    f32,
+    FunctionType,
+)
+from xdsl.context import Context
+from xdsl.dialects import func, arith
+import spensor.dialects.spensor_dialect as spensor
+
+def getConsantIndex(val: int):
+    return arith.ConstantOp.from_int_and_width(val, IndexType())
+
+
+def getReduceCascade(ctx: Context) -> ModuleOp:
+    L1_memory = spensor.MemoryType("L1", [6])
+    L2_memory = spensor.MemoryType("L2", [])
+    l3_memory = spensor.MemoryType("L3", [])
+    declare_l1 = spensor.DeclareMemoryOp(
+        "L1", L1_memory.memory_shape, ["L2", "L3"], ["L2", "L3"], 2
+    )
+    declare_l2 = spensor.DeclareMemoryOp(
+        "L2", L2_memory.memory_shape, ["L3"], ["L3"], 1
+    )
+    declare_l3 = spensor.DeclareMemoryOp("L3", l3_memory.memory_shape, [], [])
+
+    spensor_input_type = spensor.SpensorType(TensorType(f32, [16, 32]), l3_memory)
+    spensor_output_type = spensor.SpensorType(TensorType(f32, [16, 1]), l3_memory)
+    function_type = FunctionType.from_lists(
+        [
+            spensor_input_type,
+        ],
+        [spensor_output_type],
+    )
+    func_op = func.FuncOp("reduce", function_type)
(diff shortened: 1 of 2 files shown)


python/test/spensor/spensor_ops.py
```diff
@@ -0,0 +1,116 @@
+# ./python/test/spensor/spensor_ops.py -*- Python -*-
+#
+# Copyright (C) 2025, Xilinx Inc.
```
erwei-xilinx: We should remove the Xilinx line for new files created after the merger.

python/test/spensor/spensor_ops.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    func_op = func.FuncOp("reduce", function_type)
+    block = func_op.body.block
+    arg0 = block.args[0]
+
+    const_1 = getConsantIndex(1)
+    const_0 = getConsantIndex(0)
+    const_4 = getConsantIndex(4)
+
+    # arg0: <16x32xf32> at L3
```
erwei-xilinx: Where does the 16 and 32 come from? If they are showing an example of how they're parsed, it might be useful to specify with say e.g. to avoid confusion.

python/test/spensor/spensor_ops.py
```diff
@@ -0,0 +1,116 @@
...
+# CHECK-NEXT:     %10 = "spensor.reduce_sum"(%9) : (!spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x4xf32>, !spensor.memory<"L1", [6 : i64]>>>>) -> !spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L1", [6 : i64]>>>>
+# CHECK-NEXT:     %11 = "spensor.move"(%10) <{memory_name = "L2"}> : (!spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L1", [6 : i64]>>>>) -> !spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L2", []>>>>
+# CHECK-NEXT:     %12 = "spensor.move"(%11) <{memory_name = "L3"}> : (!spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L2", []>>>>) -> !spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L3", []>>>>
+# CHECK-NEXT:     %13 = "ndspensor.combine_to_spensor"(%12) <{combine_index = [0 : i64]}> : (!spensor.ndspensor<tensor<1x!spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L3", []>>>>) -> !spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L3", []>>
+# CHECK-NEXT:     func.return %13 : !spensor.spensor<tensor<16x1xf32>, !spensor.memory<"L3", []>>
+# CHECK-NEXT:   }
+# CHECK-NEXT: }
```
erwei-xilinx: NitPick: Add new line at EoF.

Yuyou-Fan: > NitPick: Add new line at EoF.

New line added

python/test/spensor/spensor_ops.py
Yuyou-Fan: > We should remove the Xilinx line for new files created after the merger.

Xilinx removed! But I copied from this file: https://github.com/Xilinx/mlir-air/blob/main/python/test/torch_mlir_e2e/add.py

erwei-xilinx: > Xilinx removed! But I copied from this file: https://github.com/Xilinx/mlir-air/blob/main/python/test/torch_mlir_e2e/add.py

Thanks. That file is different because it was created in 2022 before the Xilinx-AMD merger.

python/test/spensor/spensor_ops.py
Yuyou-Fan: > Where does the 16 and 32 come from? If they are showing an example of how they're parsed, it might be useful to specify with say e.g. to avoid confusion.

Updated with a comment saying the function type, it should be clearer https://github.com/Xilinx/mlir-air/pull/1072/commits/7f8cf54288dad379bfbd496b78b95a206b68f40e#diff-4f6f3f67f33757fab19f98045ff71a2c90731616c667b712f2592715e576d914R45

erwei-xilinx (approved):

PR #1073, opened 2025-08-15
[Spensor] Add Constant Append Pass

Yuyou-Fan:
This PR includes two changes:

1. Add Append Constant Pass used in lowerings
2. Add a test for Constant Pass

### DIFF
--- a/python/spensor/spensor/passes/append_constant.py
+++ b/python/spensor/spensor/passes/append_constant.py
@@ -1,3 +1,8 @@
+# ./python/spensor/spensor/passes/append_constant.py -*- Python -*-
+#
+# Copyright (C) 2025, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
 from xdsl.dialects.builtin import (
     ModuleOp,
 )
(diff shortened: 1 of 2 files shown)


Yuyou-Fan:
> Can you enable the spensor installation in the github ci workflow, say in [here](https://github.com/Xilinx/mlir-air/blob/db750df17c093c00b6c1737372f2817b78768d12/.github/workflows/buildAndTestRyzenAI.yml#L154)? So that we can test spensor tests using `ninja check-air-python`?
> 
> Cc: @fifield

Thanks for pointing it out!
There are four test files, should I add to all of them?

erwei-xilinx:
> Thanks for pointing it out! There are four test files, should I add to all of them?

I think of the four workflow files, we can start with the one that I pointed to, because it deploys to a runner with an NPU on it. Eventually when we land spensor board tests, this is the workflow file to test those, anyway.

For the other three, maybe we can wait for now. Maybe we can add a new workflow file, which is hosted on a generic ubuntu runner, which tests for air + spensor only.

erwei-xilinx (approved):

PR #1082, opened 2025-08-22
Add a new example showing vectorized matmul tiling in transform dialect

erwei-xilinx:
Depends on https://github.com/Xilinx/mlir-aie/pull/2526 and https://github.com/Xilinx/mlir-aie/pull/2524 and then bump mlir-aie.


erwei-xilinx:
Cc: @stephenneuendorffer @fifield

PR #1159, opened 2025-10-21
Support for varying tile sizes for matrix multiplication with runner and profiler support

AdamLow1:
- Adjusted mm.cc so that linalg fill function has correct name for tile sizes
- Updated makefile so that tile sizes are passed into kernel compile
- Added support for simulating design with air-runner
- Added variable tile size support for running on hardware, profilingon hardware, and with air-runner
- Updated default L1 tile sizes to values which have slightly better performance
- Added a bried readme

### DIFF
--- a/programming_examples/matrix_multiplication/bf16/mm.cc
+++ b/programming_examples/matrix_multiplication/bf16/mm.cc
@@ -277,15 +277,17 @@ extern "C" {
 // the kernels.
 
 #ifndef DIM_M
-#define DIM_M 64
+#define DIM_M 128
+#define DIM_M_DIV_4 32
 #endif
 
 #ifndef DIM_K
-#define DIM_K 64
+#define DIM_K 32
 #endif
 
 #ifndef DIM_N
 #define DIM_N 64
+#define DIM_N_DIV_4 16
 #endif
 
 #ifndef combos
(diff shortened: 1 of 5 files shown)


programming_examples/matrix_multiplication/bf16/mmult_aie2.py
```diff
@@ -0,0 +1,168 @@
+# mmult_aie2.py -*- Python -*-
```
erwei-xilinx, line 1: Could you please run `black` on this file to auto-update its code format? This is needed to unblock merging this PR.

programming_examples/matrix_multiplication/bf16/mm.cc
```diff
@@ -308,14 +311,23 @@ extern "C" {
...
     zero_vectorized<ctype_out, DIM_M, DIM_N, 32>(c_out);                       \
   }
 
-combos(matmul_vectorized_c_func) combos(matmul_scalar_c_func)
-    combos(zero_vectorized_c_func)
+combos(matmul_vectorized_c_func) combos(matmul_scalar_c_func) combos(zero_vectorized_c_func)
 
 } // extern "C"
+
+//CAT9(linalg_fill_, mlir_in, _view1x1x, DIV4(N), x, DIV4(M), x4x4x, mlir_out, as2)
```
erwei-xilinx, line 332: NitPick: Commented code.

erwei-xilinx:
Landed in https://github.com/Xilinx/mlir-air/pull/1169.

PR #1243, opened 2026-01-22
Add air.herd and segment verifiers

erwei-xilinx:
Add verifier that returns error message when memref.alloc in their body have incompatible memory space.


erwei-xilinx:
Depends on this experimental pr: https://github.com/Xilinx/mlir-air/pull/1244

PR #1279, opened 2026-02-04
Add a new triton matmul example demonstrating `matmul`+`truncf` fusion

erwei-xilinx:
Fusion is performed using shared vector register. Key transform op: `transform.air.fuse_truncf_linalg`.

Input IR generated via torch inductor.


erwei-xilinx:
Cc: @amd-jiajli

PR #1287, opened 2026-02-06
feat(aircc): add --debug-ir mode to emit IR after each individual compiler pass

amd-jiajli:
This adds a debugging feature to aircc that enables fine-grained inspection of intermediate IRs throughout the compilation pipeline.

When enabled via --debug-ir flag (or debug_ir=True in XRTBackend/XRTRunner):

1. Each MLIR pass is run individually instead of as a batched pipeline
2. IR is saved after each pass to <tmpdir>/debug_ir/pass_XXX_after_<name>.mlir
3. A pass.log file documents all passes executed with sequence numbers
4. Checkpoint markers identify which debug IRs correspond to standard output files (placed.air.mlir, aie.air.mlir, npu.air.mlir)

This is useful for:
- Understanding the compilation pipeline flow
- Debugging pass failures by identifying exactly which pass caused issues
- Analyzing IR transformations for optimization work

Usage:
  aircc.py --debug-ir input.mlir 
  XRTBackend(debug_ir=True).compile(module) 
  XRTRunner(debug_ir=True).run_test(module, ...)


erwei-xilinx (approved):
Thanks for working on this! I think this is awesome.

PR #1289, opened 2026-02-10
[42 Triton softmax BF16] Add profiling support, make input IR an argument, use input IR that triggers the efficient AIE computation

amd-jiajli:
42 Triton softmax BF16 test: add profiling support
Includes:
- Makefile with profiling support and more
- Input IR as an argument. Support different input IR with different input dimension.

### DIFF
--- a/test/xrt/42_triton_softmax_bf16/run.py
+++ b/test/xrt/42_triton_softmax_bf16/run.py
@@ -25,8 +25,10 @@ parser.add_argument(
     "--input-mlir",
     type=str,
     dest="input_mlir",
-    default=os.path.join(script_dir, "input_ir/4x1024_mul_inv_bf16.mlir"), # A softmax kernel for 4x1024 input tile. We test with 256x1024 input. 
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Some of the failing tests should work after bumping aie: https://github.com/Xilinx/mlir-air/pull/1290

test/xrt/42_triton_softmax_bf16/run.py
```diff
@@ -35,9 +47,34 @@
     "--N",
     type=int,
     dest="N",
-    default=256,
+    # default=256,
```
erwei-xilinx: NitPick: Commented code.

amd-jiajli: Left here because the previous `input_ir/original.mlir` (also left in commented code) requires N=256. But I can remove them though

erwei-xilinx: I see. Would be useful to keep the comment that 256 is for `input_ir/original.mlir`

PR #1354, opened 2026-03-04
Switch Peano (llvm-aie) from pinned version to latest nightly

erwei-xilinx:
## Summary
- Replace pinned `llvm-aie 19.0.0.2025071101+b3cd09d3` wheel URL with `expanded_assets/nightly` pip index pattern in both the build script and CI workflow
- This resolves the `TODO: Use nightly latest llvm-aie once it is fixed` comment
- Matches the pattern already used by upstream mlir-aie CI

## Test plan
- [x] CI passes with nightly Peano
- [x] Investigate flash_attention kernel_fusion failure with Peano 20 (100% mismatch with near-zero output on `attn.cc` compiled with `-DAIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16`)


erwei-xilinx:
## Root cause identified: stack overflow into DMA buffer

### Summary

The `vector_div` NPU2 CI failure is caused by a **stack overflow into an adjacent DMA buffer**, not a codegen bug in llvm-aie.

### Details

The AIE linker script reserves exactly `stack_size` bytes (default 1024, set via `aie.core` op attribute) for the stack, then places DMA buffers immediately after. The `vector_div` kernel's full call chain is:

| Function | Frame size | Cumulative SP offset |
|----------|-----------|---------------------|
| `core_0_2` | 960–1024 bytes (register spills) | 960–1024 |
| `__aie2p_scalar_fdiv` | 64 bytes | 1024–1088 |
| `__mulsf3` | 64 bytes | 1088–1152 |

The linker places `buf5` at `_sp_start_value_DM_stack + 0x400` (1024). So `__mulsf3`'s stack frame writes into `buf5` in **both** the old and new llvm-aie versions. The llvm-aie `orderFrameObjects` commit (`f02c358e`) reduced the root frame from 1024 to 960 bytes, which shifted which bytes of `buf5` get overwritten by the callee frames, making the latent corruption visible as a wrong division result.

### Verification

Setting `stack_size=2048` on the `aie.core` op (giving enough room for the full 1152-byte call chain) makes `vector_div` pass even with the "bad" llvm-aie version that has `orderFrameObjects` sorting enabled. Confirmed on local Strix NPU2 hardware.

### Fix

PR #1547 adds a `--stack-size` option to `aircc` (threaded through `air-to-aie` pass, `XRTBackend`, and `XRTRunner` Python APIs) so users can increase the per-core stack allocation. The `vector_div` example is updated to use `stack_size=2048`.

For this PR (#1354), switching to latest nightly Peano should work once the `vector_div` example uses the increased stack size from #1547.

Issue #1414, opened 2026-03-12
New op: air.rank — multi-device / multi-host parallelism above air.launch


SamuelBayliss:
## Summary

Add a new `air.rank` operation and supporting `air.universe` type to the AIR dialect.
`air.rank` is a fourth hierarchy level that sits above `air.launch` and expresses a
communicating world of rank instances, where each instance corresponds to a complete
GPU device or a CPU host process (with its attached devices).

A detailed design document is in `docs/AIRRankOp.md` on the `compute-model-explanation`
branch. This issue captures the implementation work.

---

## Motivation

The existing three-level hierarchy (`air.launch` → `air.segment` → `air.herd`) describes
work that runs on a **single device**. There is no dialect-level primitive for expressing
programs that span multiple GPUs or multiple hosts. Without such a primitive, multi-device
programs must be assembled entirely at the runtime or framework layer, losing the
compiler's ability to reason about inter-device data movement, resource allocation,
and communication scheduling.

`air.rank` fills this gap by extending the hierarchy upward:

```
air.rank           — communicating world of devices or hosts    ← NEW
  air.launch       — co-resident work on one device
    air.segment    — tile rectangle + L2 memory
      air.herd     — array of PEs
```

---

### `air.rank`

```
[%token =] air.rank (%r₀, …, %rₙ) in (%sr₀=%M₀, …, %srₙ=%Mₙ)
           args(%a₀=%v₀, …) : <types>
           [universe = %u]
           [dependency  = [%t₀, …]]
           [affinity    = [%t₀, …]]
           {
  …
  air.rank_terminator
}
```

**Key properties:**

- Defines an N-dimensional iteration space. Each point is a **rank instance**.
- Each rank instance maps to:
  - A distinct **GPU device** when nested inside `air.launch`.
  - A distinct **host process** (with attached devices) when at top level.
- The body is `IsolatedFromAbove`; values are passed via explicit kernel operands.
- The compiler is free to distribute the serial preamble of the body across all
  participating rank instances (SPMD fork semantics).
- **May-be-parallel** semantics (like `air.launch`): `concurrency` token list is
  not permitted. `dependency` and `affinity` lists are permitted.
- A `universe` operand of type `!air.universe` constrains the physical pool from
  which rank instances are scheduled.

**Terminator:** `air.rank_terminator` (analogous to `air.launch_terminator`).

### `air.universe` type and `air.universe.alloc`

```
%u = air.universe.alloc(%capacity) : !air.universe
[...]

PR #1481, opened 2026-03-31
Add air.rank Python API and pipeline integration

erwei-xilinx:
## Summary
- Add `Rank` builder class and `@rank` decorator to `_air_ops_ext.py`, following the same pattern as `Launch`/`Segment`/`Herd`
- Insert `air-rank-to-launch` pass into `aircc.py` pipeline (before `air-insert-launch-around-herd`), enabling automatic serialization of `air.rank` ops during compilation
- Add Python builder and decorator tests verifying the full 4-level hierarchy (`rank > launch > segment > herd`)

Follows up on #1479 (Phase 1: dialect and C++ implementation). Part of #1414.

## Test plan
- [x] `check-air-python`: 8/8 passed
- [x] `check-air-mlir`: 349 passed, 0 failures
- [x] CI: Build and Test
- [x] CI: clang-tidy

### DIFF
--- a/python/air/compiler/aircc/main.py
+++ b/python/air/compiler/aircc/main.py
@@ -813,6 +813,7 @@ extern "C" {
         air_placed = opts.tmpdir + "/placed." + air_mlir_filename
         pass_pipeline = ",".join(
             [
+                "air-rank-to-launch",
                 "air-insert-launch-around-herd{insert-segment=true}",
                 "func.func(air-lower-herd-parallel)",
                 "scf-forall-to-parallel",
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Regarding the Copilot review comments:

1. **Mutable default arguments** (`sizes=[]`, `async_dependencies=[]`, etc.): This matches the existing convention used by `Launch`, `Segment`, and `Herd` in the same file. All four classes use the same pattern. Fixing only `Rank` would create an inconsistency — fixing all four is out of scope for this PR.

2. **Unused `attributes`/`loc`/`ip` parameters**: Same situation — `Launch`, `Segment`, and `Herd` all accept these without forwarding them. `Rank` follows the existing pattern for consistency. Plumbing these through across all four classes would be a separate cleanup PR.

PR #1523, opened 2026-04-09
Fix TileDMAAllocator S2MM packet-flow channel reuse

erwei-xilinx:
## Summary
- Restrict packet-flow channel reuse in `TileDMAAllocator::simpleDmaChannelAlloc` to MM2S (outbound) channels only
- Previously, S2MM (inbound) channels were also reused across packet flows on the same tile, which made it impossible for the packet switch to route different packet IDs to different buffer descriptors
- The fix adds an `isMM2S.value()` guard to the existing `isPacketFlowOp` check

## Details
When multiple `dma_packet` channels target the same destination tile, each needs its own S2MM DMA channel so the packet switch can distinguish between packets with different IDs and route them to separate buffer descriptors. The previous code unconditionally reused the first packet-flow allocation found on the tile for all subsequent packet-flow allocations, causing routing conflicts at the destination.

For MM2S (outbound) channels, reuse is correct — multiple outbound packet flows from the same tile can share a single DMA channel since they are multiplexed by the packet switch based on packet ID.

## Test plan
- [ ] Existing `check-air-mlir` tests pass (no regressions)
- [ ] Multi-destination packet-flow patterns now get separate S2MM channel allocations per destination flow

### DIFF
--- a/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
+++ b/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
@@ -878,8 +878,11 @@ air::TileDMAAllocator::simpleDmaChannelAlloc(air::MemcpyInterface &memcpyOp,
       return t;
     }
     // Search for existing packet-flow allocations on this tile, and try to
-    // reuse the channel allocation.
-    if (isPacketFlowOp && t.foundPacketFlowAllocInTile(col, row)) {
+    // reuse the channel allocation. Only reuse for MM2S (outbound) channels;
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Closing: This fix is not needed. The existing TileDMAAllocator S2MM packet-flow channel reuse is correct behavior — multiple packet flows can share the same destination DMA channel because the packet switch matches by packet ID at the BD level. The isMM2S guard would break existing designs that rely on this sharing. The dual-herd example passes without this change.

PR #1528, opened 2026-04-10
Auto-detect packet switching need in air-dma-to-channel

erwei-xilinx:
## Summary
- Auto-detect when packet switching is needed for multi-herd designs in the air-dma-to-channel pass
- When a segment has multiple herds and the per-column L3-bound channel count exceeds 2 (the physical shim DMA limit), input channels are auto-upgraded to channel_type=dma_packet
- Also unrolls scf.parallel around shim-side channel ops in air-to-aie so each tile gets a discrete channel put with correct packet ID

## Details
Two changes:
1. **AIRDmaToChannel.cpp**: Post-processing after channel creation counts per-column channel pressure. If input channels exceed 2 per column across herds, they are marked as dma_packet.
2. **AIRToAIEPass.cpp**: Runs scf.parallel unrolling on the parent function scope before shim-side channel op collection. Without this, DMA-generated channel puts stay wrapped in scf.parallel and all tiles get the same packet ID.

## Dependencies
- #1525 (ShimDMA packet-flow channel reuse) must be merged first

## Test plan
- [x] check-air-mlir passes (360 tests, no regressions)
- [x] New LIT test dma_to_channel_auto_packet.mlir verifies 4 input channels get dma_packet
- [x] Hardware-verified on NPU2 with dma_memcpy_nd dual-herd example (PASS)

### DIFF
--- a/mlir/include/air/Transform/Passes.td
+++ b/mlir/include/air/Transform/Passes.td
@@ -1595,6 +1595,13 @@ def DmaToChannel : Pass<"air-dma-to-channel", "ModuleOp"> {
     }
     ```
   }];
+  let options = [
+    Option<"clShimDmaChannelsPerCol", "shim-dma-channels-per-col", "int",
+           /*default=*/"2",
+           "Number of physical DMA channels per direction per shim tile "
+           "column. Used by auto-packet-switching detection to determine "
+           "when channel count exceeds physical capacity.">,
+  ];
 }
 
 def AIROverrideMemRefMemorySpace : Pass<"air-override-memref-memory-space", "ModuleOp"> {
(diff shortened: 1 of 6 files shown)


erwei-xilinx:
The updated threshold `totalCapacity = shimChannelsPerCol * herdWidth` is too permissive for multi-column herds. With an 8x1 herd, totalCapacity = 2*8 = 16, so 4 input channels don't trigger auto-detection. But the actual constraint is per-column: each channel creates flows on ALL columns (the channel bundle size is [8,1], meaning one flow per column). So per-column pressure is still 4 channels, exceeding the 2-per-column physical limit.

The correct comparison is per-column: `inputChannels.size() > shimChannelsPerCol`, because every channel targets every column after segment unrolling. The `totalCapacity` approach would only be correct if channels were distributed across columns (e.g., channel A only targets columns 0-3, channel B only targets columns 4-7), but that's not how the compilation works — all channels target all columns.

Tested: the 8x1 dual-herd example fails with "out of channels" using the updated threshold, but passes with the per-column comparison.

PR #1530, opened 2026-04-11
Fix DMA offset for transposed memrefs and simplify test 55

erwei-xilinx:
## Summary

- Fix a bug in `extractOperandsFromReinterpretCast` (`ConvertToAIRPass.cpp`) where a single flat offset from a `reinterpret_cast` was placed in the wrong DMA dimension for transposed memrefs
- Simplify test 55 from a 4-herd pattern to a 3-herd pattern, merging the truncf herd into the compute herd

## Bug description

When `air-copy-to-dma` converts a `memref.copy` backed by a `reinterpret_cast` with a single flat offset (e.g., a launch-tile M offset of 256) and a transposed layout (strides `[1, 504]`), the old code padded zeros at the front of the offset array:

```
offsets = [0, 256]   strides = [1, 504]
→ flat_addr = 0*1 + (256+i1)*504 + i0    // WRONG: offset*504
```

The fix places the flat offset in the stride-1 dimension instead:

```
offsets = [256, 0]   strides = [1, 504]
→ flat_addr = (256+i0)*1 + (0+i1)*504    // CORRECT: offset*1
```

This produced NaN values for the second launch tile (M=256..511) because the DMA read from out-of-bounds memory. The first tile (M=0..255) was unaffected since offset=0 gives the same result either way.

## Test 55 simplification

The original test 55 used a 4-herd pattern (prologue, truncf, compute, epilogue) to separate `truncf` (f32→bf16) from the matmul, based on the assumption that the combined pattern fails on NPU1. Testing with the DMA fix reveals the combined truncf+matmul pattern works correctly on NPU1 (aie2). The 4-herd separation was only needed because the DMA offset bug caused failures that were misattributed to truncf+matmul incompatibility.

The simplified 3-herd pattern (prologue, compute, epilogue) merges DMA + truncf + matmul into a single compute herd, producing correct results on NPU1 hardware (verified with M=500, N=500, K=784).

## Test plan

- [x] Run test 55 on NPU1 with 3-herd pattern: PASS (M=500, N=500, K=784)
- [x] Run Triton-XDNA padded matmul example on NPU1 with the DMA fix: PASS (M=500, N=500, K=1024, all 109 sampled elements match)
- [ ] CI: `test/xrt/55_matmul_padding_bf16_npu1/run_npu1_peano.lit`

### DIFF
--- a/mlir/lib/Conversion/ConvertToAIRPass.cpp
+++ b/mlir/lib/Conversion/ConvertToAIRPass.cpp
@@ -99,9 +99,34 @@ static void extractOperandsFromReinterpretCast(
     sizes.push_back(getValueOrCreateConstantIndexOp(builder, loc, ofr));
   for (auto ofr : reinterpretCast.getMixedStrides())
     strides.push_back(getValueOrCreateConstantIndexOp(builder, loc, ofr));
-  while (offsets.size() < sizes.size())
-    offsets.insert(offsets.begin(),
-                   arith::ConstantIndexOp::create(builder, loc, 0));
+  // When the reinterpret_cast has fewer offset dimensions than the memref
+  // rank (e.g., a single flat offset for a 2D memref), we need to place
+  // the flat offset in the correct dimension. For transposed memrefs
+  // (stride-1 in the first dimension), the flat offset corresponds to the
+  // stride-1 dimension, not the last dimension. Find the stride-1
+  // dimension and place the offset there; pad others with zero.
+  if (offsets.size() < sizes.size()) {
+    int strideOneIdx = static_cast<int>(strides.size()) - 1;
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Addressed both review comments:

**Copilot (standalone regression test):** The commit already includes three standalone FileCheck tests in `subview_reinterpret_cast_to_dma.mlir` that exercise the direct `memref.copy %rc, %dst` path without an intervening `memref.subview`:
- `@standalone_transposed`: dynamic offset with strides `[1, 504]`
- `@standalone_normal`: dynamic offset with strides `[504, 1]`  
- `@standalone_transposed_const_offset`: constant offset 256 with strides `[1, 504]`

**clang-format:** Fixed in the latest push.

PR #1548, opened 2026-04-22
Fix wheel release timestamp drift and duplicate-hash check

erwei-xilinx:
## Summary

- **Timestamp consistency**: Each matrix job was computing `DATETIME=$(date +"%Y%m%d%H")` independently. If jobs straddled an hour boundary, wheels from the same commit got different version strings (e.g. `0.0.1.2025042123+abc1234` vs `0.0.1.2025042200+abc1234`). Added a `get-air-project-commit` preliminary job that computes `AIR_PROJECT_COMMIT` and `DATETIME` once and passes them as job outputs, following the pattern from mlir-aie's `mlirAIEDistro.yml`.
- **Duplicate-hash check**: `.assets[0].name` returned the oldest asset (GitHub API sorts ascending by `created_at`), not the newest. Replaced with a jq expression that sorts by `created_at` descending and picks the most recent `mlir_air` wheel.
- **Asset accumulation**: Set `replacesArtifacts: true` unconditionally to prevent stale same-name assets from accumulating in the release (matches mlir-aie's approach).

## Test plan

- [x] Verify YAML syntax is valid and job dependency chain is correct: `check-new-commits` → `get-air-project-commit` → `build-repo` / `build-windows`
- [x] After merge, trigger a `workflow_dispatch` run and confirm all matrix jobs produce wheels with identical `DATETIME+HASH` in their version strings

### DIFF
--- a/.github/workflows/buildAIRWheels.yml
+++ b/.github/workflows/buildAIRWheels.yml
@@ -45,33 +45,42 @@ jobs:
           CURRENT_SHORT=$(git rev-parse --short=7 HEAD)
           echo "Current HEAD: $CURRENT_SHORT"
 
-          # Get the most recently uploaded wheel from the latest release to extract the commit hash
-          LAST_WHEEL=$(gh api repos/${{ github.repository }}/releases/tags/latest-air-wheels \
-            --jq '.assets | sort_by(.created_at) | reverse | map(select(.name | startswith("mlir_air")) | .name) | .[0]' \
-            2>/dev/null || echo "")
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Addressed Copilot feedback:

- **`first` → `.[0]`**: Applied in 32c4be7c. Using `.[0]` is simpler and avoids jq/gojq compatibility concerns.
- **No-RTTI tag not checked**: Valid observation, but pre-existing — the `check-new-commits` job was already only checking `latest-air-wheels` before this PR. Keeping out of scope to avoid expanding the change.

PR #1550, opened 2026-04-22
Fix hash regex to handle _no_rtti suffix in wheel names

erwei-xilinx:
## Summary

- Follow-up to #1548. The `check-new-commits` hash extraction regex required a dash (`-`) immediately after the 7-char commit hash, but no-RTTI wheel names have `_no_rtti` between the hash and the dash (`+c8ec089_no_rtti-cp310-...`). The regex never matched, so `check_tag` always reported `latest-air-wheels-no-rtti` as stale, triggering redundant builds on every scheduled run even when HEAD hadn't changed.
- This is the root cause of the duplicate `c8ec089` releases with timestamps `2026042202` and `2026042205`.
- Fix: match any character after the 7 hex digits instead of requiring a dash.

## Test plan

- [ ] Verify regex matches both RTTI (`+c8ec089-cp310-...`) and no-RTTI (`+c8ec089_no_rtti-cp310-...`) wheel names
- [ ] After merge, confirm next scheduled run at 4 AM UTC correctly skips if HEAD hasn't changed

### DIFF
--- a/.github/workflows/buildAIRWheels.yml
+++ b/.github/workflows/buildAIRWheels.yml
@@ -60,8 +60,9 @@ jobs:
             fi
 
             # Extract short hash from wheel name: mlir_air-0.0.1.DATETIME+HASH-...
+            # No-RTTI wheels have _no_rtti after the hash, so match any char after the 7 hex digits.
             local LAST_HASH
-            LAST_HASH=$(echo "$LAST_WHEEL" | sed -n 's/.*+\([a-f0-9]\{7\}\)-.*/\1/p')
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Addressed both Copilot comments in e81ebeb6:

- **`.[0]` returning `null`**: Added `// empty` so jq outputs nothing instead of the literal string `"null"` when no assets match, allowing the `-z` check to fire correctly.
- **Loose sed boundary**: Changed `.*` after the 7 hex digits to `[^a-f0-9].*`, requiring a non-hex boundary to avoid false matches on longer hex sequences.

PR #1579, opened 2026-05-03
[multi-gpu] Phase 4: air-symmetric-alloc-to-mgpu lowering pass

erwei-xilinx:
## Summary

Phase 4 of the multi-GPU stack: a new `air-symmetric-alloc-to-mgpu` conversion pass that uplevels symmetric-heap allocations from the runtime-ABI level (`mgpuSymmetricAlloc` + a hand-built memref descriptor) to the standard memref dialect.

## What it does

Replaces the runtime-ABI dance:

```mlir
%ptr   = call @mgpuSymmetricAlloc(size, stream)
%bytes = wrap_bytes(%ptr, size)            // hand-built memref<?xi8>
%m     = memref.view %bytes[0][]           // retype to memref<T>
...
call @mgpuSymmetricFree(%ptr, stream)
```

with the MLIR-native form:

```mlir
%m = memref.alloc() {air.symmetric} : memref<T>
...
memref.dealloc %m
```

The pass lowers each `memref.alloc {air.symmetric}` to a runtime call plus an LLVM memref descriptor build + `unrealized_conversion_cast` back to the original memref type (so downstream uses keep working through `convert-to-llvm`). Each `memref.dealloc` whose operand traces back through the cast becomes `mgpuSymmetricFree`. No-op when no `air.symmetric` allocations are present.

## Why this matters

Every test in main today carries a 15-line `wrap_bytes` helper that builds a memref descriptor by hand from a runtime `!llvm.ptr` — leaking the LLVM-struct ABI into the test IR. Phase 4 replaces this with the standard `memref.alloc` op + an attribute, putting the symmetric allocation at the same dialect level as the rest of the IR.

This is the foundation for phases 5/6 (DMA + channel ops): they can operate on clean `memref<T>` values from the user's perspective, without each test having to reconstruct memrefs from raw pointers.

## What's new

- **`mlir/include/air/Conversion/AIRSymmetricAllocToMgpuPass.h` / `.cpp`** — pass implementation (~200 LOC)
- **`mlir/include/air/Conversion/GPUPasses.td`** — `air-symmetric-alloc-to-mgpu` def
- **`mlir/test/Conversion/AIRSymmetricAllocToMgpu/symmetric_alloc.mlir`** — FileCheck unit tests with `// REQUIRES: gpu` (the pass isn't registered in non-GPU builds):
  - 1D alloc + dealloc shape (size, descriptor, cast, free)
  - 2D alloc with row-major strides in descriptor
  - Element type byte-size: f32 (4B), f64 (8B), i32 (4B)
  - Multiple symmetric allocs share one decl pair
  - Pass is a no-op for non-symmetric allocs
  - Pass is a no-op when there are zero symmetric allocs
- **`test/gpu/multi_gpu/air_alloc/cacheline.mlir`** — e2e test that wraps the producer/consumer cacheline reference in `air.rank` AND uses `memref.alloc {air.symmetric}` for the symmetric data buffer. 1:1 wrap of the air_rank/cacheline reference, with only the symmetric alloc/free swapped for the new abstraction. After lowering through `-air-rank-to-mgpu -air-symmetric-alloc-to-mgpu -air-translate-to-llvm`, functionally equivalent to handwritten/cacheline.mlir.
- **`test/gpu/multi_gpu/air_alloc/Makefile`** — self-contained, follows the same pattern as `air_rank/Makefile`.

## Test plan

- [x] FileCheck unit tests pass (6 cases above)
- [x] E2E on real 2x MI325X (rad-mi325x-1, NUM_RANKS=2): cache-line message PASS (`data[0]=100, flag=1`) — output structurally identical to `INPUT=cacheline` in handwritten/ and air_rank/, only distinguished by the `[mlir/alloc]` log tag. 3/3 stability runs.
- [x] Equivalence demonstrated: the same payload values (lane+100 for lanes 0..30, flag=1 at lane 31) and the same cache-line atomicity contract reach the consumer in all three variants (handwritten/cacheline, air_rank/cacheline, air_alloc/cacheline).
- [x] `git clang-format origin/main` applied; "Python and C/C++ Check Format" check should pass

## Followup

The non-symmetric staging allocations (verify_buf, heap_bases device copy) still go through `mgpuMemAlloc + wrap_bytes`. A future pass for non-symmetric device allocs would remove that helper too.

### DIFF
--- a/mlir/include/air/Conversion/AIRSymmetricAllocToMgpuPass.h
+++ b/mlir/include/air/Conversion/AIRSymmetricAllocToMgpuPass.h
@@ -1,0 +1,22 @@
+//===- AIRSymmetricAllocToMgpuPass.h ----------------------------*- C++ -*-===//
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===-----------------------------------------------------------------------===//
+
+#ifndef AIR_CONVERSION_AIR_SYMMETRIC_ALLOC_TO_MGPU_PASS_H
+#define AIR_CONVERSION_AIR_SYMMETRIC_ALLOC_TO_MGPU_PASS_H
+
+#include "mlir/Pass/Pass.h"
+#include <memory>
+
+namespace xilinx {
+namespace air {
+
+std::unique_ptr<mlir::Pass> createAIRSymmetricAllocToMgpuPass();
+
+} // namespace air
+} // namespace xilinx
+
+#endif // AIR_CONVERSION_AIR_SYMMETRIC_ALLOC_TO_MGPU_PASS_H
--- a/mlir/include/air/Conversion/GPUPassDetail.h
+++ b/mlir/include/air/Conversion/GPUPassDetail.h
@@ -28,6 +28,7 @@ using namespace mlir;
 #define GEN_PASS_DEF_CONVERTAIRTOROCDL
 #define GEN_PASS_DEF_CONVERTGPUKERNELOUTLINE
 #define GEN_PASS_DEF_AIRRANKTOMGPU
+#define GEN_PASS_DEF_AIRSYMMETRICALLOCTOMGPU
 #include "air/Conversion/GPUPasses.h.inc"
 
 } // namespace air
--- a/mlir/include/air/Conversion/GPUPasses.td
+++ b/mlir/include/air/Conversion/GPUPasses.td
@@ -49,6 +49,29 @@ def ConvertGPUKernelOutline : Pass<"air-gpu-outlining", "ModuleOp"> {
   let options = [];
 }
 
+def AIRSymmetricAllocToMgpu : Pass<"air-symmetric-alloc-to-mgpu", "ModuleOp"> {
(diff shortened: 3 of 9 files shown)


erwei-xilinx:
Superseded by [erwei-xilinx:multigpu-air-hierarchy](https://github.com/erwei-xilinx/mlir-air-erwei/tree/multigpu-air-hierarchy), which carries a refactored phase-4 pass dispatching on `#air.symmetric_heap` memref memory_space instead of the `{air.symmetric}` op attribute (commit 4a6cd868).

Rationale for the refactor: the `{air.symmetric}` op-attribute approach required walking the defining-op chain from each memref.load/store site to know whether the memref was symmetric-heap-backed. Encoding the property as a memory_space attribute carries it on the memref *type* — verifiers, helper passes, and `air.translate` can all answer "is this on the symmetric heap?" by inspecting the type, without a chain walk. It also lets memrefs constructed via `memref.view` / `wrap_bytes` (not just `memref.alloc`) carry the tag, which is needed for the cacheline tests where buffers are wrapped from raw runtime pointers.

Part of a larger redesign documented in `docs/MultiGPUPhase56Redesign.md` that also reshapes phases 5 and 6 (see #1580, #1581). Consolidated PR forthcoming.

PR #1588, opened 2026-05-06
weighted_rms_norm: add multi-tile herd_x support and profile mode

tonyjie:
## Summary

Adds a multi-tile execution path to the weighted RMS normalization example, plus a `--profile` mode for benchmarking.

- `build_module` gains an `herd_x` parameter (default `1`, backward-compatible). When `herd_x > 1`, the herd shape becomes `[herd_x, 1]` and rows are partitioned across tiles via affine maps. Weight is broadcast to all tiles via DMA. Function signature is unchanged: `(input, weight, output)`.
- New `--profile` CLI flag with timing and bandwidth measurement.
- Makefile follows the `eltwise_add` pattern: `AIE_TARGET ?= aie2p`, with `HERD_X` defaulting to 8 on aie2p and 1 on aie2.
  - `make run` → multi-tile (default for NPU2)
  - `make run_single_tile` → original single-tile baseline
  - `make profile` → multi-tile + timing/bandwidth (correctness covered by `make run`)
- Reuses `XRTRunner._check_outputs(min_correlation=0.99, ...)` for verification in the `compile-and-run` path (no duplicate `np.corrcoef` logic).

## Test plan

- [x] `programming_examples/weighted_rms_norm/run_makefile_peano.lit` — single-tile (`herd_x=1`): PASS on Strix (NPU2, AIE2P)
- [x] `programming_examples/weighted_rms_norm/run_makefile_peano_multi_tile.lit` — multi-tile (`herd_x=8`) + profile mode: PASS on Strix
- [x] Manually verified `--herd-x 4`: PASS

### DIFF
--- a/programming_examples/weighted_rms_norm/Makefile
+++ b/programming_examples/weighted_rms_norm/Makefile
@@ -2,6 +2,17 @@
 # SPDX-License-Identifier: MIT
 srcdir := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
+# AIE target: aie2 (NPU1) or aie2p (NPU2, default)
+AIE_TARGET ?= aie2p
+
+# Default herd_x for the multi-tile `run` target. NPU2 (aie2p) has enough
+# compute rows to fan out to 8 tiles; NPU1 (aie2) is single-tile.
+ifeq ($(AIE_TARGET),aie2p)
+  HERD_X ?= 8
+else
+  HERD_X ?= 1
+endif
+
 ifdef PEANO_INSTALL_DIR
   BUILD_DIR := build_peano
 else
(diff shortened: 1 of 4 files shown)


tonyjie:
Thanks @copilot-pull-request-reviewer for the review! Pushed `ad0006ee` addressing 4 of the 5 comments:

1. **Lock scope** — extended `FileLock("/tmp/npu.lock")` to cover the full warmup + timed loop, so the BO writes/syncs and kernel executions are now serialized along with `backend.load()`.
2. **Unused `invoker`** — dropped the assignment; `backend.load(artifact)` is now called for its side effect.
4. **Misleading "stride=0 BD" comment** — rewrote to describe the actual behavior: each tile DMAs the full weight vector, and the compiler may merge identical loads into a broadcast.
5. **Stale "+ correlation" Makefile comment** — updated to `(timing + bandwidth; correctness via \`make run\`)`.

For comment **3 (CLI defaults M=N=2048)** — we'd like to keep these as the defaults. They reflect the realistic LLAMA-3.2 prefill shape this kernel is designed for, and the LIT tests still complete in ~3 seconds on Strix (the multi-tile path partitions the work across 8 tiles). Happy to revisit if reviewers feel CI runtime is a concern; one alternative would be smaller defaults plus an explicit `--M 2048 --N 2048` in the `profile` Makefile target.

Both LIT tests still PASS on Strix (NPU2, AIE2P).

Issue #1567, opened 2026-05-01
[RFC] Migrate AIR to emit aie.logical_tile and remove placement-aware logic


erwei-xilinx:
## Status (updated)

Stage A is **complete** in PR Xilinx/mlir-air#1570. Every direct `getPhysTileOp` / `AIE::TileOp::create` call for memtiles, shim tiles, and compute tiles in the live AIR codegen path now routes through `aie.logical_tile` + mlir-aie's `SequentialPlacer`. Test posture: 368/368 `check-air-mlir`, plus `programming_examples/segment_alloc`, `data_transfer_transpose/dma`, and `cascade_reduction` validate end-to-end on real NPU2 hardware.

| Milestone | What changed | Constraint emitted | Status |
|---|---|---|---|
| M1 | `outlineAIEMemtiles` | `aie.logical_tile<MemTile>(col, ?)` | ✅ |
| M2 | `ShimTileAllocator` (objectfifo path) | `aie.logical_tile<ShimNOCTile>(?, ?)` | ✅ |
| M3 | `ShimDMAAllocator` (DMA path) | `aie.logical_tile<ShimNOCTile>(col, ?)` | ✅ |
| M4 | `outlineAIECores` | `aie.logical_tile<CoreTile>(col, row)` (plumbing only) | ✅ |
| M5 | Compute-tile constraint relaxation `(col, row)` → `(col, ?)` | — | 🟢 unblocked by mlir-aie wheels @ `52dd9bc` (#3041, #3042, #3046) |

Stage A landed conservative-by-design: most logical tiles still carry full `(col, row)` constraints, so the placer is a near-pass-through and physical placement is byte-identical to before. The architectural value is the API path: AIR no longer emits physical tiles directly; everything goes through mlir-aie's `LogicalTileOp` first. Stage B will progressively relax constraints; Stage C will eliminate the remaining coordinate reads.

## M5: upstream asks — landed

The original RFC listed "rectangular cluster constraint" as the upstream prerequisite. After implementing Stage A that turned out to be **the wrong abstraction**. Two upstream-side asks remained for compute-tile relaxation, and both have now landed in mlir-aie:

1. **Adjacency constraint** — for cross-herd cascade and shared-L1: a group of `LogicalTileOp`s that must be physically adjacent (cascade direction or `isLegalMemAffinity`).
   - Cascade: Xilinx/mlir-aie#3042
   - Shared-L1: Xilinx/mlir-aie#3046
[...]

PR #1597, opened 2026-05-08
DMA allocator re-key on TileOp + outlineAIECores batched cleanup (RFC #1567 Stage C #1)

erwei-xilinx:
## Summary

Two RFC #1567 cleanup items. The original M5 row-relaxation scope was dropped after design discussion (see below).

### Stage C #1 — DMA allocator identity re-key on TileOp

17 `foundAlloc` callsites converted from `(tile.getCol(), tile.getRow(), ...)` integer-pair identity to `(tile, ...)` pointer identity. New TileOp-keyed overloads on `allocation_info_t`:
- `foundAlloc(AIE::TileOp tile)`
- `foundAlloc(AIE::TileOp tile, air::MemcpyInterface memcpyOp)`
- `foundAlloc(AIE::TileOp tile, air::ChannelOp channel_op)`
- `foundPacketFlowAllocInTile(AIE::TileOp tile)`

The `(col, row)` integer overloads stay (used by the `ShimDMAAllocator` column-search path and four DMAAllocator API entry points — separate cleanup item). Also fixes a typo (`"dma_packet"` → `"npu_dma_packet"`) in one of the new overloads.

### outlineAIECores cleanup — batched createTilesViaPlacer

Replaces per-cell `air::createTileViaPlacer` calls with a single batched `air::createTilesViaPlacer` invocation per herd. Functionally identical to per-cell calls when hints are fully constrained — placer is a pass-through, IR is byte-identical — but consistent with the memtile/shim allocators in stages A/B that already use the batched API.

## On the dropped M5 row relaxation

The original PR scope included relaxing compute-tile row constraints to `(col, ?)` and letting the SequentialPlacer pick the row. **Dropped after design discussion**: the herd's `x_loc`/`y_loc` attributes carry deliberate placement intent set by `air-place-herds` (or upstream callers). Letting the placer override them:

- Loses placement determinism across releases.
- Risks correctness issues. Concrete failure case found: multi-herd-per-segment configurations that share L1 buffers across herds rely on `getPhysTileOp` deduplication at the same physical tile. The shared `aie.buffer` is materialized AFTER `outlineAIECores` runs, so the placer's buffer-adjacency mechanism can't see the cross-herd constraint at placement time. With `(col, ?)` hints, the second herd's placer would pick a different row, breaking the buffer sharing.

For compute tiles, AIR keeps the legacy flow that explicitly sets physical coordinates. RFC tracking issue updated.

## Dependencies

Xilinx/mlir-aie#3060 (placer swap-based search fix) is no longer required by this PR. It remains a real bug fix in mlir-aie's `SequentialPlacer` (partial-constraint hints can produce non-contiguous row placement) but is independent — leaving it open as standalone improvement.

## Test plan

- [x] `check-air-mlir` 384/384 (2 pre-existing AIRToROCDL failures unrelated)
- [x] CI on this PR
- [x] clang-format-17 clean

### DIFF
--- a/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
+++ b/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
@@ -108,8 +108,16 @@ struct allocation_info_t {
   bool foundAlloc(int32_t col, int32_t row, AIE::DMAChannel channel);
   bool foundAlloc(int32_t col, int32_t row);
   bool foundAlloc(int32_t col, int32_t row, air::ChannelOp channel_op);
-  bool foundAlloc(AIE::TileOp tile, AIE::DMAChannel channel);
   bool foundPacketFlowAllocInTile(int32_t col, int32_t row);
+
+  // TileOp-keyed overloads (RFC #1567 Stage C #1). Identify allocations by
+  // their owning AIE::TileOp pointer rather than by (col, row) coordinates so
+  // bookkeeping does not depend on physical placement.
+  bool foundAlloc(AIE::TileOp tile);
+  bool foundAlloc(AIE::TileOp tile, air::MemcpyInterface memcpyOp);
+  bool foundAlloc(AIE::TileOp tile, air::ChannelOp channel_op);
+  bool foundAlloc(AIE::TileOp tile, AIE::DMAChannel channel);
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Consolidated into #1601 (RFC #1567 Stage C #1) — see PR_GROUPS.md.

PR #1598, opened 2026-05-08
Re-key DMAAllocator API entry points on AIE::TileOp (RFC #1567 Stage C #1)

erwei-xilinx:
## Summary

Finishes the Stage C #1 cleanup started in #1597. The previous PR re-keyed the 17 identity-hash `foundAlloc` callsites; this PR re-keys the four `DMAAllocator` API entry points themselves so callers no longer need to extract `(col, row)` integers from a `TileOp` they already have.

### API changes

```cpp
// Before
DMAAllocator::lookupDMAAllocation(int64_t col, int64_t row, MemcpyInterface&);
DMAAllocator::getLockForDMA(MemcpyInterface&, int col, int row, ...);
TileDMAAllocator::simpleDmaChannelAlloc(MemcpyInterface&, int col, int row, int chan);
TileDMAAllocator::getBuffer(uint64_t, int64_t col, int64_t row, MemcpyInterface&);
ShimDMAAllocator::getBuffer(uint64_t&, int64_t col, int64_t row, MemcpyInterface&);
MemTileDMAAllocator::getBuffer(uint64_t, int64_t col, int64_t row, MemcpyInterface&);
CascadeAllocator::getBuffer(uint64_t, int64_t col, int64_t row, MemcpyInterface&);

// After
DMAAllocator::lookupDMAAllocation(AIE::TileOp tile, MemcpyInterface&);
DMAAllocator::getLockForDMA(MemcpyInterface&, AIE::TileOp tile, ...);
TileDMAAllocator::simpleDmaChannelAlloc(MemcpyInterface&, AIE::TileOp tile, int chan);
TileDMAAllocator::getBuffer(uint64_t, AIE::TileOp tile, MemcpyInterface&);
ShimDMAAllocator::getBuffer(uint64_t&, AIE::TileOp tile, MemcpyInterface&);
MemTileDMAAllocator::getBuffer(uint64_t, AIE::TileOp tile, MemcpyInterface&);  // tile unused
CascadeAllocator::getBuffer(uint64_t, AIE::TileOp tile, MemcpyInterface&);     // tile unused
```

[...]

PR #1601, opened 2026-05-08
Re-key DMA allocator on AIE::TileOp (RFC #1567 Stage C #1)

erwei-xilinx:
## Summary

Consolidates RFC #1567 Stage C #1 work into a single PR (was split across closed PRs #1597 and #1598, plus this one).

Reworks `allocation_info_t.foundAlloc` and the `DMAAllocator` API to identify allocations by `AIE::TileOp` pointer instead of `(col, row)` integer pairs. Pure mechanical refactor + dead code deletion. **No behavior change.**

## Commits (4)

1. **Re-key DMA allocator identity lookups on AIE::TileOp** — 17 callsites converted from `foundAlloc(tile.getCol(), tile.getRow(), ...)` to `foundAlloc(tile, ...)`. Adds 4 new TileOp-keyed overloads.
2. **Batch outlineAIECores via createTilesViaPlacer** — Uses the batched placer API in `outlineAIECores`. No behavior change.
3. **Re-key DMAAllocator API entry points on AIE::TileOp** — 4 API entry points converted: `lookupDMAAllocation`, `getLockForDMA`, `simpleDmaChannelAlloc`, `getBuffer × 4 allocators`. `generateDmaBdProgram` and `allocateCoreLocksPerMemcpyOp` no longer take `int x, int y`. Deletes 3 dead `(col, row)` overloads.
4. **Convert ShimDMAAllocator column-search to foundAllocInColumn** — Last 2 holdouts: ShimDMA's column-search path that picks a shim column before any TileOp exists. The `(col, row=0)` was a meaningless implementation artifact (shim is always row 0). Renamed to `foundAllocInColumn(col)` to make column-keyed semantics explicit. Deletes the last `(col, row)` integer overloads.

## End state

The API is uniformly TileOp-keyed for identity hashes, with column-keyed search overloads explicitly named for the one search-before-tile use case (ShimDMA's column-pick path).

## Diff

+184 / −151 across 3 files (`AIRToAIESchedulingUtils.h`, `AIRToAIESchedulingUtils.cpp`, `AIRToAIEPass.cpp`).

## Test plan

- [x] `check-air-mlir` 384/384 (2 pre-existing AIRToROCDL failures unrelated)
- [x] CI on this PR
- [x] clang-format-17 clean

### DIFF
--- a/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
+++ b/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
@@ -108,8 +108,16 @@ struct allocation_info_t {
   bool foundAlloc(int32_t col, int32_t row, AIE::DMAChannel channel);
   bool foundAlloc(int32_t col, int32_t row);
   bool foundAlloc(int32_t col, int32_t row, air::ChannelOp channel_op);
-  bool foundAlloc(AIE::TileOp tile, AIE::DMAChannel channel);
   bool foundPacketFlowAllocInTile(int32_t col, int32_t row);
+
+  // TileOp-keyed overloads (RFC #1567 Stage C #1). Identify allocations by
+  // their owning AIE::TileOp pointer rather than by (col, row) coordinates so
+  // bookkeeping does not depend on physical placement.
+  bool foundAlloc(AIE::TileOp tile);
+  bool foundAlloc(AIE::TileOp tile, air::MemcpyInterface memcpyOp);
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Folded into #1599. The C #1 work (TileOp-keyed allocator API) was API tidying without a real RFC #1567 deletion — combining with C #3 keeps it as a single review unit instead of a standalone PR that doesn't pull weight.

PR #1614, opened 2026-05-17
Enable Windows Wheels for Python 3.13 and 3.14

thomthehound:
### Summary

This expands the matrix in the CI/CD workflow to build and distribute Windows wheels for Python 3.13 and 3.14, enabled by unpinning the `numpy` version in the `requirements.txt` file. The package metadata used for the wheel already depends on plain `numpy`, and the Linux wheel jobs for Python 3.13/3.14 build fine with that policy. Pip should automatically resolve the NumPy version for the selected Python interpreters.

Also, while testing, I noticed that the bundled tools were not executable from the command line (at least on Windows) as it appears they were intended to be. So I added a Python wrapper to get those working as well.

### Changes

- Add Windows wheel builds for Python 3.13 and 3.14, matching the Linux wheel matrix.
- Add pip retry/timeout policy to the workflow to prevent the expanded matrix jobs failing due to network errors.
- Remove stale `numpy<2.0` build-time pin from `utils/requirements.txt`.
- Add `air.tools` wrappers for bundled native tools.
- Route `air-opt`, `air-translate`, `aircc`, and `air-runner` console entry points through the bundled tools.

### Testing

Locally tested Windows wheel builds with Python 3.13 for both RTTI modes.

Both wheels built successfully, installed into a clean virtual environment, exposed the expected console entry points, and passed a basic AIR Python import/parse test.

### DIFF
--- a/.github/workflows/buildAIRWheels.yml
+++ b/.github/workflows/buildAIRWheels.yml
@@ -27,6 +27,13 @@ concurrency:
   group: ci-build-air-wheels-${{ github.event.number || github.sha }}
   cancel-in-progress: true
 
+# Keep wheel matrix jobs from failing on transient package-download issues.
+env:
+  PIP_RETRIES: "3"
+  PIP_RESUME_RETRIES: "3"
+  PIP_TIMEOUT: "60"
+  PIP_DISABLE_PIP_VERSION_CHECK: "1"
+
 jobs:
   check-new-commits:
     name: Check for new commits since last release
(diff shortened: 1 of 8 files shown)


erwei-xilinx (approved):
Thanks for putting this together, @thomthehound! The Windows wheel matrix expansion and the numpy unpin are nice cleanups on their own, but I especially appreciate catching the air.tools wrapper gap. Routing all four entry points (air-opt, air-translate, aircc, air-runner) through the bundled tools wrapper is the right fix.

LGTM from my side once CI is happy.

thomthehound:
> Thanks for putting this together, @thomthehound! The Windows wheel matrix expansion and the numpy unpin are nice cleanups on their own, but I especially appreciate catching the air.tools wrapper gap. Routing all four entry points (air-opt, air-translate, aircc, air-runner) through the bundled tools wrapper is the right fix.
> 
> LGTM from my side once CI is happy.

It was absolutely my pleasure, Erwei! And thanks again for your catch involving the Windows wheels over at `mlir-aie`.

If you'd like me to help on the Windows side with anything else you have going on, please feel free to ask.

thomthehound:
I just looked at the CI failures, and I don't think they have anything at all to do with my patch. They look like something to do with the latest `llvm-aie` nightly. 

I'd look into it more, but I'm kind of busy right now with another task (and the issue, if true, is out of scope for this PR anyway).

erwei-xilinx:
> I just looked at the CI failures, and I don't think they have anything at all to do with my patch. They look like something to do with the latest `llvm-aie` nightly.
> 
> I'd look into it more, but I'm kind of busy right now with another task (and the issue, if true, is out of scope for this PR anyway).

Yes indeed, it seems to be a regression in the recent llvm-aie commits. Let me investigate. I think I have some ideas.

erwei-xilinx:
Confirmed that the failing CI tests are irrelevant from the PR itself, as I verified locally using older llvm-aie version. Landing.

thomthehound:
I really do have the most amazing CI luck. At this point, it's just getting funny to me.

PR #1615, opened 2026-05-18
[wheels] add air.tools shim so air-opt/air-translate entry points work

erwei-xilinx:
## Summary
- The wheel's [pyproject.toml](utils/mlir_air_wheels/pyproject.toml#L24-L27) declares `air-opt = "air.tools:air_opt"` and `air-translate = "air.tools:air_translate"` as `[project.scripts]` entry points, but the `air.tools` module was never created. After `pip install mlir-air` running `air-opt` fails immediately with:
  ```
  ModuleNotFoundError: No module named 'air.tools'
  ```
- Add `python/air/tools/__init__.py`, a small shim that `os.execl`s the matching native binary out of the wheel's bundled `bin/` directory (same pattern as `aie.tools` in mlir-aie).
- Register the file in `python/CMakeLists.txt` via `declare_mlir_python_sources(AirPythonSources.Tools …)` so `cmake --install` copies it into the installed `air/` package, where the wheel picks it up.

`aircc` was already wired to `air.compiler.aircc.main:main` (a real Python entrypoint) so was unaffected.

## Test plan
- [x] `cmake --build . --target install` installs `air/tools/__init__.py` alongside the rest of the `air` package.
- [x] `PYTHONPATH=install/python python3 -c "import sys; sys.argv=['air-opt','--version']; from air.tools import air_opt; air_opt()"` exec's the native binary and prints LLVM version info.
- [ ] Build a wheel via the `buildAIRWheels.yml` workflow and confirm `pip install` + `air-opt --version` succeeds end-to-end.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -38,6 +38,14 @@ if(AIE_ENABLE_BINDINGS_PYTHON)
     ADD_TO_PARENT AirPythonSources
   )
 
+  # Console-script shims that exec the native air-* binaries in the wheel.
+  declare_mlir_python_sources(AirPythonSources.Tools
+    ADD_TO_PARENT AirPythonSources
+    ROOT_DIR "${AIR_PYTHON_ROOT_DIR}"
+    SOURCES
+      tools/__init__.py
+  )
+
   declare_mlir_dialect_python_bindings(
     ADD_TO_PARENT AirPythonSources.Dialects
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Closing as a duplicate of #1614, which landed one day earlier and fixes the same `air.tools` issue (plus several other improvements: Windows wheel builds for Python 3.13/3.14, numpy unpin, `air-runner` entry point). Deferring to #1614.

PR #1625, opened 2026-05-28
[AIRToAIE] Distribute mem_bank across extern-kernel call args

erwei-xilinx:
## Summary

After `allocL1Buffers`, walk each `aie.core`'s `func.call` ops whose callee carries `link_with` (extern kernel) and greedy-assign `mem_bank` attributes on the call's memref args so that arguments of the same call sit on distinct AIE2/AIE2P L1 banks. Without this, the downstream `aie-assign-buffer-addresses` pass can place A and B of a matvec call in the same bank, causing AIE2P bank arbitration to serialize the kernel's parallel vector loads and (combined with the Peano "No memory bank assigned" code path) stall the kernel at runtime.

Arguments already pinned by an upstream pass are preserved. Buffers reused across calls keep their first assignment.

## Test plan

- [ ] `ninja check-air-mlir` (full lit suite) passes.
- [ ] On a matvec call where A and B previously shared a bank, after the pass A/B/partial each sit on distinct banks (verified against the handwritten reference bank assignment in `aie_la_2tile.mlir.prj`).

### DIFF
--- a/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
+++ b/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
@@ -61,6 +61,11 @@ generateBufferNameInStringStream(StringRef prefix, uint64_t &BufferId,
                                  mlir::StringAttr attr = nullptr, int x = -1,
                                  int y = -1);
 
+// Walks a Value back to its underlying AIE::BufferOp through chains of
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Closing as the fix is too specific to func.call kernel.

PR #1626, opened 2026-05-28
[Transform] Fix air-ping-pong-transform nested-loop SSA dominance failure

erwei-xilinx:
## Summary

`ConstructPingPongDependencyPattern`'s classification walk descended into nested `scf.for` bodies via the default post-order walk. When greedy pattern application had already pp-transformed an inner `scf.for` first, the inner-region ops carried their own producer/consumer/async_front/async_back annotations. The outer pass then picked up those inner ops and wired the outer-scope sink to an inner-region SSA value, producing:

```
error: operand #0 does not dominate this use
```

Fix: walk in pre-order with `WalkResult::skip` to prune nested `scf.for` bodies before they are visited. The inner `scf.for` op itself is still classified (it may legitimately be a producer/consumer of the outer-level buffer), but the walk does not descend into its body.

## Test plan

- [ ] Existing pp lit tests pass unchanged.
- [ ] A matvec-like design with nested `scf.for` + alloc children on both levels compiles cleanly with pp enabled (previously hit the dominance error).

### DIFF
--- a/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
+++ b/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
@@ -1216,47 +1216,56 @@ struct ConstructPingPongDependencyPattern
     SmallVector<Operation *> pong_consumer_fronts;
     SmallVector<Operation *> pong_consumer_backs;
 
-    new_loop_op.getBody()->walk([&](Operation *op) {
-      if (op->hasAttr("ping_pong") || op->hasAttr("unrolled_iteration")) {
-        auto ping_pong_id =
-            op->hasAttr("ping_pong")
-                ? (op->getAttrOfType<IntegerAttr>("ping_pong").getUInt())
-                : (op->getAttrOfType<IntegerAttr>("unrolled_iteration")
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Addressed the Copilot review on 1d11ad245 in the amended commit (6f1a0a69f):

1. **`async_front`/`async_back` dropped on inner `scf.for`** — fixed by extracting all eight bucket classifications into a single `classifyOp` lambda, called for every op including the nested `scf.for` before `skip()`. The four families are now treated symmetrically.

2. **Redundant `op != new_loop_op.getOperation()` guard** — removed. `Block::walk` never invokes the callback on the parent op, so any `scf.for` reached by the walk is by definition nested.

3. **No regression test** — added `nested_ping_pong` in `construct_ping_pong.mlir`. Outer `scf.for {unroll=2}` with hoisted L1 ping/pong allocs, body containing inner `scf.for {unroll=2}` with its own L2 ping/pong allocs. Verified the test fails on the pre-fix code (`operand #0 does not dominate this use`, exit 1) and passes FileCheck with the fix.

PR #1630, opened 2026-05-28
[AIRRtToNpu] Recognize stride-0 outer replay dims as still-contiguous

erwei-xilinx:
## Summary

**⚠️ Stacked on #1627** — base contains that PR's commit. Rebase to `main` after #1627 lands.

Extends `isContiguousAirrtDma` (added in #1627) to accept `*prevStride == 0` on outer dims. A stride-0 outer dim is a pure replay (broadcast-like transfers e.g. shim → compute broadcasts of a B vector replayed each outer iter) — the underlying inner block is still contiguous and qualifies for linear-mode BD via the wide `buffer_length` register.

Without this, replay patterns like `sizes=[N, K] strides=[0, 1]` with K > 1023 still get chunked by the wrap-limit splitter even though the data layout is contiguous.

## Test plan

- [ ] Extend the lit test added in #1627 with a stride-0 outer-dim case to lock the behavior in.
- [ ] A broadcast `air.channel.put` with `sizes=[N, K] strides=[0, 1]` produces a linear-mode shim BD instead of getting chunked.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -1181,7 +1181,11 @@ bool isContiguousAirrtDma(airrt::DmaMemcpyNdOp dma) {
     return false;
   // Each outer stride must equal the product of all inner sizes (skipping
   // size-1 dummy dims, whose stride is irrelevant since they are never
-  // stepped).
+  // stepped). An outer dim with stride=0 is a pure replay (broadcast-like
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Superseded by #1627. The merged version of #1627 (commit 2b3b277f2) already promoted the helper to `air::isContiguousRowMajorOrRepeated` in `mlir/lib/Util/Util.cpp` and accepts stride==0 outer dims (the shim BD repeat pattern). This PR would re-introduce a file-local duplicate of that logic. Closing as obsolete.

PR #1645, opened 2026-06-03
[AIRToAIE] L2 round-robin: reset memtile counter per bucket-shape group

erwei-xilinx:
## Summary
- L2MemrefToMemTileMap's saturated-fallback round-robin uses a single monotonic memtile_id counter across all buckets. When a workload has multiple L2 operands (A/B/C in a matmul), each becomes a contiguous group of splits and the counter walks past 0 — leaving subsequent groups' bucket→memtile mapping shifted by the prior group's remainder.
- For Triton-XDNA 8x4 i8 matmul: 8 A-splits land on memtiles 0..7, 4 B-splits on 0..3 (counter ends at 4), then 8 C-out splits start at memtile_id=4 and wrap → memtiles[4..7,0..3]. C-out cores at col i should write intra-col to mem_tile_i, but instead write to mem_tile_(i+4)%8 → 32 cross-col flows of distance 4. Pathfinder spends 150+s untangling.
- Fix: reset memtile_id when the bucket's alloc memref type changes. Each distinct shape is a new operand class; within a class, bucket-i corresponds to consumer/producer col i. Grouping by channel-symbol doesn't work because `air-split-l2-memref` gives each split its own unique symbol post-split.

## Repro and timing on NPU2 (npu2_8col)

| Workload | Before | After |
|---|---|---|
| `programming_examples/matrix_multiplication/int4_awq` M=N=K=2048, 8x4 herd | 2m24s | **0.95s** (150x) |
| Triton-XDNA bare_matmul_i8 (saved IR) | 30s+ timeout | **1.6s** |
| `flash_attention/dataflow_based` | unchanged (single-shape, no shift) | unchanged |
| `matrix_multiplication/i8` (4x4) | 7s | 7s |

## Test plan
- [x] `ninja check-air-mlir` — 64/64 AIRToAIE lit tests pass
- [ ] CI hardware tests (amdhx370/amd8845hs) — verify int4_awq + Triton-XDNA workloads no longer hit pathfinder timeout

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -2182,8 +2182,28 @@ void L2MemrefToMemTileMap(
     }
 
   if (saturated) {
+    // Round-robin per-alloc-shape: reset memtile_id when the bucket's
+    // alloc type changes. Each shape corresponds to a logical L2 buffer
+    // class (A, B, C in a matmul; one shape per operand). Within a class,
+    // bucket-i targets consumer/producer col-i, so resetting at the start
+    // of each class makes bucket[i]->memtiles[i] = col i (intra-col).
+    // Without the reset, a global counter shifts the second/third class's
+    // mapping by the prior class's remainder (Triton i8 matmul: B fills 4
+    // slots, leaving memtile_id at 4, so C[0] lands on memtiles[4] -> the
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Pushed e4c5b7c9 to address Copilot's review on bucket-shape reset — see commit message and the new singleton_spread test case.

PR #1658, opened 2026-06-04
[Transform] AIRSpecializeChannelWrapAndStrideInScfFor: allow fold-then-canonicalize when adjacent dims collapse

erwei-xilinx:
## Summary

The pre-fold gate in `AIRSpecializeChannelWrapAndStrideInScfFor` rejected loops whose channel op already had `maxNumDims` active wrap dims, even when the new outer dim added by folding could be collapsed back with an existing adjacent dim by `canonicalizeWrapAndStrideList` (the `stride[i-1] == size[i] * stride[i]` rule).

Concretely, an scf.for `%xb = 0..4` around a channel op with the 4D wrap `<2,6272>·<98,8>·<8,784>·<8,1>` (12544 B) and `offset[%xb]` at stride 12544 should fold: the prepended `<4,12544>` collapses with the inner `<2,6272>` into `<8,6272>`, producing a final 4D wrap `<8,6272>·<98,8>·<8,784>·<8,1>` (50176 B) that fits `maxNumDims=4`. The pre-gate prevented this and the unroll fallback emitted 4 separate BDs instead of 1.

This PR:
- Relaxes the pre-gate from `> maxNumDims - 1` to `> maxNumDims`.
- Adds a post-canonicalize legality check that returns failure if the dim count still exceeds `maxNumDims`, so we never emit illegal >`maxNumDims` BD wraps when canonicalize cannot reduce.

## Test plan

- [x] New positive test `@postfold_collapse_4d_to_4d` in `opt_memtile_dma_bds.mlir`: scf.for that folds + collapses to a single 4D BD (catches a regression of the pre-gate relaxation).
- [x] New negative test `@postfold_no_collapse_rejected`: scf.for whose folded form cannot collapse; verifies the unroll fallback produces N separate puts (catches a regression of the post-fold check — without it an illegal 5D wrap would leak through).
- [x] Existing `@func0` CHECK lines updated where the new fold now produces a strictly better single BD instead of an unrolled chain.
- [x] Full `mlir/test/Transform/` lit suite passes (210/210).
- [x] Independently verified the new positive test fails if only the gate is reverted, and the new negative test fails if only the post-fold check is removed.


erwei-xilinx:
Closing as superseded.

The two pieces of this PR have both landed in canonical form via downstream PRs:

- **Post-canonicalize legality check** (`activeDims > maxNumDims → reject`) and the two regression tests (`@postfold_collapse_4d_to_4d`, `@postfold_no_collapse_rejected`) landed in #1669 (commit `0428137e`).

- **Pre-fold gate relaxation**: the blanket `> maxNumDims - 1` → `> maxNumDims` change here was too broad — it broke `29_gemm` channel_3/10-17 and `46_triton_matmul` (the no-IV-in-offsets fold-noop case erases the surrounding scf.for and loses iteration count semantics). #1669 instead narrowed it to a `skipZeroStride + IV-at-offset[0]` discriminator. #1675 further extends the discriminator to admit IV at any `offset[i]` where `wraps[i] == 1` (size-1 dim that canonicalize will erase), which is what the conv2d_14x14 L2→L1 act gather actually needs, and pairs it with a compound-trip cascade guard on `AIRUnrollScfForIntoBDChain` to keep lock `init` semantics correct.

Branch also conflicts with main now. Closing.

PR #1674, opened 2026-06-05
[programming_examples] Add kernel registry (GEMM BF16)

tonyjie:
## Summary

Adds a documentation registry under `programming_examples/kernel_registry/` for the leaf kernels used to deploy decoder-only LLMs on NPU2 (Strix / AIE2P). It records, per kernel: which shapes have been tested, their tile configs, and measured performance + accuracy, with copy-paste reproduce commands.

The registry is built up **one verified kernel at a time** — this first contribution covers **BF16 GEMM only**. Other leaf kernels (GEMV, RMSNorm, RoPE, FlashAttention, SiLU+Mul, Eltwise Add) are listed on the roadmap and added in later PRs.

This is **documentation, not executable code** — it records results produced by the existing `programming_examples/matrix_multiplication/bf16` example on real NPU2 (RyzenAI-npu4).

## Contents

- `README.md` — what the registry is, scope (NPU2 / aie2p only), data sources, methodology, roadmap.
- `supported_kernels.md` — high-level index: which kernels are covered, tested shapes, and best measured performance.
- `details/GEMM_bf16.md` — full GEMM detail: external vs direct codegen (bit-identical, external fastest), numerical datapath (BFP16-emulated 8×8×8, FP32 accumulate), tunable parameters (recommended values + constraints), tolerances, per-path measured data across 7 shapes, and reproduce commands.
- `.gitignore` — keeps local-only working notes out of the repo.

## Notes

- All performance/accuracy numbers were measured on NPU2 (RyzenAI-npu4), full-chip 8×4 herd. The tile for each shape is the fastest found by an external-path sweep over `tile_m × tile_k_l2 × tile_n`.
- Correctness uses a full-output element-wise check against an FP32 reference (relies on the `XRTRunner` timing + precision reporting + full-coverage GEMM check landed in #1655).
- Shapes are written `M×K×N` (`C[M,N] = A[M,K] @ B[K,N]`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1678, opened 2026-06-08
[programming_examples] kernel registry: add GEMV BF16

tonyjie:
## Summary

Adds the **BF16 GEMV** kernel to the kernel registry (`programming_examples/kernel_registry/`), following the same methodology established for GEMM (#1674): a standalone harness run on real NPU2, a full-output FP32 reference, a full tile sweep for the best config, and tolerances aligned to the PyTorch/vLLM bf16 standard.

GEMV is the exact kernel llama-3.2-1B decode uses for the **Q/K/V projections** and the **LM-head**. Unlike GEMM, GEMV uses a plain **FP32 vector accumulate** (no BFP16-emulated MMA), so it is near-bit-exact: measured `mean_rel_L1 <= 2.7e-8` across all five tested shapes, several bit-identical to the FP32 reference.

## Changes

- **`matrix_vector_multiplication/bf16/matvec.py`**: wire `--perf-iters` + `report_precision`/`perf_flops` through `XRTRunner` (reuses the timing/precision support landed in #1655). Tighten `rtol` from `4e-2` to `1.6e-2` to match PyTorch's canonical bf16 tolerance (`test_addmv`). The `--perf-iters` flag defaults to 0, so existing lit-test behavior is unchanged.
- **`kernel_registry/details/GEMV_bf16.md`**: full detail page — FP32 vector-accumulate datapath, tunable parameters (`herd_m` fixed at 8; tune `tile_m`/`m_input`), tolerances vs PyTorch `test_addmv`, the 99-config tile sweep space and legality rules, per-shape results, and reproduce commands.
- **`kernel_registry/supported_kernels.md`** / **`README.md`**: add the GEMV row, tested-shapes table, and scope.
- **`kernel_registry/.gitignore`**: ignore a local-only planning doc.

## Test plan

- [x] All 5 llama-3.2-1B GEMV shapes pass on real NPU2 (RyzenAI-npu4) at their swept-best tiles, with the tightened `rtol=1.6e-2`.
- [x] Tile sweep (99 configs) completed; every legal config passes correctness, illegal ones fail at build time (clear assertion / DMA-alignment error), never silently.
- [x] Reproduce command from the detail page verified end-to-end (PASS + perf printed).
- [x] `matvec.py` change is additive (`--perf-iters` defaults to 0); existing lit test unaffected.

### DIFF
--- a/programming_examples/kernel_registry/.gitignore
+++ b/programming_examples/kernel_registry/.gitignore
@@ -2,3 +2,6 @@
 # Any file whose basename starts with internal_ is local-only.
 internal_*
 **/internal_*
+
+# Local-only planning doc — our own kernel-adding tracker, not for upstream.
+kernel_adding_todolist.md
--- a/programming_examples/kernel_registry/README.md
+++ b/programming_examples/kernel_registry/README.md
@@ -13,23 +13,24 @@ This is **documentation, not executable code** — nothing here compiles or runs
 
 ## Scope
 
-This registry targets **NPU2 (Strix / AIE2P) only** — all shapes, tile configs, tolerances, and measured numbers are for the aie2p target. It is built up **one kernel at a time** — only kernels we have independently understood and verified are included, so every number here can be trusted and reproduced. Currently the registry covers **GEMM only**.
+This registry targets **NPU2 (Strix / AIE2P) only** — all shapes, tile configs, tolerances, and measured numbers are for the aie2p target. It is built up **one kernel at a time** — only kernels we have independently understood and verified are included, so every number here can be trusted and reproduced. Currently the registry covers **GEMM** and **GEMV**.
 
 | File | What it is |
 |---|---|
 | [`supported_kernels.md`](supported_kernels.md) | High-level index: which kernels are covered, tested shapes, and best measured performance. |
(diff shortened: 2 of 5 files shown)


erwei-xilinx (approved):
Looks good to me. Will proceed to land once the Copilot reviews are addressed.

tonyjie:
Thanks for the review. Addressed the Copilot comments in f406d182:

1. **`--perf-iters` allows negative values** — fixed: now rejected with `parser.error("--perf-iters must be >= 0")` before the value reaches the runner.

2. **`herd_m` "always 8" too absolute** — fixed: the doc now qualifies the recommendation with its legality constraint (`M % (tile_m × herd_m) == 0`, `M ≥ tile_m × herd_m`). 8 is always legal for the llama-3.2-1B decode shapes (M ∈ {512, 2048, 8192, 16384}), but a model with an M not divisible by `8 × tile_m` would need a smaller `herd_m`.

3. **`n_perf_iters=0` + `perf_flops=None` mixed sentinels** — left as-is, intentionally. This matches the `XRTRunner` API contract established in #1655 and used identically by the GEMM example (`matrix_multiplication/bf16/run.py`): `n_perf_iters=0` means "no timing" and `perf_flops=None` means "no throughput line". Switching GEMV to a different convention (`None`) would make the two examples inconsistent for no functional gain. Happy to revisit if you'd prefer changing both examples + the runner together in a separate change.

PR #1679, opened 2026-06-08
[programming_examples] kernel registry: add RMSNorm BF16 (FP32 reduction)

tonyjie:
## Summary

Adds the **BF16 weighted RMSNorm** kernel to the kernel registry, and **fixes its reduction precision** to follow the GPU / HuggingFace numerical standard.

**The precision fix.** The kernel previously accumulated `sum(x²)` in **bf16**. Checked element-wise against an FP32 reference, that gives `mean_rel_L1 ≈ 6.9e-2` — about 10× worse than the bf16 standard (the earlier example only ran a loose `correlation ≥ 0.99` check, which is blind to the resulting systematic per-row scale error and passed at 0.9999). This PR accumulates the reduction in **FP32**, matching PyTorch `rms_norm_composite`, HF `LlamaRMSNorm`, and vLLM `RMSNorm` (all upcast to f32 for the whole reduction and cast back to bf16 only at the end). Result: **`mean_rel_L1` drops to `4.2e-3`** (in line with the GEMM tier), passing the canonical bf16 `rtol = 1.6e-2`. The FP32 reduction is essentially free on this memory-bound kernel (~911 µs vs ~876 µs at `herd_x=8`).

## Changes

- **`weighted_rms_norm.py`**: accumulate `sum(x²)` in an FP32 buffer (both single- and multi-tile paths); switch the correctness check from a correlation metric to a **full-output element-wise** compare; wire `--perf-iters` through `XRTRunner`; use `randn` inputs and the bf16-standard tolerance (`rtol=1.6e-2`, `atol=5e-2`). The `--perf-iters` flag defaults to 0, so existing lit-test behavior is unchanged.
- **`kernel_registry/details/RMSNorm_bf16.md`**: full detail page — FP32-reduction datapath, tunable parameters (`herd_x` fixed at 8), tolerances vs the GPU/HF standard, per-shape data, reproduce commands.
- **`kernel_registry/supported_kernels.md`** / **`README.md`**: add the RMSNorm row, tested shapes, and a methodology note (gate on element-wise checks, never correlation/cosine — matching PyTorch/vLLM practice).

> **Note: stacked on #1678** (GEMV registry). Until #1678 merges, this PR's diff also includes the GEMV commit; please review/merge #1678 first, after which this diff reduces to the RMSNorm commit alone.

## Test plan

- [x] `mean_rel_L1 = 4.2e-3` on real NPU2 (RyzenAI-npu4), passing `rtol=1.6e-2`, both single-tile and multi-tile (`herd_x=8`).
- [x] `herd_x` sweep {1,2,4,8} — near-linear scaling (7.5× at 8 columns), accuracy bit-identical across all (reduction precision, not tiling, sets it).
- [x] FP32-reduction datapath cross-checked against PyTorch `rms_norm_composite` / HF `LlamaRMSNorm` / vLLM `RMSNorm`.
- [x] `weighted_rms_norm.py` change keeps `--perf-iters` defaulting to 0; existing lit tests unaffected.

### DIFF
--- a/programming_examples/kernel_registry/README.md
+++ b/programming_examples/kernel_registry/README.md
@@ -13,25 +13,27 @@ This is **documentation, not executable code** — nothing here compiles or runs
 
 ## Scope
 
-This registry targets **NPU2 (Strix / AIE2P) only** — all shapes, tile configs, tolerances, and measured numbers are for the aie2p target. It is built up **one kernel at a time** — only kernels we have independently understood and verified are included, so every number here can be trusted and reproduced. Currently the registry covers **GEMM** and **GEMV**.
+This registry targets **NPU2 (Strix / AIE2P) only** — all shapes, tile configs, tolerances, and measured numbers are for the aie2p target. It is built up **one kernel at a time** — only kernels we have independently understood and verified against the GPU/HF numerical standard are included, so every number here can be trusted and reproduced. Currently the registry covers **GEMM**, **GEMV**, and **RMSNorm**.
 
 | File | What it is |
 |---|---|
 | [`supported_kernels.md`](supported_kernels.md) | High-level index: which kernels are covered, tested shapes, and best measured performance. |
 | [`details/GEMM_bf16.md`](details/GEMM_bf16.md) | Full detail for the BF16 GEMM kernel: numerical datapath, tunable parameters, tolerances, per-path data, reproduce commands. |
 | [`details/GEMV_bf16.md`](details/GEMV_bf16.md) | Full detail for the BF16 GEMV kernel: numerical datapath, tunable parameters, tolerances, per-shape data, reproduce commands. |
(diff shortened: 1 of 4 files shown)


tonyjie:
Thanks for the review. All three Copilot comments addressed in be9c927d:

1. **Misleading FP32-types comment** (`weighted_rms_norm.py:54`) — fixed. The comment now describes the actual datapath: FP32 reduction + FP32 scalar `rsqrt`, with the per-element `x * rstd * weight` epilogue in bf16 vectors (the aie vector unit doesn't legalize f32 vector elementwise mul), leaving the single bf16 output rounding as the only remaining quantization.

2. **`l1_out` reused as reduction scratch** — fixed. Added a dedicated `[vector_size]` bf16 L1 scratch buffer (`l1_sq`) for the square round-trip that breaks the `mulf->addf` chain, in both the single- and multi-tile paths, so the reduction and output phases no longer alias. Re-verified on NPU2: both paths PASS with identical precision (`mean_rel_L1=4.2e-3`).

3. **Mixed throughput units under one header** — fixed. Renamed the column to "Best measured throughput (NPU2, units per entry)"; each entry already carries its unit (GFLOP/s for GEMM/GEMV, GB/s for RMSNorm).

erwei-xilinx (approved):

PR #1680, opened 2026-06-09
[programming_examples] kernel registry: add FlashAttention BF16 (GQA)

tonyjie:
## Summary

Adds the BF16 FlashAttention kernel to the kernel registry, following the same methodology as GEMM / GEMV / RMSNorm: a standalone harness on real NPU2, an FP32 SDPA reference, a full-output element-wise check aligned to the GPU bf16 standard, and a tunable sweep.

- The kernel (`attn_npu2.cc` -> `attn_npu2.o`) is driven by the **heads-first** harness `attn_npu2.py`. A **seq-first** harness (`attn_npu2_seqfirst.py`) drives the same `.o` for llama-3.2-1B prefill and is numerically **bit-identical** (verified: same randn Q/K/V packed two ways, run on NPU2, de-permuted and compared -> `max abs diff = 0`, at both 512x512 and the 2048 llama shape).
- **Verified on NPU2 across a range of shapes**: head dim 64/128 (`dv_chunks` 1/2), MHA & GQA ratios (2/2 .. 64/8), seq 512/2048/16384, causal & non-causal.
- **Accuracy** `mean_rel_L1 ~= 3.9-5.5e-2` (~4x the GEMM tier): FA chains two BFP16-emulated MMAs plus a bf16 online-softmax, looser than a single matmul -- consistent with GPU FA also being looser than GEMM (PyTorch fused SDPA bf16 uses 5e-2). Accuracy is set by the datapath, not the tile config.
- **Tile config is determined, not tuned.** A tunable sweep shows the full-32-tile requirement plus the NPU2 8x4 fabric, shim-DMA channels, L1 budget and the causal block-mask collapse the config to a near-unique point (`lqp=256, num_q_tiles=4, num_heads_per_unroll=2, num_cascade_stages=4`, or the ~equivalent `512/8/1/4`).

## Changes

- `attn_npu2.py` / `attn_npu2_seqfirst.py`: switch the correctness check to randn inputs + full-output element-wise (`rtol=1.6e-2, atol=1e-1`), drop the correlation / mismatch-tolerance gate; wire `--perf-iters` through `XRTRunner`; expose the previously-hardcoded `num_q_tiles` and `num_heads_per_unroll` as `build_module` params + CLI args. Defaults unchanged, so existing lit behavior is preserved. Also fix compile-only mode passing `XRTRunner`-only opts to `XRTBackend`.
- `details/FlashAttention_bf16.md`: full detail page -- online-softmax datapath, how it differs from the GPU FP32-carry standard, parameters & where each constraint comes from, tolerances vs GPU FA, per-shape data, tile-choice analysis, reproduce commands.
- `supported_kernels.md` / `README.md`: add FlashAttention row, tested shapes, and methodology notes.

## Test plan

- [ ] Existing FA lit tests still pass (defaults unchanged).
- [ ] Reproduce the tested-shapes table via the commands in `details/FlashAttention_bf16.md`.

Generated with Claude Code

### DIFF
--- a/programming_examples/flash_attention/kernel_fusion_based/attn_npu2.py
+++ b/programming_examples/flash_attention/kernel_fusion_based/attn_npu2.py
@@ -65,6 +65,7 @@ def build_module(
     num_heads=2,
     num_kv_heads=None,
     causal=False,
+    num_heads_per_unroll=2,
 ):
     """Build flash attention module with selective Q capture pattern.
 
@@ -81,6 +82,10 @@ def build_module(
         num_kv_heads: Number of key/value heads for grouped-query attention
             (GQA). If None, defaults to num_heads (standard MHA).
         causal: Whether to enable causal (autoregressive) masking.
+        num_heads_per_unroll: Heads processed per segment instance (default: 2).
+            Acts as the physical-column multiplier — physical columns =
+            num_heads_per_unroll * num_q_tiles (must be <= 8 on NPU2). Requires
+            num_heads % num_heads_per_unroll == 0.
     """
     # Validate
     assert lq % lqp == 0, f"lq ({lq}) must be divisible by lqp ({lqp})"
@@ -115,7 +120,6 @@ def build_module(
     )
     gqa_group_size = num_heads // num_kv_heads
 
-    num_heads_per_unroll = 2
     assert num_heads % num_heads_per_unroll == 0, (
         f"num_heads ({num_heads}) must be divisible by "
         f"num_heads_per_unroll ({num_heads_per_unroll})"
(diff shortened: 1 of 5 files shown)


tonyjie:
Thanks @Copilot — all five points addressed in the latest push:

1. **FLOP formula** (`attn_npu2_seqfirst.py`, `attn_npu2.py`, and `details/FlashAttention_bf16.md`): now `2·num_heads·lq·lk·(dk+dv)` so P@V correctly scales with `dv`. (For the reported shapes `dk==dv`, so the published GFLOP/s are unchanged.)
2. **CLI validation**: added `--num-q-tiles >= 1` and `--num-heads-per-unroll >= 1` checks via `parser.error(...)` in both harnesses (clean error instead of `ZeroDivisionError`).
3. **`build_module` docstring**: documented `num_heads_per_unroll` (segment width / physical-column multiplier; requires `num_heads % num_heads_per_unroll == 0`) in both harnesses.
4. **Config completeness** (`supported_kernels.md`): the FA section now states all rows use the one full-chip config `lqp=256, num_q_tiles=4, num_heads_per_unroll=2, num_cascade_stages=4` explicitly.
5. Covered by (1) — the doc FLOPs expression is updated too.

erwei-xilinx (approved):

PR #1681, opened 2026-06-11
[programming_examples] kernel registry: add Element-wise Add (BF16)

tonyjie:
## Summary

Adds the BF16 element-wise add kernel (`c = a + b`) to the kernel registry, following the same methodology as GEMM / GEMV / RMSNorm / FlashAttention: a standalone harness on real NPU2, an FP32 reference, a full-output element-wise check, and a tunable sweep.

- **The simplest leaf kernel** — pure per-element add, no reduction or non-linearity, so the only quantization is the single bf16 rounding of each output. Measured `mean_rel_L1 = 1.9e-3`, the cleanest tier in the registry (below GEMM ~9.3e-3 and RMSNorm ~4.2e-3).
- **Numerically equivalent to `torch.add`**: PyTorch's bf16 element-wise add (via TensorIterator) is an f32 sum of two bf16 operands rounded once to bf16, with no cross-element accumulation — the NPU kernel does the same. (The residual add in a HF/vLLM decoder is exactly `residual + hidden_states`.)
- **Memory-bound; tunable space fully swept.** `herd_y > 1` cannot place (the 3-DMA-per-tile shim-channel demand runs out of channels), so the herd is capped at one 8-column row — the kernel cannot fill the 32-tile array. Within `herd_y=1`, `herd_x` scales near-linearly to **57.7 GB/s** at `herd_x=8` (the highest bandwidth in the registry — pure streaming). `tile_n` has a small monotonic effect (~4%, saturating at 2048). Best config `herd_x=8, herd_y=1, tile_n=2048` for every shape.
- llama-3.2-1B's prefill residual add is the fused `o_ffn` inline 2-D variant (same `c=a+b` math); this entry measures the reusable standalone `eltwise_add`.

## Changes

- `eltwise_add.py`: switch the correctness check to randn inputs + full-output element-wise (`rtol=1.6e-2, atol=5e-2`) instead of the 100-sample stochastic check; wire `--perf-iters` + `report_precision`. Defaults unchanged, so existing lit behavior is preserved.
- `details/EltwiseAdd_bf16.md`: full detail page — datapath, accuracy (with the `torch.add` equivalence), parameters & constraints, tunable-space sweep, tolerances, per-shape data, reproduce commands.
- `supported_kernels.md` / `README.md`: add the row, tested shapes, scope.

## Test plan

- [ ] Existing eltwise_add lit tests still pass (defaults unchanged).
- [ ] Reproduce the tested-shapes table via the commands in `details/EltwiseAdd_bf16.md`.

### DIFF
--- a/programming_examples/eltwise_add/eltwise_add.py
+++ b/programming_examples/eltwise_add/eltwise_add.py
@@ -258,7 +258,18 @@ if __name__ == "__main__":
         dest="output_format",
         help="Output format for the compiled binary (default: xclbin)",
     )
+    parser.add_argument(
+        "--perf-iters",
+        type=int,
+        default=0,
+        dest="perf_iters",
+        help="If >0, time the kernel over this many iters (after warmup) and "
+        "print Latency + bandwidth in addition to the correctness check",
+    )
     args = parser.parse_args()
+
+    if args.perf_iters < 0:
+        parser.error("--perf-iters must be >= 0")
 
     if args.dtype == "bf16":
         INPUT_DATATYPE = bfloat16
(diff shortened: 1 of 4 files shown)


tonyjie:
Thanks @Copilot — addressed in the latest push:

1. **float64 intermediate**: inputs are now generated directly as `float32` (`rng.standard_normal(n, dtype=np.float32)`), avoiding the large f64 array + conversion for big `n`.
3. **"FP32 reference" wording**: reworded to make clear it's an FP32-compute reference *rounded to the output dtype* (a bf16-rounded reference when `dtype=bf16`).
2. **non-bf16 atol**: `--dtype` only accepts `bf16 | f32` (no fp16 path), so the `else` branch is always f32, where add is effectively exact and `rtol=1e-3 / atol=1e-5` is fine; added a comment documenting the two branches. Happy to switch to `np.finfo`-derived tolerances if an fp16 path is added later.

Regression after the change: `mean_rel_L1 = 1.88e-3`, PASS (unchanged).

erwei-xilinx (approved):

PR #1682, opened 2026-06-11
[programming_examples] kernel registry: add SiLU-and-Mul (BF16) + isolate kernel

tonyjie:
## Summary

Adds the **SiLU-and-Mul (SwiGLU activation)** BF16 kernel to the kernel registry, and promotes its harness to a standalone first-class example. Two commits:

1. **Isolate `silu_and_mul` kernel from LLM model code** — `git mv` the kernel out of `llms/llama_kernel_builder/ffn_swiglu/` to a standalone `programming_examples/silu_and_mul/` (kernel `.py` + `.cc` + `Makefile` + NPU2 LIT test), consistent with the other kernel example dirs (`eltwise_add`, `weighted_rms_norm`). The FFN integration harness (`run.py`) stays under `ffn_swiglu/` since it depends on the LLM builder ecosystem; the 3 LLM importers and `external_kernels.py` are updated to the new path.
2. **kernel registry: add SiLU-and-Mul (BF16)** — detail page, `supported_kernels.md` entry, README scope/index/roadmap, and a new "sweep only the configs that actually place" methodology note.

## Kernel facts (measured on real NPU2)

- `out = SiLU(gate) * up`, per-element BF16; SwiGLU activation of llama-3.2-1B prefill FFN.
- Best config `herd_x=8, herd_y=1, tile_n=4096` → **25.1 GB/s** at `N=16777216` (memory-bound).
- `mean_rel_L1 = 1.02e-2` — the "bf16 + one transcendental" tier (hardware `aie::tanh` LUT + chained bf16 roundings). Verified element-wise over the full FP32 reference at `rtol=1.6e-2 / atol=8e-2`.

## Test plan

- [x] Standalone kernel passes on NPU2 via `make run` (`mean_rel_L1=1.02e-2`) and `make profile`
- [x] NPU2 LIT test passes (`lit ... run_makefile_peano_npu2.lit` → PASS)
- [x] All three LLM FFN builders (bf16 / int4 / bfp16 `o_ffn_*`) still build the SwiGLU stage after the move
- [x] Registry detail page paths / line refs verified accurate


tonyjie:
Thanks for the review. Responses to the Copilot comments:

**Fixed**
- `details/SiLU_Mul_bf16.md:23` (builder default mismatch) — corrected the `build_module` signature to `herd_x=8, herd_y=None->1` to match the actual default. (commit `fec4afa7`)

**Intentional / no change (lit + Makefile `cd`/env)**
The lit `cd test_peano_npu2` line and the `PEANO_INSTALL_DIR=... cd $(BUILD_DIR) && python3 ...` recipe pattern are copied verbatim from the existing registry sibling examples (`eltwise_add`, `weighted_rms_norm`) to stay consistent with the established convention. The observation that `cd` doesn't persist across separate lit RUN lines is technically correct, but it doesn't affect correctness here: the `make` invocations use the absolute `%S/Makefile` with a self-computed `srcdir` and `BUILD_DIR`, so the actual working directory is controlled by the Makefile, not by the lit `cd`. The recipe's `VAR=val cmd1 && cmd2` form runs as a single `/bin/sh` recipe line where `VAR` is visible to the chain. Both are confirmed working: the NPU2 LIT test was run locally with the real lit runner and passes (`PASS: AIR_PROGRAMMING_EXAMPLES :: silu_and_mul/run_makefile_peano_npu2.lit`), as do `make run` / `make profile`. Changing only this example would diverge it from the rest of the registry; if the maintainers prefer to fix the pattern, it should be a separate sweep across all the kernel examples.

erwei-xilinx (approved):

PR #1683, opened 2026-06-12
[programming_examples] kernel registry: add RoPE BF16 (half-split) + isolate kernel

tonyjie:
## Summary

Adds the **RoPE (rotary positional encoding, half-split)** BF16 kernel to the kernel registry, and promotes its microkernel to a standalone first-class example. Two commits:

1. **Add standalone `rope_halfsplit` kernel example** — move the half-split RoPE `.cc` out of `llms/llama_kernel_builder/` into `programming_examples/rope_halfsplit/` (builder `.py` + `.cc` + `Makefile` + NPU2 LIT test), and point llama's `external_kernels.py:compile_rope()` at the new path (`_PROJ_ROOT/rope_halfsplit/...`), matching how every other kernel `.cc` is already referenced. Single source, no duplicate. llama previously had no standalone RoPE harness (it inlines RoPE into fused launches).
2. **kernel registry: add RoPE BF16 (half-split)** — detail page, `supported_kernels.md` entry, README scope/index/roadmap/harness-list, and a variant-fidelity methodology note.

## The variant trap (why this matters)

llama-3.2-1B uses **half-split** RoPE (HuggingFace `rotate_half`), matching `apply_rotary_pos_emb`. The `programming_examples/rope_lut` and `rope_sincos` harnesses are **interleaved** (even/odd pairing) — different math (`max|half-split − interleaved| ≈ 7.8`) that merely happens to link the same overloaded `@rope` symbol. This registry entry is built around the exact `rope_halfsplit.cc` llama links, with a genuine half-split FP32 reference — not the decoys, and not the stale interleaved builder reference (which used a forbidden `min_correlation=0.99` correlation gate).

## Kernel facts (measured on real NPU2)

- Half-split rotation, `head_dim=64`, BF16. No reduction.
- Best config `herd_x=8, herd_y=1` → **43.4 GB/s** at 65536×64 (memory-bound, herd_x scales 7.4× near-linear).
- `mean_rel_L1 = 2.8e-3` — second-cleanest in the registry. Verified element-wise over the full FP32 half-split reference (matches HF `apply_rotary_pos_emb` to ~1 bf16 round) at `rtol=1.6e-2 / atol=5e-2`. **No correlation/cosine gate.**
- 6 tested shapes (decode 8/32 → prefill 65536).

## Test plan

- [x] Standalone kernel passes on NPU2 via `make run` (`mean_rel_L1=2.8e-3`) and `make profile`
- [x] NPU2 LIT test passes
- [x] `compile_rope()` builds `rope.o` from the new path; llama prefill RoPE builder (`build_rms_gemms_rope_module`) still builds after the move
- [x] Registry detail page / supported_kernels / README cross-doc consistent


tonyjie:
Thanks for the review. Responses to the Copilot comments:

**Fixed**
- `rope_halfsplit.py:67` — removed the unused `rows_per_tile` local (the DMA loop strides by `total_tiles` directly, so it was dead). (commit `1d3e6b49`)

**Intentional / no change (lit + Makefile `cd`/env pattern)**
The lit `cd test_peano_npu2` line and the `PEANO_INSTALL_DIR=... cd $(BUILD_DIR) && python3 ...` recipe pattern are copied verbatim from the existing registry sibling examples (`eltwise_add`, `weighted_rms_norm`, `silu_and_mul`) to stay consistent with the established convention — the same points were raised and addressed on the SiLU-and-Mul PR (#1682). The observation about `cd` not persisting across separate lit RUN lines is technically correct, but it doesn't affect correctness here: the `make` invocations use the absolute `%S/Makefile` with a self-computed `srcdir`/`BUILD_DIR`, so the working directory is controlled by the Makefile, not the lit `cd`; and the recipe's `VAR=val cmd1 && cmd2` form runs as a single `/bin/sh` line where `VAR` is visible to the chain. Both are confirmed working: the NPU2 LIT test runs locally with the real lit runner and passes, as do `make run` / `make profile`. Changing only this example would diverge it from the rest of the registry; if the maintainers want to fix the pattern (incl. `%t` hermetic temp dirs), it should be a single sweep across all the kernel examples.

erwei-xilinx (approved):

PR #1685, opened 2026-06-17
[CI] Gate llama32_1b_int4 verify lit behind hf_token to prevent cancellation

erwei-xilinx:
## Problem

The gated HF_TOKEN sub-shell inherits `-j8` from `LIT_OPTS` and its `--filter` matches **3 verify tests simultaneously**:
- `llama32_1b/run_npu2_verify.lit`
- `llama32_1b_int4/run_npu2_verify.lit`
- `llama32_1b_int4/run_npu2_verify_prefill_bfp16.lit`

Each loads a ~1B parameter model via PyTorch + HuggingFace transformers. The int4 variants additionally hold the full bf16 model in memory during AWQ quantisation calibration. Running all 3 concurrently exhausts RAM on the `amdhx370` runner, triggering the Linux OOM killer (exit 137). GitHub Actions reports this as the runner receiving a shutdown signal because the killed subprocess is part of the runner process tree. The `amd8845hs` runner has more RAM and is not affected.

## Fix

Add `-j1` to `LIT_OPTS` inside the gated sub-shell so the 3 LLM verify tests run sequentially rather than concurrently, avoiding the OOM condition on memory-constrained runners.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -150,17 +150,16 @@ jobs:
           ninja check-programming-examples-peano || ninja check-programming-examples-peano
 
           # Programming examples set 1b: hf_token-gated tests only.
-          # Matches verify lits under llms/ that require the hf_token feature.
-          # llama32_1b_int4/run_npu2_verify.lit is included here: the AWQ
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Closing — replaced by a cleaner approach.

PR #1694, opened 2026-06-27
[AIRRt] Re-arm cascade locks across re-dispatch for single-trip launches

atassis:
## Problem
A kernel built as a single-trip launch (`@launch sizes=[1,1]`) that uses an `npu_cascade` channel is correct on the first host dispatch but **aborts on the second** with `qds_device::wait() unexpected command state`. Re-dispatching the same kernel/BOs is the normal XRT pattern (the `test_*.cpp` harnesses loop over warmup+iters), so any single-trip cascade kernel (e.g. an M=1 decode kernel) hits this.

## Root cause
`AIRRtToNpuPass` emits the per-launch `load_pdi` reset -- which re-runs `initLocks` (core reset/unreset, re-arming cascade locks/credits) -- only when `deviceHasRepeatCountDMAs` is true. A single-trip launch has `repeat_count == 0`, so the predicate is false and the reset is never emitted; the cascade core-locks stay in their post-dispatch-1 state and the second dispatch's command never completes. Multi-trip launches (`repeat_count > 0`) get the reset, which is why they are re-entrant.

## Fix
Broaden the gate: a device also needs the reset if it uses the cascade bus. Adds `deviceHasCascade` (walks for `CascadeFlowOp` / `GetCascadeOp` / `PutCascadeOp`) and `deviceNeedsLockReset = deviceHasRepeatCountDMAs || deviceHasCascade`, used at the three reset gate sites. Reuses the existing, proven `load_pdi` / `initLocks` reset path -- no new infrastructure.

## No regression
Multi-trip kernels already satisfied `deviceHasRepeatCountDMAs` (a cascade device with repeat-count DMAs stays true -- no double reset). Single-trip cascade kernels previously aborted (could not run), so this only enables newly-working kernels; nothing that worked before changes.

## Validation (NPU2 / Strix)
A single-trip bf16 cascade FFN that aborted on dispatch #2 now completes 505 re-dispatches with no abort, output Pearson correlation 0.9998 vs golden, ~566 us warm.

## Scope
The reset machinery is on the ELF output path (`outputElf`); this fixes single-trip cascade re-entrancy there.

## Test
Added Test 8 to `mlir/test/Conversion/AIRRtToNpu/load_pdi_repeat_count.mlir`, alongside the existing `repeat_count` cases: a `npu2` device with an `aie.cascade_flow` and no `repeat_count` DMA. With `output-elf=true` it checks `load_pdi` is emitted (and the reset-device clone exists); with `output-elf=false` it checks it is not. Pure `air-opt -airrt-to-npu` + FileCheck, no hardware.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -104,12 +104,12 @@ static bool deviceHasCascade(xilinx::AIE::DeviceOp device) {
   return hasCascade;
 }
 
-// Static trip count of the affine loops enclosing `op` (1 if none). Used to tell
-// a SINGLE-TRIP launch (the launch boundary runs once per host dispatch) from a
-// MULTI-ITERATION launch (the boundary sits inside an iteration loop, e.g.
-// flash-attention's lq_iters loop). A non-constant affine loop or any scf loop
-// is treated as multi-trip. Must be evaluated before unrollAffineFors strips
-// these loops.
+// Static trip count of the affine loops enclosing `op` (1 if none). Used to
+// tell a SINGLE-TRIP launch (the launch boundary runs once per host dispatch)
+// from a MULTI-ITERATION launch (the boundary sits inside an iteration loop,
+// e.g. flash-attention's lq_iters loop). A non-constant affine loop or any scf
+// loop is treated as multi-trip. Must be evaluated before unrollAffineFors
+// strips these loops.
 static int64_t enclosingLoopTrips(mlir::Operation *op) {
   int64_t trips = 1;
   for (mlir::Operation *p = op->getParentOp(); p; p = p->getParentOp()) {
@@ -131,12 +131,12 @@ static int64_t enclosingLoopTrips(mlir::Operation *op) {
 // Cascade core locks/credits need the per-launch reset only for a SINGLE-TRIP
 // launch: across a host re-dispatch they are stale and the kernel aborts on the
 // 2nd dispatch without the reset. A MULTI-ITERATION launch re-arms them every
-// iteration on its own, so inserting the reset between iterations is unnecessary
-// and costs a load_pdi PDI reload per boundary (a silent perf regression, e.g.
-// flash-attention prefill). Detect single-trip by the air.launch_end boundary
-// not sitting inside an iteration loop. MUST be evaluated before the launch_end
-// wait_all ops are converted/erased and before unrollAffineFors -- see
-// markDevicesNeedingLockReset.
+// iteration on its own, so inserting the reset between iterations is
+// unnecessary and costs a load_pdi PDI reload per boundary (a silent perf
+// regression, e.g. flash-attention prefill). Detect single-trip by the
+// air.launch_end boundary not sitting inside an iteration loop. MUST be
+// evaluated before the launch_end wait_all ops are converted/erased and before
+// unrollAffineFors -- see markDevicesNeedingLockReset.
 static bool deviceHasSingleTripCascade(xilinx::AIE::DeviceOp device) {
   if (!deviceHasCascade(device))
     return false;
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Hi @atassis, thanks for working on this PR!

I tested on local NPU2 (Strix/AIE2P). The int4 GEMV and attention decode numbers are clean — no regression there. However, flash attention prefill (`flash_attention/kernel_fusion_based/attn_npu2.py`) shows a significant performance regression.

Measured at LQ=LK=512, NUM_HEADS=2 (the CI default parameters), 20 warm-up + 20 profiling iterations via `test_elf_npu2.exe`: 704 µs before → 1000 µs after (+42%).

Why CI passes: `run_npu2_makefile_peano_elf.lit` uses make run, which only checks PASS! (correctness). The slowdown is silent.

Root cause: Flash attention uses `npu_cascade` channels but has `repeat_count=0` on all DMA BDs (it uses while-true loops). Before this PR, `deviceHasRepeatCountDMAs` was false for it, so `createLightweightResetDevice` was never called and no reset PDI was emitted. After this PR, `deviceHasCascade` is true, so `deviceNeedsLockReset` fires and inserts `attn_seg_reset.pdi` into the ELF config. For a 2-iteration launch (lq_iters = LQ/LQP = 512/256 = 2) this reset fires between iterations, adding ~280 µs of PDI reload overhead per boundary.

Pre-patch ELF PDIs: `empty_0, attn_seg, main`

Post-patch ELF PDIs: `empty_1, empty_0, attn_seg_reset, attn_seg, main`

Could we update the guard for inserting resets accordingly?

atassis:
Thanks @erwei-xilinx, you nailed the root cause. `deviceHasCascade` was too broad. It fired the per-launch `load_pdi` reset for any cascade device, so a multi-iteration launch like flash-attention prefill got the reset between its loop iterations, where it isn't needed since the cascade locks re-arm each iteration on their own. That accounts for the +42%.

I narrowed the guard to single-trip cascade only, meaning a cascade device whose `air.launch_end` boundary isn't nested in an iteration loop. The decision has to be made before the wait_all markers are erased and before `unrollAffineFors`, so I compute it up front, cache it on the device, and read it back at both the `load_pdi` insertion and the reset-clone sites. `repeat_count` behaviour is unchanged.

For tests I added a multi-iteration cascade case to `load_pdi_repeat_count.mlir` with `CHECK-NOT` on the reset PDI, and kept the single-trip case asserting it still gets one. That also covers your point about CI only checking correctness, since this class of regression now shows up in the lit test.

I haven't re-run your attn_npu2 numbers on hardware yet. My NPU toolchain setup is still coming together, so I'll follow up with the before and after once it's ready. If you get a chance to re-run it on your NPU2 before then it would confirm directly, but no rush.

erwei-xilinx:
Hi @atassis — I pushed two commits to your branch (edits-from-maintainers) after testing on a local NPU2 (Strix/AIE2P). Summary below.

**Why:** the narrowed single-trip guard still regressed flash-attention prefill on hardware. I measured it at the CI defaults (LQ=LK=512, 2 heads, 20 warmup + 20 iters, `test_elf_npu2.exe`):

| Build | Avg | Min |
|---|---|---|
| main (before PR) | 708 µs | 702 µs |
| this PR (before my commits) | 1004 µs | 1000 µs |
| this PR + my commits | 707 µs | 701 µs |

**Root cause the loop-trip heuristic missed:** by the time `airrt-to-npu` runs, a multi-iteration launch (flash-attention's `lq_iters=2`) has already been **unrolled**, so it doesn't appear as one `air.launch_end` inside a multi-trip `affine.for` — it appears as *two* `air.launch_end` markers under a degenerate `affine.for 0 to 1`. `enclosingLoopTrips()` returns 1 for both, so `deviceHasSingleTripCascade` classified it as single-trip and still inserted the `attn_seg_reset` `load_pdi` between the iterations (2 reloads, the +42%). I confirmed this from the `pass_049` debug IR: two top-level `air.launch_end`, no enclosing multi-trip loop.

**Commit 1 — `bb43517c` (the fix):** treat `>1 air.launch_end` as multi-iteration too. Single-trip is now *exactly one* `air.launch_end` not inside a multi-trip loop. Your single-trip re-entrancy fix is preserved unchanged (Test 8 still emits the reset; `deviceHasRepeatCountDMAs` path untouched). Added Test 10, a flash-attention-shaped guard (degenerate outer loop + two unrolled `air.launch_end` + cascade) that asserts no `load_pdi`/`_reset` — I verified it **fails** on the pre-fix predicate and **passes** after, so this class of perf regression is now caught in CI.

**Commit 2 — `28136867` (Copilot review):**
- Guard `createLightweightResetDevice` on `clOutputElf` — it was cloning `*_reset` devices even in non-ELF mode where `load_pdi` is never emitted (dead IR). Verified: unguarded produced 5 clones in `output-elf=false`; guarded → 0. Added an `EMIT-FALSE-NOT` for the clone.
- Fixed the stale "repeat_count only" comment at the `load_pdi` site and the test-file header to reflect the `deviceNeedsLockReset` (repeat_count OR single-trip cascade) condition.

`ninja check-air-mlir` is 404 pass / 7 xfail on the combined stack. Happy to adjust naming/wording if you'd prefer — thanks for the original fix, the single-trip abort was a real one.

atassis:
@erwei-xilinx thanks for catching this up and finishing, would spend more time on it by myself

PR #1719, opened 2026-07-09
[AIRMiscPasses] Add air.no_split opt-out for L2 memref splitting

erwei-xilinx:
## Summary

`AIRSplitL2MemrefForBufferConstraintPass` splits an L2 buffer whenever its channel access pattern is multiple-in-single-out or single-in-multiple-out, to satisfy per-tile buffer constraints. Some hand-written aggregator patterns (e.g. a per-column output assembler that deliberately funnels many contributors into one wide L2 buffer) must **not** be split: splitting multiplies the launch-level shim endpoint count and defeats the aggregation, and in dense cases overflows the memtile block-descriptor limit (`aie.memtile_dma op has more than 48 blocks`).

- Add an opt-out: an `air.no_split` attribute on the alloc (or its enclosing `air.execute` wrapper) makes the pass leave that L2 buffer intact.
- Buffers without the attribute are unchanged.

## Test plan

- [x] New lit test `air_split_l2_memref_no_split.mlir`: a 256x256 L2 buffer whose access pattern would split into four 64x256 fragments stays a single buffer when marked `air.no_split`. Without the opt-out the pass splits it (verified FAIL-without).
- [x] `check-air-mlir`: no new failures (AIRMiscPasses suite 22 pass / 1 pre-existing xfail).

### DIFF
--- a/mlir/lib/Transform/AIRMiscPasses.cpp
+++ b/mlir/lib/Transform/AIRMiscPasses.cpp
@@ -2160,6 +2160,18 @@ AIRSplitL2MemrefForBufferConstraintPass::getTargetMemrefAllocs(
   };
 
   for (auto allocOp : allocOps) {
+    // Opt-out: a `air.no_split` attribute on the alloc (or its enclosing
+    // air.execute wrapper) tells the pass to leave this L2 buffer intact. Used
+    // by hand-written aggregator patterns (e.g. a per-column output assembler)
+    // where splitting would multiply the launch-level shim endpoint count and
+    // defeat the aggregation, overflowing the memtile BD limit.
+    bool optOut = allocOp->hasAttr("air.no_split");
+    if (!optOut)
+      if (auto exec =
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Addressed the review comments in 98668eed (+ earlier f9fb3167):

1. **Execute-wrapper lookup only checked the immediate parent** — switched the opt-out to `allocOp->getParentOfType<air::ExecuteOp>()`, so `air.no_split` on an enclosing `air.execute` is honored even when the alloc is nested in control flow inside the execute region.
2. **`CHECK-NOT: air.channel @channel_2` placed after the label** — moved it before `CHECK-LABEL` so it covers the module-scope channel decl a split would introduce (which is emitted before the `func.func`, i.e. outside the old after-label window).
3. **Execute-wrapper branch not exercised** — added `@test_exec_attr` (attribute on the `air.execute`, not the alloc) in f9fb3167; also added `@test0_split` (same pattern without the attribute) so the opt-out CHECK-NOTs can't pass vacuously.

PR #1722, opened 2026-07-10
[AIRToAIE] Honor air.memtile_col to pin an L2 buffer's memtile column

erwei-xilinx:
## Summary
Adds an `air.memtile_col` IntegerAttr on an L2 `memref.alloc` to pin the memtile column its bucket lands on in `L2MemrefToMemTileMap`, overriding the default round-robin / column-affinity assignment.

- **Hint-only**: invalid columns (no memtile in the target) warn and fall through; strict-first-wins on conflict within a bucket.
- **Pin-aware bucketing**: two L2 buffers sharing one `air.channel` but carrying distinct pins are refused a merge, so they land in different buckets and the shared channel becomes a real cross-memtile flow. This makes explicit memtile→memtile hops expressible.
- Pins are honored on **both** the column-affinity path and the saturation path (the latter now routes pinned buckets through `findOrCreateLtoForCol`).

Without this, designs that must spread L2 aggregator buffers across specific memtile columns collapse onto too few memtiles and can exhaust a memtile channel's BD-ID budget.

## Test plan
- [x] New lit test `mlir/test/Conversion/AIRToAIE/l2_memtile_column_pin.mlir`: two allocs pinned to distinct columns (7 and 5) lower to `MemTile(7, ?)` / `MemTile(5, ?)`; passes with the change, fails without (stock places them by affinity).
- [x] `ninja check-air-mlir` — no new failures (pre-existing Util/Channel env failures only).

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -2479,7 +2479,11 @@ void L2MemrefToMemTileMap(
     if (pin >= 0)
       col = pin;
     bucketCols.push_back(col);
-    if (col >= 0)
+    // Only affinity-derived cols feed the saturation trigger. A user pin is an
+    // explicit request that is honored on both branches, so it must not push
+    // the module into the saturation fallback and thereby perturb the
+    // placement of unrelated, unpinned buckets.
+    if (pin < 0 && col >= 0)
       colDemand[col]++;
   }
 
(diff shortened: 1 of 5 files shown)


erwei-xilinx:
Addressed the Copilot review comment in 556c3da6: `deriveBucketPin` (which emits the invalid/conflicting-pin warnings) is now called exactly once per bucket in the `bucketCols`/`colDemand` loop, and the saturation path reuses the cached `bucketPins[i]` instead of re-deriving. This removes the duplicate warnings under saturated placement. Lit test still passes.

PR #1724, opened 2026-07-10
[python] expose Channel buffer_resources (objectFIFO depth) knob

atassis:
The placed-API `Channel` wrapper only exposed `sym_name`/`broadcast_shape`/`size`,
so a Python dataflow design could not pin the lowered `aie.objectfifo` depth even
though the mechanism already exists at the IR level: `air.channel` carries a
`buffer_resources` attribute (default 1) that `AIRToAIEPass` uses as the
objectFIFO depth (`ChannelOp::getBufferResources`, `AIRToAIEPass.cpp`), covered by
`test/Conversion/AIRToAIE/air_channel_to_objectfifo_buffer_resources.mlir`.

This adds `buffer_resources: Optional[int]` to `Channel.__init__`, mirroring the
existing `broadcast_shape` attribute handling, plus a python-binding lit test. It
lets dataflow pin a deeper FIFO (e.g. depth 2 for producer/consumer overlap)
without hand-editing the lowered IR.

## Changes
- `python/air/dialects/_air_ops_ext.py`: accept an optional `buffer_resources`
  argument on `Channel.__init__` and emit it as an i64 attribute when set.
- `python/test/dialect/channel_buffer_resources.py`: new lit test asserting the
  default (no attribute) and explicit depth-2 / depth-4 cases.


erwei-xilinx:
Thanks for this PR, @atassis. It's a well-scoped fix that closes a real gap. 

A few notes on the Copilot review comments before merge:

Default-case check (worth fixing): `# CHECK: air.channel @chan_default []` is a prefix match — it would still pass even if `buffer_resources` were erroneously emitted. Please add a `// CHECK-NOT: buffer_resources` after that line so the default (depth 1, no attribute) is actually guarded.

Positional-arg order (cheap hardening, optional): inserting `buffer_resources` before `loc/ip` shifts their positional meaning. I checked all current callers and they use keyword args, so nothing breaks today — but making the new param keyword-only (or placing it after `ip`) removes the future footgun for a one-line cost. Your call.

The `brittle-[] {…}` and `IntegerAttr-normalize/reject-depth-<1` comments I'd suggest declining: the assembly format is stable, callers pass plain ints, and the extra validation adds surface for an internal knob without real benefit.

So: no. 1 please, no. 2 if you agree it's worth it, and feel free to dismiss the other two with a quick note.

atassis:
@erwei-xilinx hi! Have made a commit per your request, hope that is enough to merge

PR #1725, opened 2026-07-11
[AIRRtToNpu] Support per-readback air.await_appends barrier

erwei-xilinx:
## Summary

Generalizes the `air.await_appends` / `air.append_barrier` same-DDR
read-after-write ordering to a runtime sequence that contains **more than one
independent readback** (e.g. an unrolled loop with N append/readback pairs).

Previously only a single readback was honored: every `air.append_barrier`
append's completion await was collapsed before the *first* tagged readback and a
warning was emitted for any others. With multiple readbacks that also produces an
`operand #0 does not dominate this use` error, because a later readback's append
await is moved ahead of the op that defines it.

Now, when multiple readbacks are tagged, each append's completion await is moved
before the **first tagged readback that follows the append in program order** —
the readback that consumes it, per readback. The single-readback path is
byte-for-byte unchanged.

## Details

- `AIRRtToNpuPass.cpp`: split the append-barrier walk into a single-readback
  path (unchanged behavior) and a per-readback path that, for each
  `air.append_barrier` append, finds the first following `air.await_appends`
  readback (by program-order index) and moves the append's await before that
  readback's start.

## Test plan

- [x] Adds case (5) to `runtime_sequence_ordering.mlir`: two readbacks, each fed
  by two appends (mirrors a KV-cache append shape). Uses SSA-identity CHECKs so
  each readback is verified to await *its own* two appends before starting.
- [x] Load-bearing: on `main` this input errors with `operand #0 does not
  dominate this use`; with the change it lowers correctly and the test passes.
- [x] `ninja check-air-mlir`: no regressions (pre-existing `Util/Channel`
  host-compile env failures unrelated to this change).

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -1807,67 +1807,30 @@ struct AIRRtToNpuPass : public impl::AIRRtToNpuBase<AIRRtToNpuPass> {
       // appends' completion awaits to just BEFORE the tagged readback's start,
       // so the runtime blocks on append completion before reading back.
       //
-      // A runtime sequence may contain MORE THAN ONE independent readback (e.g.
-      // an unrolled loop with N append/readback pairs). In that case each
-      // append's completion await is ordered before the FIRST tagged readback
-      // that follows the append in program order -- i.e. per readback, not
-      // collapsed onto the first readback (which would move a later append's
-      // await ahead of an earlier readback, violating SSA dominance and the
-      // append->readback ordering).
+      // A runtime sequence may contain one or MORE independent readbacks (e.g.
+      // an unrolled loop with N append/readback pairs). Each append's
+      // completion await is moved before the FIRST tagged readback start that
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Addressed the review feedback in 447d7b3a:

- **Copilot (warning suppression):** the "no ordering enforced" warning previously keyed off whether any `air.append_barrier` *config* op existed, so an append with no `dma_await_task` wrongly suppressed it. Now tracks whether an actual await was found (`anyAppendAwait`), matching the intended semantics.
- **Duplication:** folded the single-readback special case into the general per-readback algorithm — it was a strict subset (move each append's await before the first following readback start; with one readback that *is* the single readback). This also collapses the repeated configure→start-task user-scan into one `getStart` lambda. Net: −102/+... in that block, single code path for 0/1/N readbacks.
- **Tests:** added case (6) — a trailing `air.append_barrier` append with no readback after it must be left deferred (await not hoisted before the earlier readback), guarding the "first following readback" match.

`ninja check` for `Conversion/AIRRtToNpu`: 13/13 pass.

PR #1727, opened 2026-07-11
[Transform] Fix O(N^2) scaling in air-opt-shim-dma-bds on large unrolled launches

erwei-xilinx:
## Summary

`air-opt-shim-dma-bds` compile time blows up on modules with large fully-unrolled `air.launch` iteration spaces (surfaced by the `llms/llama32_1b_int4` `o_ffn` kernel group, whose `launch (16,32)` fully unrolls into a very large BD chain). Profiling the pegged worker thread showed two independent super-linear paths inside the pass; this PR removes both.

### 1. `IsolateAsyncDmaLoopNestInSCFForPattern` (AIRDependencyScheduleOpt.cpp)
The pattern hoisted marked allocs by running a greedy driver over the whole enclosing `IsolatedFromAbove` op. That greedy run does a full-region worklist walk **and** region DCE, yet it fired once per `scf.for` the pattern rewrote — O(N²) in the number of loops, since every launch/segment holds all the loops.

Fix: collect only this loop's `hoist_alloc`-marked allocs and seed the driver with them via `applyOpPatternsGreedily`, with region simplification disabled. Hoisting only relocates alloc/dealloc ops; CFG cleanup is handled by the pipeline's later canonicalize pass. The `hoist_alloc` match condition is factored into a single `isMarkedForHoist()` helper shared by the pattern and the seed walk, so the seed set stays provably identical to what the pattern would fire on.

### 2. `preserveAsyncDependenciesAfterUnroll` (Util/Dependency.cpp)
After a full unroll, this walked each unrolled op set and called `properlyDominates` (→ `isBeforeInBlock`) while **interleaving** `WaitAllOp` insertions. Each insertion invalidates the block's op-order cache, so `Block::recomputeOpOrder` (O(block)) re-ran on essentially every token use — O(N²) on the fully-unrolled block.

Fix: split into two passes — materialize all `WaitAllOp` wrappers first, then add async dependencies in a second pass using a single hoisted `DominanceInfo`. Adding an async-dependency operand doesn't change the CFG, so the op-order cache stays valid across the queries.

## Notes
- No functional change intended — both are algorithmic restructurings of existing behavior.
- These remove the two MLIR-pass hotspots; on the int4 `o_ffn` case the remaining compile cost is downstream (backend lowering of the large unrolled module), so this does not by itself bring that particular kernel group under the CI lit timeout.
- Functional equivalence is covered by the existing `check-air-mlir` suite plus a new multi-launch case in `isolate_async_dma_loop_nest.mlir` (two launches, each with a two-subgraph loop sharing one in-loop alloc, verifying both launches are hoisted/split independently under per-loop seeding). No wall-clock assertion is added because lit cannot reliably gate on compile time.

## Test plan
- [x] `ninja check-air-mlir` (456 passed, 0 failed)
- [ ] CI: full Ryzen AI e2e suite


atassis (commented):
Nice work - two genuine hotspots, and both restructurings look right to me. Two things:

Fix #1, seeding coverage. (Worked the hoisting-out-of-herds path in #1695 , so this caught my eye.) The old driver hoisted over the whole `IsolatedFromAbove` parent, so a `hoist_alloc` alloc got hoisted even if its own loop returned early on `target_ops_sets.size() < 2`. The new `for_op.walk` seed only reaches allocs under a qualifying loop. Is every marked alloc guaranteed to sit under one? A marked alloc in a non-qualifying sibling loop would now be dropped. (Nested loops are fine, since `walk` recurses.)

Regression guard. You note lit can't gate wall-clock. A pass `Statistic` on `DominanceInfo` constructions (or hoist-driver runs), with a CHECK that it stays `1` / `== num-loops`, catches a reintroduction on a tiny input, no timing. I have a synthetic N-loop generator that reproduces the scaling if a microbenchmark helps. Also, does fix #2's two-pass path get a dedicated test, or is it covered by an existing unroll case?

Otherwise a clean, functionally-equivalent restructuring.

erwei-xilinx:
Thanks @atassis.

**Fix #1 seeding coverage** — you're right, the invariant didn't hold: `hoist_alloc` was set inside `identifyTargetOpsInSCFFor`, i.e. *before* the `target_ops_sets.size() < 2` early-return, so a non-qualifying sibling loop could leave marks that only the old whole-region driver would sweep up. Fixed in 435cf999: the marking now runs in `matchAndRewrite` only after the size check, so a mark is only ever created under a loop that is about to be split — the `for_op.walk` seed is then complete. Non-qualifying loops are no longer marked, which is correct since hoisting only exists to share an alloc across the split loops. `check-air-mlir` still green.

**Fix #2 coverage** — it's exercised by the existing unroll paths (`specialize-channel-wrap-and-stride.mlir`, `opt_shim_dma_bds*.mlir`), not a dedicated case; those go through `loopUnrollFullWithAsyncTokenPreserved` → `preserveAsyncDependenciesAfterUnroll` and validate the output is unchanged.

**Regression guard** — there's no `STATISTIC` precedent in `mlir/lib` today, so I held off on wiring one + a `-stats` CHECK rather than introduce a new mechanism in this PR. Happy to add your synthetic N-loop generator as a follow-up (a `Statistic` on hoist-driver runs / `DominanceInfo` constructions with a CHECK it stays O(num-loops)) if you think it's worth a standalone guard.

atassis (approved):
Cool, I have checked that on my machine, seems good!

PR #1730, opened 2026-07-13
[CI] Nightly LLM perf benchmark on Krackan Point runner

erwei-xilinx:
## Summary
- New scheduled workflow `nightlyPerfBenchmark.yml` (cron `17 4 * * *` + `workflow_dispatch`) on the dedicated `amdryzenai5pro340` (Ryzen AI 5 PRO 340 / Krackan Point) runner.
- Runs the `llama32_1b` **verify** correctness gate (top-k token-set vs HF bf16), then captures **profile** numbers, reporting **TTFT** (prefill) and **decode tok/s**.
- `programming_examples/llms/bench/extract_perf.py` parses `make profile` stdout into `perf.json` (metrics + mlir-air/mlir-aie/llvm-aie SHAs); a job summary surfaces PASS + the two metrics; `perf.json`+`profile.txt` uploaded as an artifact.
- Kept **out of the per-PR matrix** (`buildAndTestRyzenAI.yml` untouched) so perf is contention-free. This is PR 1 of 2 — a follow-up wires up a GitHub Pages dashboard.

## Design notes
- verify runs before profile: perf on wrong output is meaningless, so the job fails if the correctness gate fails.
- Pinned `MODEL=base N_TOKENS=128` + fixed prompt for run-to-run comparability.
- All three toolchain SHAs captured per datapoint (llvm-aie is pulled nightly, mlir-aie from pinned wheel).

## Test plan
- [x] Hardware-verified locally on NPU2 (Strix): compile OK, `[verify] PASS` (2/2 prompts), profile TTFT=1235.34 ms, decode=11.18 tok/s.
- [x] `extract_perf.py` validated against real hardware profile output (correct JSON, exit 0; exits non-zero on missing metric line).
- [ ] After merge: `gh workflow run nightlyPerfBenchmark.yml` to validate end-to-end on the runner (secrets/self-hosted runner only available post-merge on the default branch).

### DIFF
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -30,13 +30,15 @@ env:
   XILINXD_LICENSE_FILE: /opt/xilinx/Xilinx.lic
   VITIS: /opt/ryzen_ai-1.3.0.1/vitis_aie_essentials
   # Pinned run params, shared across the benchmark + extract steps for
-  # run-to-run comparability (single source of truth).
+  # run-to-run comparability (single source of truth). Model is data, not a
+  # fixed name — add further models incrementally without renaming anything.
+  BENCH_MODEL: llama32_1b
   BENCH_PROMPT: What is the capital of France?
   BENCH_N_TOKENS: "128"
 
 jobs:
   benchmark:
-    name: Nightly LLM Perf Benchmark (llama32_1b)
+    name: LLM perf benchmark
     runs-on: amdryzenai5pro340
 
     steps:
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Thanks @Copilot — all three addressed in 68b13aca:

- **`grep -P` portability** → switched to `awk` for the mlir-aie HASH extraction, plus a `:` `${VAR:?}` guard so the step fails if the HASH line ever disappears.
- **Duplicated prompt** → hoisted the pinned prompt/n-tokens into workflow-level env (`BENCH_PROMPT` / `BENCH_N_TOKENS`); the benchmark and extract steps now share a single source.
- **File opened without context manager/encoding** → now `with open(..., encoding="utf-8")`.

PR #1731, opened 2026-07-14
[CI] Fix nightly LLM perf benchmark on Krackan runner

erwei-xilinx:
Follow-up to #1730. The nightly perf workflow failed on the dedicated `amdryzenai5pro340` (Krackan Point) runner; this fixes it. Verified green end-to-end on the runner via `workflow_dispatch` (compile → verify PASS → profile; captured TTFT=1409 ms, decode=9.24 tok/s).

## Root causes & fixes (workflow only)
- **npu2 detection fell back to npu1 → ELF rejected.** The NPU driver mmaps ~64 MB of *locked* memory; the benchmark step ran with the default `RLIMIT_MEMLOCK` (each GHA step is a fresh shell, so the build step's `prlimit` didn't carry over), so device probing failed with `EAGAIN` and the backend silently defaulted to npu1. → `sudo prlimit -lunlimited --pid $$` in the benchmark step.
- **Kernel compile couldn't find the AIE toolchain** (it runs via `make`, not lit). → set `MLIR_AIE_INSTALL_DIR` + `PYTHONPATH` + tool dirs on `PATH` the way `lit.cfg.py` does, *without* polluting `LD_LIBRARY_PATH`.
- **HF model download hung at 0% (~52 min).** The `hf_xet` transfer backend can't reach the Xet CAS endpoint on this runner. → `HF_HUB_DISABLE_XET=1` (classic HTTPS CDN).
- **No runtime bound.** → `timeout-minutes: 40` (step) / `90` (job) so a stuck download/dispatch can't squat on the dedicated runner.

No changes to shared code — the stock `xrt-smi` detection works once memlock is raised.

## Test plan
- [x] `workflow_dispatch` on `amdryzenai5pro340`: all steps success; `perf.json` artifact + job summary produced.

### DIFF
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -168,7 +168,7 @@ jobs:
           export MLIR_AIE_INSTALL_DIR="$MLIR_AIE_DIR"
           export PEANO_INSTALL_DIR="$PEANO_DIR"
           export PATH="$(pwd)/install/bin:$MLIR_AIE_DIR/bin:$PEANO_DIR/bin:$PATH"
-          export PYTHONPATH="$(pwd)/install/python:$MLIR_AIE_DIR/python:/opt/xilinx/xrt/python:$PYTHONPATH"
+          export PYTHONPATH="$(pwd)/install/python:$MLIR_AIE_DIR/python:/opt/xilinx/xrt/python${PYTHONPATH:+:$PYTHONPATH}"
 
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Thanks @Copilot — addressed in ef6dca64. Switched the append to `${PYTHONPATH:+:$PYTHONPATH}` so an empty `PYTHONPATH` no longer leaves a trailing `:` (which would put cwd on `sys.path`). Verified: empty → `x:y`, set → `x:y:/z`.

PR #1745, opened 2026-07-18
[AIRToAIE] Unify per-arm on-chip buffers of an index_switch onto one ring

erwei-xilinx:
## Summary

For an `scf.index_switch` whose mutually-exclusive arms each consume the **same** `air.channel` into **different** per-arm on-chip (L1/L2) buffers, unify those arm buffers onto **one shared buffer + one count-free ring**.

Only one arm executes per dispatch, so per-arm buffers and per-arm DMA tasks are redundant. On the same physical channel, distinct `dma_start` tasks ahead-block the idle arm's task at the head of the channel queue and stall the active arm; the extra unaccounted buffers also unbalance the lock contract (more `aie.buffer`/`dma_start`/`use_lock` on the consuming tile than the ring can service → the ring waits forever).

This is the device-IR companion to the fused multi-iteration launch runtime support (`#1741`): the runtime feeds one channel across heterogeneous waves, and this pass makes the on-chip consumer a single shared ring instead of one buffer per wave-arm.

**Scoping guard:** the transform is restricted to **on-chip (L1/L2)** switches. The launch-scope host-feed selector (arms move L3/DDR launch args; condition is the per-wave launch index) is owned by the runtime lowering — mutating it here would orphan its wave-selector `arith.cmpi` when air-to-aie dissolves the launch.

## Test plan

- [x] New lit test `mlir/test/Conversion/AIRToAIE/air_channel_index_switch_shared_ring.mlir`: two mutually-exclusive `scf.index_switch` arms consuming one channel into different buffers (asymmetric 7-vs-1 read counts) lower to exactly one `aie.dma_start(S2MM, 0` self-looping ring (`CHECK-NOT` a second S2MM task). Verified it fails without the fix (two S2MM tasks) and passes with it.
- [x] `ninja check-air-mlir` — no new failures vs `origin/main` baseline; all AIRToAIE tests pass (the pass runs unconditionally but is a no-op unless the exact same-channel/multi-arm/multi-buffer pattern is present).

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -3505,17 +3505,33 @@ void unifyIndexSwitchArmBuffers(AIE::DeviceOp d) {
             return true;
         return false;
       };
-      auto isArmVaryingGet = [&](SmallVector<air::ChannelInterface> &ops) {
-        if (ops.size() < 2)
-          return false;
-        if (!isa<air::ChannelGetOp>(ops[0].getOperation()))
-          return false;
+      // A channel name's byChan bucket holds BOTH gets and puts; this
+      // dedication is about S2MM (get) channels only, so classify from the get
+      // ops alone. Filtering here (rather than assuming ops[0] is a get) keeps
+      // the result independent of get/put visitation order within the bucket.
+      auto getsOf = [](SmallVector<air::ChannelInterface> &ops) {
+        SmallVector<air::ChannelInterface> gets;
+        for (auto ci : ops)
+          if (isa<air::ChannelGetOp>(ci.getOperation()))
+            gets.push_back(ci);
+        return gets;
+      };
+      auto armsOf = [&](SmallVector<air::ChannelInterface> &ops) {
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Addressed in ba9aae1e1. All three comments point at the same root cause: the S2MM channel-dedication logic classified channels via `ops[0]` and computed the `arms` set over **all** ops in the `byChan` bucket, which holds both gets and puts. If a put was visited first, an arm-varying get channel could be misclassified (skipping needed pinning) or a channel could look multi-arm due to puts alone.

Fix: added `getsOf()` (filter the bucket to `air.channel.get` ops) and `armsOf()` (arm set over a given op list), and rewrote `isArmVaryingGet`, the `numArmGetChans` count, and the dedication loop to classify from the **get ops only** — using `gets[0]` for the declaration lookup and `armsOf(gets)` for arm-spanning. The result is now independent of get/put visitation order in the bucket.

Verified: the new lit test still passes, and the motivating fused-decode design still lowers to the correct per-tile buffer count (behavior unchanged on the ordered case, robust in general).

PR #1748, opened 2026-07-24
ci: ccache mlir-air C++ across wheel runs

atassis:
## What

Cache mlir-air's own C++ compilation across wheel-build runs. `buildAIRWheels`
fans out 8 legs (python 3.11-3.14 x RTTI on/off); each recompiles the mlir-air
core from scratch with no ccache, so nothing is reused between runs even though
only the Python bindings differ across the Python legs.

## How

- A ccache restored per `(ENABLE_RTTI, python_version)` leg. The legs are
  separate parallel jobs, so they cannot share a live cache within one run; the
  reuse is cross-run. `run_id` in the save key writes a fresh immutable entry
  each run, and `restore-keys` drops it to restore the leg's own latest prior
  cache.
- A second `restore-keys` fallback on RTTI alone lets a cold or evicted leg warm
  the Python-independent C++ core from a same-RTTI sibling. ccache content-hashes
  the preprocessed source, so shared core objects hit while each leg's
  version-specific binding objects miss and rebuild. Nothing stale is reused, and
  it cushions GitHub's 10 GB per-repo cache eviction (8 legs times up to 2 GB).
- `compiler_check=content` because the manylinux compiler mtime is unstable and
  would otherwise miss every object; `sloppiness=locale,time_macros`;
  `max_size=2G`. ccache is installed as a static binary so it does not need the
  image's yum repos, and landing it on `PATH` is what routes the compile through
  it: the wheel's `setup.py` already adds the CMake compiler launcher when it
  finds ccache.

## Result

The C++ core is Python-version-independent, so once a leg is warm the wheel
build recompiles only what actually changed; the first run after a core change
correctly misses and rebuilds. Same `compiler_check=content` reasoning as #1747,
applied to the wheel build instead of lintAndFormat.

### DIFF
--- a/.github/workflows/buildAIRWheels.yml
+++ b/.github/workflows/buildAIRWheels.yml
@@ -176,6 +176,23 @@ jobs:
         with:
           submodules: "true"
 
+      # Cross-run ccache. The 8 matrix legs are separate parallel jobs, so they
+      # cannot share a live cache within a run; reuse is run to run. run_id in
+      # the save key writes a fresh immutable entry each run, and restore-keys
+      # drops it so a leg restores its own latest prior cache. The second
+      # fallback, on RTTI alone, lets a cold or evicted leg warm the
+      # Python-independent C++ core from a same-RTTI sibling; ccache content-
+      # hashes preprocessed source, so each leg still rebuilds its own
+      # version-specific Python bindings.
+      - name: Restore ccache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        with:
+          path: ${{ github.workspace }}/.ccache-air-wheels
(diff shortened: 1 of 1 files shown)


erwei-xilinx (approved):
LGTM, thanks for working on this, @atassis! Happy to land this. Before merge, could you please resolve the two Copilot comments?

atassis:
@erwei-xilinx Hi! Applied the curl `-f` one. 
The second one turned out to point at something else: the `CMAKE_ARGS` line is dead. `utils/mlir_air_wheels/setup.py` already adds `-DCMAKE_{C,CXX}_COMPILER_LAUNCHER=ccache` when ccache is on PATH, so the flags were landing on the cmake line twice. Dropped it instead of reordering.

PR #1757, opened 2026-07-29
Skip NFL channel fusion for puts/gets inside a multi-result scf.if

erwei-xilinx:
## Summary

The "new for loop" (NFL) channel-merge path in `air-fuse-channels` (`AIRDependencyScheduleOpt.cpp`) wraps the merged channel's put/get in a freshly created `scf.for` via `wrapRegionsWithForLoops`. That helper can only wrap a region whose parent op yields a **single** async token.

Broadcast specialization can place several `air.channel.put`s in **one multi-result `scf.if`** (e.g. a per-row multicast of a broadcast channel). Feeding such a channel into the NFL merge makes `wrapRegionsWithForLoops` erase the `scf.if` while its results `#1..` are still in use, crashing the pass.

This PR guards the NFL merge: skip a candidate pair when either channel has a put/get inside a multi-result `scf.if`. Leaving those channels un-multiplexed is correct — channel fusion is only an optimization, so the result is still valid, just un-fused for that channel.

## Test plan

- [x] `ninja check-air-mlir` (Transform / Conversion suites) pass
- [x] Verified on a broadcast-channel flash-attention module that previously crashed `air-fuse-channels` in the NFL path; it now compiles and the end-to-end LLM prefill token gate passes.

### DIFF
--- a/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
+++ b/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
@@ -4182,6 +4182,28 @@ public:
     std::map<air::ChannelOp, air::ChannelOp> chan_merge_map;
     std::vector<std::pair<air::ChannelOp, air::ChannelOp>> nfl_merge_pairs;
 
+    // wrapRegionsWithForLoops (used by the NFL merge below) can only wrap a
+    // region whose parent op yields a single async token. Broadcast
+    // specialization can place several channel puts in one multi-result scf.if
+    // (e.g. a per-row multicast of a broadcast channel), which that helper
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Both points addressed in the follow-up commit: (1) the guard now walks **all** enclosing `scf.if` ancestors (not just the nearest) and flags if any has >1 result, so a put/get nested in a single-result `scf.if` inside a multi-result `scf.if` is caught; (2) the per-channel result is memoized in a `DenseMap<Operation*, bool>` so each `air.channel` is scanned at most once across the O(N^2) pair loop.

PR #1761, opened 2026-07-31
[AIR] Decline wrap-and-stride fold on non-constant wraps/strides instead of crashing

erwei-xilinx:
## Summary
`air-specialize-channel-wrap-and-stride` aborted with an assertion when a channel op inside a loop carried an **IV-dependent wrap size or stride** (e.g. a causal-triangle transfer of size `(i+1)*N`, or an `(i+1)*S` stride). The specialize patterns and the wrap/stride canonicalizer dereferenced `getConstantIntValue` unconditionally.

A variant-shape transfer can never fold into a single constant-shape buffer descriptor, so the fold is genuinely impossible — but the loop is still valid IR that lowers by **unrolling** into a chain of constant-shape descriptors. The pass should decline the fold, not crash.

## Changes
- `AIRSpecializeChannelWrapAndStrideInScfFor` / `...InAffineFor`: early-`return failure()` when the channel op carries a non-constant **wrap or stride**, so the loop falls through to `AIRUnrollScfForIntoBDChain` / `AIRUnrollAffineForIntoBDChain`.
- `canonicalizeWrapAndStrideList`: after list-length normalization, return early when any size/stride is dynamic, before the constant-folding steps that assume constant values.
- `AIRCanonicalizeChannelPutGetOpWrapAndStrideList`: guard the `highestDimRepeatActive` reads of `strides.front()`/`sizes.front()` (a dynamic highest dim is neither a provable repeat nor foldable → treated as not-a-repeat). Closes the `enableRepeatAtHighestDim=true` path used by `AIROptimizeShimDMABDs`.
- New lit test `specialize-nonconstant-wrap.mlir`: non-constant **wrap** (`scf.for` and `affine.for`) and non-constant **stride** (`scf.for`) unroll without crashing; a rectangular IV-in-offset loop still folds to one strided descriptor.

## Test plan
- [x] `ninja check-air-mlir` (477 passed, 0 failed)
- [x] New lit test passes via FileCheck (wrap / stride / affine / rectangular)
- [x] Rectangular wrap-and-stride folding unchanged

### DIFF
--- a/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
+++ b/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
@@ -2742,7 +2742,7 @@ struct AIRSpecializeChannelWrapAndStrideInAffineFor
     SmallVector<Value> strides = channel_op.getStrides();
 
     // Non-constant wrap/stride can't fold into a constant-shape BD; decline so
-    // the loop falls through to AIRUnrollScfForIntoBDChain (see the scf.for
+    // the loop falls through to AIRUnrollAffineForIntoBDChain (see the scf.for
     // twin above for the rationale).
     auto isNonConst = [](Value v) { return !getConstantIntValue(v); };
     if (llvm::any_of(wraps, isNonConst) || llvm::any_of(strides, isNonConst))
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Thanks for the review — both addressed in 3f4234d:

1. **Unconditional deref under `enableRepeatAtHighestDim`** (AIRDependencyScheduleOpt.cpp:2883-2886): `highestDimRepeatActive` now reads `strides.front()`/`sizes.front()` through `std::optional` guards. A dynamic highest stride/size can be neither proven a repeat dim (stride 0) nor folded, so it's treated as not-a-repeat and falls through to `canonicalizeWrapAndStrideList` (which declines on dynamic dims) instead of asserting. This closes the remaining crash path for the `AIROptimizeShimDMABDs` (repeat-enabled) case.

2. **Misleading comment**: the affine.for specialize pattern now names `AIRUnrollAffineForIntoBDChain` (not the scf.for unroller).

`ninja check-air-mlir` still 477/0.

PR #1763, opened 2026-07-31
[AIRToAIE] Infer shared L1 across opaque calls; own the buffer on its DMA tile

erwei-xilinx:
## Summary

Two related fixes to **intra-herd (core-to-core) shared L1 buffer lowering** in AIRToAIE, for herds where neighbor cores communicate through a segment-scope L1 buffer passed as a herd operand and per-core roles are selected from the tile index.

1. **`air::herdBufferHasCrossCoreDependence` (Util.cpp)** — the reader/writer scan is memory-effect based and cannot see accesses hidden behind an **opaque external kernel call** (a `CallOpInterface` op with no `MemoryEffectOpInterface`). Such a call now classifies its buffer operand as both read and written, **but only when the call is reachable on a strict, non-empty subset of the herd's cores** (an iv-guarded role split = genuine neighbor communication). A call reachable on *every* core is replication, not communication, and is ignored so a non-partitioned herd operand is not wrongly accepted (the `AIRVerifyHierarchyLocality` negative case stays intact). Without this, a buffer touched only through opaque kernels is sunk into a private per-core copy and the cross-core hand-off is lost.

2. **Shared-L1 owner-tile selection (AIRToAIEPass.cpp)** — a `UseLockOp` inside an `aie.mem`/BD block may only access a lock on its own tile (`AccessesLocalLocks`), whereas a `UseLockOp` inside a core may access a neighbor's lock. When a shared buffer is read out by a channel put/get (which lowers to a DMA BD in the accessing tile's `aie.mem`), the buffer's owner tile **must** be that DMA-accessor tile, otherwise the BD references a cross-tile lock and verification fails with *"can only access a lock in the same tile"*. Owner selection now prefers a DMA-accessor tile among the participating cores.

## Test plan

- [x] New lit test `single_herd_shared_l1_opaque_call.mlir` — a `[1,2]` herd whose producer writes the shared buffer via an opaque `func.call`; asserts one shared `aie.buffer` + prod/cons locks. Red before fix 1 above (buffer is sunk per-core).
- [x] New lit test `single_herd_shared_l1_dma_owner.mlir` — both cores write via opaque calls, the lead reads out via `air.channel.put`; asserts the shared buffer + locks land on the DMA-accessor tile and the `aie.mem` BD uses same-tile locks. Red before fix 2 above (cross-tile lock verifier error).
- [x] Both tests confirmed red without the fixes, green with.
- [x] Full `mlir/test/Conversion/AIRToAIE` suite passes (122/122); `Transform/AIRVerifyHierarchyLocality` passes (3/3).

### DIFF
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -1776,8 +1776,41 @@ LogicalResult createAIEModulesAndOutlineCores(
 
       // Find a tile that has legal memory affinity with all other cores.
       // Use AIE target model's isLegalMemAffinity() to validate placement.
+      //
+      // Owner-tile preference: a core may acquire a lock on a neighbor tile
+      // (UseLockOp inside a CoreOp is unconstrained), but a DMA/BD block
+      // (aie.mem) may ONLY acquire a lock on its OWN tile (UseLockOp under a
+      // MemOp must AccessesLocalLocks -- see AIEDialect UseLockOp::verify).
+      // A shared buffer that is read/written by a channel put/get becomes a DMA
+      // BD on the accessing tile, so that tile's aie.mem will use the buffer's
+      // prod/cons locks -- which are placed on the owner tile. Therefore the
+      // owner tile MUST be a DMA-accessor tile, otherwise the DMA references a
+      // cross-tile lock and verification fails. Try DMA-accessor tiles first.
+      llvm::SmallDenseSet<Value> ownerBufferAliases;
+      collectBufferAliases(memref, ownerBufferAliases);
+      auto coreHasDmaAccess = [&](AIE::CoreOp coreOp) {
+        bool found = false;
+        coreOp.walk([&](Operation *op) {
+          if (!isa<air::ChannelPutOp, air::ChannelGetOp>(op))
+            return;
+          if (!accessReachableInCore(op))
+            return;
+          for (Value operand : op->getOperands())
+            if (ownerBufferAliases.contains(operand))
+              found = true;
+        });
+        return found;
+      };
+      SmallVector<AIE::CoreOp> orderedCandidates;
+      for (auto coreOp : coreOps)
+        if (coreHasDmaAccess(coreOp))
+          orderedCandidates.push_back(coreOp);
+      for (auto coreOp : coreOps)
+        if (!coreHasDmaAccess(coreOp))
+          orderedCandidates.push_back(coreOp);
+
       AIE::TileOp ownerTile = nullptr;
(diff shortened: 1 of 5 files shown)


erwei-xilinx:
Thanks for the review. Addressed both comments in 5ee7b824b:

- **Util.cpp opaque-call scan**: removed the redundant operand re-scan — since `user` comes from `alias.getUsers()`, the buffer is already an operand; the `SmallDenseSet` dedups calls reached via multiple aliases.
- **AIRToAIEPass.cpp `coreHasDmaAccess`**: the walk now interrupts on first match, and the DMA/non-DMA partition is built in a single pass so each core is walked at most once.

No functional change; the two new lit tests are unchanged and still pass (full `Conversion/AIRToAIE` + `Transform/AIRVerifyHierarchyLocality` green).

erwei-xilinx:
Fixed the CI failure (`matrix_vector_multiplication/bf16_cascade/run_npu1_2col_4cascade_peano`) in ee8c23044.

**Root cause:** the cross-core dependence check keyed on "any reader tile != any writer tile", which fires for two *independent self-contained* cores (each cascade tail tile computes its own output slot and streams it out — different tiles, but no hand-off). That wrongly hoisted a per-core-private operand into a single shared buffer, producing an `aie.mem` BD that references a cross-tile lock.

**Fix:** require **asymmetry** for a genuine cross-core hand-off — some core must be a pure producer (consumed elsewhere) or pure consumer (of another core's write). If every participating core both reads and writes the buffer, it stays per-core private and is cloned. Also classify an opaque call's buffer operand by direction (last memref operand = producer) so a real producer/consumer split across opaque kernels still yields asymmetric sets. Added `single_herd_no_share_self_contained.mlir` covering the self-contained case; full `Conversion/AIRToAIE` (126) + locality suites green locally.

erwei-xilinx:
Pushed 23a516d8d to fix the second CI failure (`bf16_cascade` add 1col_2cascade numerical mismatch).

The earlier asymmetry rework still keyed the decision on *visible* accesses, which regressed a cascade whose per-core output is filled by a tile-id-indexed `channel.get` (each core its own copy): the tail tile stores its result (a "pure writer") and another tile loads its own copy (a "reader"), misread as a cross-core hand-off.

The opaque-call handling is now **purely additive**: the visible-access sharing decision is restored to its original form and left untouched (so any buffer with no opaque call keeps its exact prior behavior — cascade/matvec unaffected), and iv-guarded opaque calls only *add* sharing when an external-kernel producer's result is consumed on a different core. Verified on device: `bf16_cascade` add 1col_2cascade **and** 2col_4cascade both PASS; the three shared-L1 lit tests green (126/126).

PR #1770, opened 2026-08-05
[CI] Fix the LLM perf benchmark: q4nx readback, warm TTFT, steady-state tok/s

erwei-xilinx:
Makes the nightly LLM perf benchmark run green and publish meaningful numbers.

### The failure

`llama32_1b_q4nx` took the job red. It compiled, prefilled correctly, and dispatched decode, then died host-side unpacking the logits:

```
ValueError: To change to a dtype of a different size, the last axis must be contiguous
```

`bo.read()` returns an array whose stride metadata depends on the pyxrt build, and `.view(bfloat16)` on it raises on the perf runner. Replaced with the readback idiom `llms/shared/infra/cache.py` already uses for every model that passes CI — `np.frombuffer` over `bo.map()`, dtype-explicit, no `.view()`, zero-copy.

### The published numbers were understated by design

- **TTFT** was the whole first-query pipeline (worker spawn + the one-time 1.8GB host weight load + prefill), which the weight load dominates. The driver now repeats the prefill with weights resident and reports `Warm time to first token (TTFT)`; `extract_perf.py` prefers it and falls back for drivers that do not emit it. The warm pass is handed back via a sidecar so the cold figure stays honest.
- **decode tok/s** came from a run that stopped at EOS after ~6 tokens, so the average was dominated by the first (cold) dispatch. `make profile` now passes `--no-eos-stop`.

### Also

- `fused_decode` preflight now requires `llvm-link`'s major to **equal** Peano's. It only rejected `>= 23`, which catches the loud failure but not the silent one: an older `llvm-link` (e.g. LLVM 20) merges the inline-attn bitcode without complaint and miscompiles the attn kernels — the decode then runs at full speed and emits fluent garbage.
- The benchmark records the NPU power mode and the ACPI platform profile it ran at.

### Results

Nightly goes from red to `success`; all 12 models pass verify. `llama32_1b_q4nx` publishes for the first time.

| | before | after (runner) | Ryzen AI 9 HX 370 |
|---|---|---|---|
| verify | crash | pass | pass |
| ttft_ms | 12290 (cold) | 1138 | 908 |
| tok/s | ~32 (6-token avg) | 38.2 | 49.9 |

The runner reads ~39-40 tok/s where an HX 370 reads ~50. That is a property of the machine, not a regression: the identical build (same air SHA, mlir-aie wheel, Peano, `llvm-link`; byte-identical decode `insts.bin`) hits 49.85 on the HX 370, and both report 51.0 TOPS from `xrt-smi validate --run gemm`, Turbo power mode and DDR5-5600. The remaining difference is the runner's ACPI platform profile (`balanced`), which is now logged each run. Landing at 40 tok/s for now.

### DIFF
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -24,6 +24,11 @@ on:
         type: boolean
         required: false
         default: false
+      probe_only:
+        description: 'Only record machine state (power/clocks/NPU TOPS), run no models'
+        type: boolean
+        required: false
+        default: false
 
 defaults:
   run:
@@ -271,6 +276,7 @@ jobs:
           # cannot inject shell.
           ONLY_MODEL: ${{ inputs.only_model }}
           SKIP_VERIFY: ${{ inputs.skip_verify }}
+          PROBE_ONLY: ${{ inputs.probe_only }}
         run: |
           source air-venv/bin/activate
           export PATH=$VITIS/bin:$VITIS/aietools/bin:$PATH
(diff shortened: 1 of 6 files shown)


erwei-xilinx:
Good catch — fixed in 2be817a0.

The banner interpolated `_ll_ver` straight into a fixed `>=23` claim, so the `_ll_ver is None` branch — which the surrounding condition explicitly handles — printed:

```
llvm-link None on the test PATH is >=23; shimming ...
```

Wrong twice over: the bare `None`, and asserting a version the code had just failed to read. Failing to parse `--version` is a *different* reason to reject a binary than being ≥23. It now says `has an unparseable version` in that case and `is LLVM N` otherwise, and names the rejected binary's path in both:

```
llvm-link on the test PATH (/…/bogus_bin/llvm-link) has an unparseable version; shimming /…/llvm-link (LLVM 21) ahead of it.
```

Diagnostic text only — no behaviour change.

I also extended the PATH-layout harness with an `llvm-link` whose `--version` prints no version string, covering it both shadowing a usable LLVM 21 and alone. Six layouts now, each asserting the feature flag matches the version that actually resolves:

| PATH layout | feature | tests run |
|---|---|---|
| LLVM 23 shadowing LLVM 21 (the CI case) | enabled | shim → LLVM 21 ✅ |
| unparseable shadowing LLVM 21 | enabled | shim → LLVM 21 ✅ |
| unparseable only | disabled | LLVM unparseable ✅ |
| LLVM 23 only | disabled | LLVM 23 ✅ |
| LLVM 21 only | enabled | LLVM 21 ✅ |
| none | disabled | none ✅ |

`black` clean.

Note on the in-flight verification run [31044221957](https://github.com/Xilinx/mlir-air/actions/runs/31044221957): it checked out `ab32df2e` and is still running against it. Since this commit changes only a print string, that run remains a valid end-to-end test of the fix — no need to re-dispatch.

PR #1771, opened 2026-08-06
[AIRToAIE] Split aperiodic compute-tile S2MM packet chains

erwei-xilinx:
## Problem

A compute-tile S2MM BD chain is walked strictly in order. A packet header routes a
transfer *to the tile*; it does not select which BD receives it. So a chain shared by
several packet flows is only correct while every flow on it contributes the same
number of BDs — the ring then advances exactly one full cycle per round of arrivals.

The compute-tile packet-collapse in `TileDMAAllocator` had no such guard. When flows
are consumed a different number of times per dispatch — e.g. a sublayer buffer read
twice beside once-per-dispatch norm weights — they end up interleaved on one chain.
`getRepeatCounts` finds no repeating pattern, one BD is emitted per transfer, and the
ring length stops being a whole number of arrival rounds. The BD pointer slips further
every dispatch until a transfer meets a BD belonging to another flow, and the receiver
deadlocks on the second dispatch.

This is not hypothetical: it is what a 4-norm decoder layer (norm around both the
attention and the MLP) produces, and it is why the Gemma3-4B decode in the stacked PR
hangs without this fix.

## Fix

`rebalanceAperiodicPacketChains`, run after `sortMemcpyOps` and before flows are
connected. For each `(tile, channel)` chain — the same grouping the BD emitter itself
uses — whose flows have unequal multiplicity, peel the flow whose removal leaves both
sides balanced onto a spare S2MM channel.

Chains that are already balanced, and tiles with no spare channel, are left alone, so
**designs that work today are bit-identical**.

The flow bundles carry their own copy of the allocation, so those are retargeted in
lockstep — otherwise the switchbox keeps routing to the channel the BDs just left.

Also hoists `getRepeatCounts`' inner comparison lambdas to file scope. No behaviour
change; it makes the pattern check readable and reusable by the new code.

## Testing

- New lit test `mlir/test/Conversion/AIRToAIE/tile_aperiodic_packet_chain_split.mlir`
  covers the split, the already-balanced no-op, and the no-spare-channel no-op.
- Full suite: 481 passed, 7 unsupported, 7 expectedly failed, 0 failures.
- On NPU2 hardware, this is what makes the Gemma3-4B fused decode reach its second
  dispatch; Llama-3.2-1B decode is unaffected (Paris gate 12366, 40.8 tok/s).

## Note for reviewers

The Gemma3-4B example that motivated this is a separate stacked PR so this pass change
can be reviewed on its own.


erwei-xilinx:
The Gemma3-4B example that motivated this fix is #1772, stacked on this branch.

PR #1775, opened 2026-08-07
[flash_attention] Derive DK_TILE/DV_TILE from SCRIPT

erwei-xilinx:
Follow-up to #1773 (now merged). Rebased onto `main`; **the whole change is one Makefile hunk.**

## Problem

`-Ddk`/`-Ddv` bake the per-tile head-dim extents into the kernel and must agree with the `dk_tile`/`dv_tile` the Python builder uses. The builders disagree by construction:

- `attn_npu2_temporal_causal.py` keeps d whole — `dk_tile = dk`
- `attn_npu2.py`, `attn_npu2_seqfirst.py`, `attn_npu1.py` — `dk_tile = lkp`

A mismatch does not fail the build. It produces wrong results.

`DK_TILE ?= $(LKP)` leaves that to the caller, and it can desync in **both** directions:

- run the temporal builder at head_dim 128 without remembering `DK_TILE=128`, and the kernel is compiled for 32 while the builder emits 128;
- set `DK_TILE` for the temporal builder, then run a non-temporal `SCRIPT` in the same tree, and that builder desyncs the other way.

## Fix

Derive the default from `SCRIPT` so neither direction is expressible, and pin the NPU1 kernel to `LKP` outright since `attn_npu1.py` takes no tile flags at all. Callers that already pass `DK_TILE`/`DV_TILE` explicitly are unaffected (`?=`).

Emitted defines at `LKP=32 DK=DV=128`:

| SCRIPT | builder wants | before | after |
|---|---|---|---|
| `attn_npu2_temporal_causal.py` | 128 | `-Ddk=32` | `-Ddk=128` |
| `attn_npu2.py` | 32 | `-Ddk=32` | `-Ddk=32` |
| `attn_npu2_seqfirst.py` | 32 | `-Ddk=32` | `-Ddk=32` |

An explicit `DK_TILE=64 DV_TILE=64` still overrides to 64, as before.

## Testing

On device at 2048 ctx / 24 heads / 8 kv / head_dim 128 / lkp 32, the derived default reproduces the configuration #1773 reports — **24.0 ms**, against **16.4 ms** for the un-derived `DK_TILE=32` kernel.

I did not use the standalone `make run` precision check as a gate: in my tree it reports failure for every config including known-good ones, so it isn't a usable signal here.

## Note

This PR previously carried a larger head_dim > lkp enablement change. #1773 covers that ground more completely — including the 8-round deadlock, which it root-causes to the round count being the number of in-flight tasks on the K/V MM2S channel. Everything that overlapped has been dropped; only this fix remains, which #1773 did not include.

### DIFF
--- a/programming_examples/flash_attention/kernel_fusion_based/Makefile
+++ b/programming_examples/flash_attention/kernel_fusion_based/Makefile
@@ -19,11 +19,16 @@ NUM_KV_HEADS ?= $(NUM_HEADS)
 # Derived: kernel tile size = LQP / num_q_tiles (4)
 NUM_Q_TILES ?= 4
 LQP_TILE := $(shell echo $$(($(LQP) / $(NUM_Q_TILES))))
-# Per-tile head-dim extents baked into the kernel. Default to LKP (the designs
-# that tile d across the DMA); attn_npu2_temporal_causal.py keeps d whole and
-# passes DK_TILE=$(DK) DV_TILE=$(DV).
-DK_TILE ?= $(LKP)
-DV_TILE ?= $(LKP)
+# Per-tile head-dim extents baked into the kernel, as -Ddk/-Ddv. These MUST
+# agree with the dk_tile/dv_tile the Python builder uses, and the two builders
+# disagree by construction: attn_npu2_temporal_causal.py keeps d whole
+# (dk_tile = dk), while attn_npu2.py / attn_npu2_seqfirst.py / attn_npu1.py all
+# hardcode dk_tile = lkp. Derive the default from SCRIPT so neither direction
+# can silently desync -- a mismatch here does not fail the build, it produces
+# wrong results.
+IS_TEMPORAL := $(filter attn_npu2_temporal_causal.py,$(SCRIPT))
+DK_TILE ?= $(if $(IS_TEMPORAL),$(DK),$(LKP))
+DV_TILE ?= $(if $(IS_TEMPORAL),$(DV),$(LKP))
 
 
 # Determine build dir based on whether PEANO_INSTALL_DIR is set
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Addressed the review in a1bfde037. All five points were valid.

**Kernel/builder tile desync (Makefile:142/149/182/189)** — the real one. `DK_TILE`/`DV_TILE` fed `-Ddk`/`-Ddv` unconditionally, so setting them while running a non-temporal builder would compile the kernel and the builder with different tile widths. Kernel defines now route through `KERNEL_DK_TILE`/`KERNEL_DV_TILE`, which fall back to `LKP` unless `SCRIPT` is the temporal-causal builder; the NPU1 kernel is pinned to `LKP` outright since `attn_npu1.py` has no tile flags at all.

Verified the desync is no longer expressible: `attn_npu2.py` with `DK_TILE=128` deliberately set measures **796.1 us**, unchanged from the 795–798 us baseline.

**`print` / `profile` / `profile-xclbin` missing `TEMPORAL_PY_FLAGS`** — fixed; they were building the kernel with the new tile widths but the builder without them. The NPU1 targets deliberately get none.

**`roll_rounds` silently ignored when invalid** — now raises naming both the value and the round count. This one mattered: falling back to a full unroll is indistinguishable from "the option had no effect" precisely when you are chasing a program-size limit.

**`kv_shim_cols` unvalidated** — now validates length, distinctness and sign in `build_module`, and fails through `parser.error()` for malformed CLI input. Two equal columns silently defeated the entire point of the option, so that is rejected explicitly.

No behaviour change for any valid configuration:

| gate | result |
|---|---|
| default path (lkp=dk=dv=64, 32h/8kv, seq 2048, causal) | 9872 us, `npu.air.mlir` md5 `69c7a1ed969c02a55dadc30870f9ef0e` — byte-identical to `main` |
| head_dim 128 path | 1685 us |
| `attn_npu2.py` with `DK_TILE=128` set | 796 us — baseline |

PR #1777, opened 2026-08-07
Add Qwen2.5-3B on NPU2: Q4_0 prefill + fused decode

erwei-xilinx:
Stacked on #1771 — it builds on a tree that has the aperiodic-packet-chain split, so the diff below includes that PR's changes until it merges.

## What this adds

An end-to-end Qwen2.5-3B deployment: a Q4_0 prefill example and a 36-layer fused single-dispatch decode, joined so the whole pipeline runs on NPU2 with **no host reference in the loop**.

- `llms/qwen25_3b_q4/` — Q4_0 weight loader + prefill driver, reusing the `qwen25_3b` block builders exactly as `llama32_1b_q4nx` reuses `llama32_1b`'s.
- `fused_decode/fused_decode_qwen.py` + `models/qwen2.5-3b.h` — the 36-layer fused decode, plus the kernel changes it needs (Q4_0 signed-int4 in `q4_k.h`, QKV bias in `rope.cc`, head_dim 128 attention).
- `fused_decode/qwen_prefill_to_decode.py` — the KV hand-off and generation loop.

## Quant path: Q4_0, not Q4NX

Qwen2.5-3B has no pre-quantized 4-bit bundle, and its NPU kernels are built with `-DQ4_0` — a symmetric signed-int4 codec (`w = q*scale`, `q ∈ [-8,7]`, `scale = amax/8`, per 32-element group along the reduction dim), not the affine Q4NX codec the Llama Q4NX examples read. So this quantizes the full-precision HF checkpoint directly, which is also strictly more accurate than requantizing someone else's 4-bit weights. Layer-0 error is rel L2 ~0.095–0.098 on every projection. QKV bias is carried as bf16 (it is not quantized).

Prefill and decode share one quantizer, so they see bit-identical weight values and the prefill's KV cache is a valid decode input.

## Hand-off

A slice, not a shuffle: prefill K/V are `[ctx, 256]` seq-major head-major and the decode wants `[ATTN_L, 256]` at the K and V region offsets — same element order. Seeding from the **last** `ATTN_L` prefill rows keeps RoPE consistent, since RoPE is absolute and a cached entry carries the roping of the position it was computed at.

## Results (NPU2, quiet box)

| gate | result |
|---|---|
| First token, `"The capital of France is"`, 36 layers | argmax 12095 = `" Paris"` |
| Warm TTFT @ seq_len 2048 | 4.17 s |
| NPU prefill -> NPU fused decode, no host reference | coherent text, 9.54 tok/s device-only |

The cross-check that the hand-off is correct: the decode's first token equals the prefill's own argmax.

## Known weak point

Decode throughput. Layers do not pipeline (3.9 ms/layer × 36) and the full weight set streams per token. That is a follow-up, not addressed here.

## Notes

No lit tests yet — the example downloads a 3B checkpoint and compiles its own kernels, so wiring it into CI is deliberately left as a follow-up rather than bolted on here.

### DIFF
--- a/programming_examples/fused_decode/fused_decode_qwen.py
+++ b/programming_examples/fused_decode/fused_decode_qwen.py
@@ -1,0 +1,2153 @@
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+#
+# Qwen2.5-3B decode layer — AIR reimplementation (SKELETON, stage 1: proj grid).
+#
+# WHY A SEPARATE FILE: fused_decode.py (the Llama-1B design) is explicitly NOT
+# parametrized — it is hardcoded to the the Llama proj CASCADE topology (16 cores as
+# 8 lead+partner pairs at cols {0,1,6,7}, group/main memtiles). The reference's Qwen2.5 decode
+# layer uses a DIFFERENT device topology (verified by IR diff of the reference
+# LLAMA_3_2_1B layer IR vs the reference QWEN2_3B layer IR):
+#   proj = INDEPENDENT 4x4 SPMD grid, cols {2,3,4,5} x rows {2,3,4,5}, NO cascade;
+#          each core writes only its own y buffer, packet-routed by phase-id to its
+#          OWN per-column memtile (mem_tile_C_1, DMA 0-3 = rows 2-5).
+#   rms=tile(1,2) rope=tile(1,3) glu=tile(6,2) attn=col7 (2 CUs).
+# Keeping this in a separate builder leaves the Llama path byte-identical.
+#
+# SHIM/MEMTILE HOPPING (mirrors reference Qwen flows; cores NEVER touch shim directly):
+#   X BROADCAST : host -> X memtile (col 1, reference mem_tile_1_1) -> inX broadcast to all 16.
+#   WEIGHTS     : host -> per-column W memtile (col 2+cx) -> wL2ToL1 fan to that col's 4 cores.
+#   OUTPUT      : core -> per-column gather memtile (col 2+cx) -> outY -> shim drain.
+# This spreads shim DMA per-column (like the reference: shim_C_0 DMA0,1 -> mem_C_1) instead of
+# collapsing 16 per-core routes onto one centroid shim tile. Reuses the two-hop staging
+# pattern from fused_decode.py (_feed_inX / weight-fan / egress).
+#
+# proj SPMD contract (the reference proj kernel + array layout):
+#   MVM_CORES=16; each core computes M_total/16 output rows per phase, in m=32-row
+#   blocks (outer i loop), contracting K in k=256 col-blocks (inner j loop). Each
+#   32-row block is emitted as ONE packet: y buffer = 16 hdr + 32 payload = 48,
+#   pkt_id at y+14 (proj_qmm_flush_hdr writes exactly *(u32)(y+14)=id; copy<32>(y+16)).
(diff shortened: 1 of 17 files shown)


erwei-xilinx:
Addressed in 715d0c0be. Five of the six were valid; one I'm pushing back on.

**`prefill()` context handling** — correct, and the worst of the six. `base = self.current_context_length` was read but the embedding write and every layer's KV write are indexed from row 0 regardless, so a second call would silently overwrite the cache from the start while the counter kept accumulating. Incremental prefill isn't supported here, so it now asserts rather than corrupting.

**`dequant_q4_0` docstring** — right, the input is one byte per weight holding a single signed-int4 code; the two-per-byte packing happens later in the cascade packer. Docstring corrected and the distinction called out explicitly.

**Unclosed index file handle** — fixed with a context manager.

**"Analize" typo** — pre-existing on `main`, but this commit already rewrites that comment line, so fixed here.

**Host-side LM head vs "no host reference in the loop"** — a fair catch, the phrasing overclaimed. The intent was "no reference *model* in the loop": every token comes from the NPU prefill and NPU decode, never from a host copy of the model. But the final RMSNorm + LM-head projection and the argmax do run on the host each token. Both the script header and the README now say precisely that — 36 transformer layers on device, last projection on host — and note that moving it on-device is a performance item, not a correctness one.

---

**Not changed: `chess_loop_range(2, )` in `attn_kv.cc`.** This is the established idiom in this tree for "minimum trip count 2, no maximum" — the trailing empty argument is the open upper bound, not a missing one. Same form already appears in:

- `vector_matrix_multiplication/i8/single_core/vm.cc:39,59`
- `vector_matrix_multiplication/block_quantized_i8/single_core/vm.cc:42,72`
- `vector_matrix_multiplication/bf16/single_core/vm.cc:46`

PARIS gate re-verified after these changes: argmax 12095.

erwei-xilinx:
Rebased onto `main` at b2959c9a0 (Gemma3-4B #1772). Two conflicts, both resolved in favour of what had already landed:

- **`models/all_models.h`** — #1772 and this PR both claimed model id `1`. Gemma keeps `GEMMA3_4B 1`; Qwen moves to `QWEN2_5_3B 2`. Only symbolic `#if MODEL_TYPE == ...` comparisons exist, no hardcoded numerics, so the renumber is inert.
- **`kernels/rope.cc`** — both PRs independently hit the same Peano bug (vector `G_FNEG` on bf16 compiles silently and emits code that stalls the rope core) and both fixed it the same way, `aie::msc` on Peano / `neg`+`mac` on Chess. Took the landed version verbatim and dropped this PR's duplicate wording. What remains in `rope.cc` here is only the Qwen QKV-bias add.

`llms/README.md` auto-merged; both example entries are present.

## Re-verified on device against `main` @ b2959c9a0

Fresh kernel compile, cache deleted:

| gate | result |
|---|---|
| First token, 36 layers | argmax 12095 = `" Paris"` |
| Warm TTFT @ 2048 | **4.053 s** (505 tok/s prefill) |

No regression — it is faster than the 4.174 s originally reported in this PR. Per-ELF vs that baseline:

| ELF | before | now |
|---|---|---|
| gate | 24.10 | 22.47 |
| up | 23.92 | 22.29 |
| flash_attn | 22.48 | 22.47 |
| down_add | 13.06 | 13.13 |
| swiglu | 10.52 | 10.54 |
| o_res_norm | 5.48 | 5.39 |
| rms_qkv_bias_rope | 7.30 | 7.25 |

The ~6.5% gain on `gate`/`up` comes from #1773's shared GEMM builder changes; everything else is flat. I have not refreshed the numbers in the PR description — the 4.17 s there is the pre-rebase figure and is now conservative.

PR #1786, opened 2026-08-12
[programming_examples] Feed decode weights on both shim MM2S channels per column

erwei-xilinx:
## Why

Batch-1 decode is a weight-streaming problem: every token reads every weight exactly once, with no reuse, so tok/s tracks weight bandwidth almost 1:1. The fused decode drove each proj column's weights on **one** shim MM2S channel; FastFlowLM drives **two**. On the same four physical columns FLM pulls 14.35 GB/s per column where we pulled 10.16 — the column was not saturated.

## What

Feed each proj column on both of its MM2S channels. `W_DUAL_CHAN` is on by default; set it to `0` for the previous single-channel feed.

Three things are required, all mirroring FLM's `mem_C_1`:

| | what | why |
|---|---|---|
| split axis | **spatial**, by cascade pair — ch0 feeds rows 2/3, ch1 rows 4/5, on two independent lock cycles | a *temporal* split (alternating fan steps) gives every core one MM2S chain alternating between both channels' buffers, couples the two shim channels at every step, and deadlocks the first dispatch |
| shim placement | per-column channels `@inW{0,1}c{cx}` carrying `air.shim_col` | a `[NCX]` bundle cannot express a per-index column; unpinned, the extra flows take the rms/rope column and silently violate *its* `air.shim_col` pin |
| X memtile | on the hub column (`XMT_PCOL = MAIN_PCOL`) | FLM has **no** memtile in column 2. Leaving a 16-way broadcast hub there alongside the shim feeds and both cores makes the pathfinder fail outright once the weight flows double |

The DDR side is a pure permutation of the packed cascade (`pack_q4k_cascade(dual_chan=True)`), so each channel still reads one contiguous 1-D run. A strided feed cannot: a 10240-element fan step exceeds the AIE2 per-dim wrap limit, and only a contiguous BD gets the wide `buffer_length` register.

The flag changes the DDR weight layout, so build and run must agree. Consumers key their requant cache and their `.decode_flags` template stamp on it; the Qwen decode, which has no Makefile build, stamps the artifact instead and its runner refuses a mismatch rather than producing silent garbage.

## Measured (Strix, quiet box, `make profile N_TOKENS=64`, 3 runs each)

| model | single-channel | dual-channel | |
|---|---|---|---|
| Llama-3.2-1B | 46.2 tok/s | 52.3 tok/s | **1.13x** |
| Llama-3.2-3B | 18.8 tok/s | 22.5 tok/s | **1.20x** |
| Gemma3-4B | 15.2 tok/s | 17.1 tok/s | **1.12x** |
| Qwen2.5-3B | 105.1 ms/tok | 105.5 ms/tok | none |

Qwen is unchanged, as expected: it streams 1.91 GB/token at ~105 ms = **18.2 GB/s**, against the 1B design's 35.7 GB/s on four channels. Its decode is not weight-bandwidth-bound (its layers do not pipeline), so more channels cannot help. It is enabled there for uniformity and is correctness-verified.

The nightly LLM dashboard is rendered from CI's `perf.json` on the Krackan Point runner, so it will pick these up on that hardware without hand-edited numbers.

## Testing

- `llama32_1b_q4nx`: `make verify-full` 8/8, and **byte-identical greedy token ids** vs the single-channel build at short and ~1170-token context (same weights, same kernels, only DMA routing differs, so output must match exactly).
- `llama32_3b_q4nx`: `make verify` PASS on both arms.
- `gemma3_4b_q4nx`: Paris gate PASS on both arms, identical generated text.
- Qwen2.5-3B: identical generated text both ways; first token matches the prefill argmax.
- With `W_DUAL_CHAN=0` the emitted AIR is **byte-identical to before this change** for all four model configs (llama-3.2-1b/3b, gemma3-4b, Qwen), and `pack_q4k_cascade(dual_chan=False)` byte-matches the previous packer — so the fallback path is a provable no-op.

### DIFF
--- a/programming_examples/fused_decode/.gitignore
+++ b/programming_examples/fused_decode/.gitignore
@@ -14,3 +14,7 @@ build_chess/
 air.mlir
 air.elf
 .decode_build_model
+
+# build-flag stamp: makes the decode-template skip W_DUAL_CHAN-aware
+.decode_flags
+decode_qwen.flags
--- a/programming_examples/fused_decode/Makefile
+++ b/programming_examples/fused_decode/Makefile
@@ -34,8 +34,9 @@ PEANO_KBASE    := -std=c++20 --target=aie2p-none-unknown-elf \
 # DECODE_GOLDEN=1 enables the post-attention-RMS decode path (a boolean flag; no
 # weight files are read at build time -- the compile is weight-free).
 # W_DUAL_CHAN=1 feeds each proj column's weights on BOTH of its shim MM2S channels
-# (@inW even fan steps / @inW2 odd) instead of ch0 only. It also changes the DDR
-# cascade order, so any consumer must pack its weights with the SAME flag -- the
+# (@inW0c{cx} takes the low half of the column's cascade rows, @inW1c{cx} the high
+# half) instead of ch0 only. It also changes the DDR cascade order, so any
+# consumer must pack its weights with the SAME flag -- the
 # llms/*_q4nx drivers read it from the environment and key their requant cache on
 # it. Inherited from the environment so a consumer can set it once for both the
 # build and the run.
--- a/programming_examples/fused_decode/Makefile
+++ b/programming_examples/fused_decode/Makefile
@@ -33,7 +33,15 @@ PEANO_KBASE    := -std=c++20 --target=aie2p-none-unknown-elf \
   -DAIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16
 # DECODE_GOLDEN=1 enables the post-attention-RMS decode path (a boolean flag; no
 # weight files are read at build time -- the compile is weight-free).
-DECODE_ENV     := VOCAB_CHUNK_I2=14 LM_HEAD=0 NLAYERS=1 DECODE_GOLDEN=1
+# W_DUAL_CHAN=1 feeds each proj column's weights on BOTH of its shim MM2S channels
+# (@inW even fan steps / @inW2 odd) instead of ch0 only. It also changes the DDR
+# cascade order, so any consumer must pack its weights with the SAME flag -- the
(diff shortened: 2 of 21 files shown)


erwei-xilinx:
Thanks — all 8 comments were right, fixed in 4ff10033f.

**The seven stale comments (same root cause).** They described a *temporal* even/odd fan-step split across `@inW`/`@inW2`. That was genuinely the first design, and it deadlocked on device: giving every core one MM2S chain that alternates between both channels' buffers couples the two shim channels at every fan step. What shipped is the *spatial* split — channel 0 carries the low half of a column's cascade rows (`cy 0..NCY/2-1`), channel 1 the high half, so each channel feeds a disjoint set of cores through its own ring and the two are never ordered against each other (this is what FLM does: `mem_C_1` takes shim ch0 on S2MM4 and ch1 on S2MM5, two independent lock cycles). The comments now say that, and record why not to go back to the temporal form. `pack_q4k_cascade`'s docstring was already correct.

Code is untouched — the emitted AIR IR is byte-identical before/after (verified via `FUSED_DECODE_EMIT_ONLY=1`).

**The Qwen guard.** Also a real hole, and the case it most needs to catch: an xclbin built before the stamp existed has an *unknown* weight layout, so treating a missing/unparseable stamp as "compatible" is exactly when silent garbage gets through. It now refuses on `_got is None` as well as on a mismatch, and derives the stamp path from `--xclbin` rather than hardcoding `decode_qwen.flags`.

CI note: the failures on the previous run were a github.com 503 window (the `eudsl-python-extras` and `torch-mlir` dev-wheel endpoints both 503'd); everything else passed. The two remaining Ryzen AI failures are runner-side — `torch_mlir_e2e/*` hitting `DRM_IOCTL_AMDXDNA_EXEC_CMD ... (err=-5)` on amd8845hs, which `main` hits identically. Those tests do not touch `programming_examples`.

PR #1790, opened 2026-08-12
[programming_examples] Close the three gaps that let the gemma decode bug hide

erwei-xilinx:
The gemma readback bug fixed in #1788 sat latent through four nightlies. Three separate holes let it, and each is fixed here. Companion to #1789, which covers the fourth gap (qwen25_3b_q4's missing build path and CI gates) — the two are disjoint.

## 1. Gemma's gate only checked token 0

`make verify` generates 9 tokens and compared just `gen_ids[0]` — which comes out of the **prefill**, so the decode loop the gate exists to cover was untested. The known failure mode here is exactly a decode that starts right and then drifts: a stale decode template yields a correct first token and fluent garbage after it, which the old gate passes.

Now gates on the full greedy sequence, with the two misses separated because they point at different halves:

```
*** MISS *** first token [4242], expected [9079]        <- prefill / weights / KV seed
*** MISS *** decode drifted at token 1                  <- decode loop
    expected [9079, 236761, 108, 50429, ...]
    got      [9079, 111, 222, 333]
```

Greedy, so deterministic. The recorded sequence is identical across repeated runs and across `b40eb7a11` → `a6ac5dd12`, which included #1786's decode weight-feed change — so it is stable across a real code change, not just run to run.

## 2. The ELF cache was not keyed on the compiler

`shared/infra/cache.py` keyed manifest entries on kernel **name** alone, so a cache built by a different mlir-air or Peano was served without complaint — a run that "passes" on binaries nobody rebuilt. This bit the #1788 investigation directly.

Now records a toolchain fingerprint; a mismatch, or a pre-fingerprint manifest, is treated as cold with a printed reason. Peano and mlir-air are routinely used from a path rather than a wheel (local builds, extracted nightlies), so the fingerprint stamps the binaries as well as the dist versions.

## 3. `llama32_1b_q4nx`'s `clean` missed the templates it consumes

Its decode templates live in `fused_decode/`, whose `compile-decode` skips when both are present — so the lit's `clean && compile-decode` was a no-op onto whatever ran last. Not hypothetical: during the #1788 investigation it verified **pre-#1784 binaries**, passing on stale artifacts. Now delegates to the `fused_decode` clean, matching what the gemma and 3B targets already do to that shared dir.

## Also: widen the #1785 Peano pin comment

It cites `xrt/50_multi_launch_attention` and `xrt/48_triton_matmul_ver4_strix_4x4_bf16_output`. LLVM 22 also breaks every Qwen LLM example — `llms/qwen25_{0_5b,1_5b,3b}` fail their top-k gate and `llms/qwen25_3b_q4` returns NaN logits. Whoever reverts that pin needs the full checklist. Confirmed by holding mlir-air fixed and varying only Peano.

## Verification

- **Cache fingerprint round-trip:** same toolchain → loads; changed Peano → invalidates with reason; legacy manifest → invalidates.
- **`llama32_1b_q4nx clean`:** confirmed it now removes the shared `decode_L2047/L2048`.
- **Gate branches**, all six, including right-first-token-then-drift (the case the old gate passed):

```
[ok] exact golden sequence            -> *** PARIS ***
[ok] short but matching prefix        -> *** PARIS ***
[ok] stale-template: right 1st, drift -> *** MISS *** decode drifted at token 1
[ok] drift late (token 7)             -> *** MISS *** decode drifted at token 7
[ok] wrong first token (prefill)      -> *** MISS *** first token [4242], expected [9079]
[ok] empty output                     -> *** MISS *** first token [], expected [9079]
```

The verdict logic is factored into `_paris_verdict()` so it is testable without hardware. The end-to-end `make verify` re-run on NPU2 is queued behind another session's use of the device; the golden sequence itself was captured from hardware runs earlier today.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -85,7 +85,10 @@ jobs:
           source air-venv/bin/activate
           # Temp pin: the LLVM 22 nightly (22.0.0.2026081201+046b11f9) miscompiles
           # xrt/50_multi_launch_attention and xrt/48_triton_matmul_ver4_strix_4x4_bf16_output
-          # on NPU2 (NaN outputs). Revert once fixed upstream.
+          # on NPU2 (NaN outputs). It also takes out every Qwen LLM example --
+          # llms/qwen25_{0_5b,1_5b,3b} (bf16) fail their top-k gate and
+          # llms/qwen25_3b_q4 (Q4_0) returns NaN logits -- so re-check those four
+          # alongside the two xrt tests before reverting. Revert once fixed upstream.
           python3 -m pip install --upgrade --force-reinstall "llvm-aie==21.0.0.2026080601+f4a72c27" -f https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
 
       - name: Build and test mlir-air
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -109,7 +109,10 @@ jobs:
           source air-venv/bin/activate
           # Temp pin: the LLVM 22 nightly (22.0.0.2026081201+046b11f9) miscompiles
           # xrt/50_multi_launch_attention and xrt/48_triton_matmul_ver4_strix_4x4_bf16_output
-          # on NPU2 (NaN outputs). Revert once fixed upstream.
+          # on NPU2 (NaN outputs). It also takes out every Qwen LLM example --
+          # llms/qwen25_{0_5b,1_5b,3b} (bf16) fail their top-k gate and
+          # llms/qwen25_3b_q4 (Q4_0) returns NaN logits -- so re-check those four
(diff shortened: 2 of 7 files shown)


erwei-xilinx:
Good catch, and it was a real bug — not just a theoretical one. Confirmed against the old expression from a directory containing a decoy `bin/clang`:

```
old code resolved to: 'bin/clang' -> stat ok? True
would have fingerprinted: 5:1786576503     # the decoy, not any toolchain
```

So with the env var unset the same toolchain fingerprints differently depending on where the run started — and two unrelated toolchains could collide on one decoy, which is the worse direction since it means a *false match* and a stale cache served silently. That is exactly the failure this fingerprint exists to prevent.

Fixed in 38ae6e95d: check the root before joining and record `"unset"`. Verified cwd-independent with a planted decoy —

```
unset, no decoy : unset / unset
unset, w/ decoy : unset / unset
cwd-independent : YES
real root       : 225824:1786572805
```

— and a real root still stamps the binary.

PR #1792, opened 2026-08-13
[programming_examples] qwen decode handoff: read via bo.map(), not bo.read()

erwei-xilinx:
The `qwen25_3b_q4` verify lit added in #1789 fails on the perf runner:

```
qwen_prefill_to_decode.py:254 in dispatch
    h = np.frombuffer(self.x_bo.read(self.X * 2, 0), dtype=bfloat16)
ValueError: ndarray is not C-contiguous
```

Same root cause as #1788's gemma fix: pyxrt's `read()` returns an array whose stride metadata is build-dependent, and where it comes back non-contiguous the consumer rejects it. Only the consumer differs — `np.frombuffer` raises *"not C-contiguous"* where `.view()` raised *"last axis must be contiguous"* — so the error text does not resemble the earlier one, which is why the grep I did while fixing #1788 (for the `.read().view()` chain) missed it.

Both readbacks in `dispatch()` (`x_bo` and `kv_bo`) move to the `np.frombuffer(bo.map(), count=...)` idiom the llama and gemma designs already use.

The path is not new. #1789 gave this example its first `run_npu2_verify.lit`, so CI exercises it for the first time — new coverage surfacing a latent bug rather than causing one. The prefill half was already passing (`*** PARIS ***`, `argmax=12095`); it is the decode handoff that trips.

## Verification

On NPU2 (Ryzen AI 9 HX 370), full `clean` → `compile-decode` → `verify`, with `FileCheck` run against the lit:

```
[qwen_q4_prefill] *** PARIS ***
[qwen_q4_prefill] first-token argmax=279 (' the')
generated : ' the city. Paris. The tower is called the Eiffel Tower, after'
### exit=0 ###

FileCheck PASSED — all CHECK lines matched in order
```

## Not fixed here

Six other `bo.read()` consumers remain:

- `matrix_vector_multiplication/handwritten_pingpong_reference/run_{la,ld,la_2tile,ld_2tile}.py`
- `matrix_vector_multiplication/int4_q4nx/harness/time_elf{,_runlist}.py`

None are lit-covered, so they are latent in exactly the way this one was — the next to bite when someone adds coverage. A sweep wants its own change, together with a way to actually exercise them; shipping six unverified edits alongside a verified one seemed the wrong trade.

### DIFF
--- a/programming_examples/fused_decode/qwen_prefill_to_decode.py
+++ b/programming_examples/fused_decode/qwen_prefill_to_decode.py
@@ -251,9 +251,10 @@ class QwenFusedDecoder:
         self.x_bo.sync(FR)
         self.kv_bo.sync(FR)
 
-        # Zero-copy view into the BO (the shared infra's readback idiom): bo.read()
-        # returns a buffer whose stride metadata is pyxrt-build dependent, and on
-        # some runners it comes back non-contiguous, which np.frombuffer rejects.
+        # Read through the mapped BO, not bo.read(): read() returns a buffer whose
+        # stride metadata is pyxrt-build dependent, and where it comes back
+        # non-contiguous np.frombuffer rejects it. (frombuffer is a view; the
+        # .astype below is where the copy happens.)
         h = np.frombuffer(self.x_bo.map(), dtype=bfloat16, count=self.X).astype(
             np.float32
         )[m.XO_RMSIN : m.XO_RMSIN + m.K]
--- a/programming_examples/fused_decode/qwen_prefill_to_decode.py
+++ b/programming_examples/fused_decode/qwen_prefill_to_decode.py
@@ -251,13 +251,16 @@ class QwenFusedDecoder:
         self.x_bo.sync(FR)
         self.kv_bo.sync(FR)
 
-        h = np.frombuffer(self.x_bo.read(self.X * 2, 0), dtype=bfloat16).astype(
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Fixed in the commit above. You are right — `np.frombuffer` is a view, but the `.astype(np.float32)` on the very next line copies, so "zero-copy" described something the code does not do.

Reworded to state the actual reason for `map()` over `read()` (the build-dependent stride metadata) and to point at where the copy really is:

```python
# Read through the mapped BO, not bo.read(): read() returns a buffer whose
# stride metadata is pyxrt-build dependent, and where it comes back
# non-contiguous np.frombuffer rejects it. (frombuffer is a view; the
# .astype below is where the copy happens.)
```

Worth flagging that the wording was inherited rather than invented here — the same "Zero-copy view into the BO" line sits in three siblings:

- `llms/gemma3_4b_q4nx/gemma3_4b_q4nx_inference.py:312`
- `llms/llama32_1b_q4nx/llama32_1b_q4nx_inference.py:463`
- `llms/llama32_3b_q4nx/q4nx_decode_3b.py:325`

All three are followed by `.astype(...)` too, so the same criticism applies to each. I have left them alone rather than touch three unrelated files in a bug-fix PR — happy to align them in a separate comment-only pass if you want the tree consistent.

PR #1795, opened 2026-08-13
[AIR] Derive forwarding-hop packet ids instead of hand-pinning them

erwei-xilinx:
## Problem

A packet `air.channel` gets one auto-assigned id per symbol in `air-to-aie`, so a channel whose destination is chosen **per packet** (a demux) can only be expressed by hand, via `packet_ids`. The q4nx/q4 decode designs pinned it on three channels each — the demux plus the two hops that forward into it — **12 declarations across 4 models**. The values were also duplicated in the kernel that stamps the header, with nothing checking the two agreed; a divergence is invisible until the device hangs.

## What this adds

`air-annotate-packet-ids`, in `AIRMiscPasses.cpp` at the same early pipeline slot as `air-annotate-refeed` and `air-annotate-append-barrier`, and for the same reason those two sit there: read the shape the front end wrote, before another pass rewrites the channel ops.

**Classification is by volume.** A broadcast REPLICATES — every destination receives the whole stream, so each destination's volume equals the put volume. A demux PARTITIONS — the destinations' volumes sum to it.

| Condition | Meaning | ids |
|---|---|---|
| `∀d: vol(d) == vol(put)` | replication | 1 |
| `Σ_d vol(d) == vol(put)`, N dests | partition | N |
| neither | unknown — reported, never guessed |  |

Only the broadcast dimension counts as a fan-out: `air.channel @c [4]` is four parallel flows, not a 4-way demux. Volume is gated on header ownership, since a DMA-stamped channel is one BD with one id and cannot route on payload content.

**Back-propagation.** A single-destination hop that forwards packets a downstream demux will route must admit every id that demux keys on, or they are filtered out mid-route. Those chains are plain SSA — a get lands a memref, a put on the next channel sends it onward — so the id list propagates backwards to a fixpoint.

**Contract check.** Mark the header-writing call and the constants reaching that operand are compared with the channel's `packet_ids`:

```mlir
func.call @proj_qmm_flush_hdr(%acc, %buf, %id) {
  air.pkt_header_channel = @outA, air.pkt_header_operand = 2 : i32
} : (memref<32xf32, 2>, memref<48xbf16, 2>, i32) -> ()
```

A proven disagreement is an error; when either side is not statically known the pass stays silent. Folded into the same pass rather than a separate verifier because it is the same information — the propagation already traces those constants, so the check is a comparison, not an analysis.

## What it deliberately does not do

It does not choose id **values**. Two ways were tried and both broke a design on device:

- **Inventing fresh ids** renumbers every other packet channel (`air-to-aie` hands out the lowest free ids) and perturbs a tuned floorplan.
- **Relabeling the kernel constants positionally** assumes the kernel's arm order matches the declared list. It is not guaranteed — on one design the arm order is `[4,1,8]` against a declared `[1,4,8]`, so it swapped two ids and sent a phase's packets to the wrong tile.

The pass only ever COPIES a list a channel already declares onto the hops that forward it — the part that carries no ordering information.

## Outcome

The demux keeps its declaration: `air-to-aie` routes destination `i` with `packet_ids[i]`, and that ORDER is not recoverable from the IR. Deriving it too would need a surface binding each id to its destination.

The forwarding hops do not, and are now derived — **the four designs declare `packet_ids` on one channel each instead of three.**

## Validation

- Post-pass channel declarations are **identical** to the previously hand-pinned IR for all four designs, with **zero** kernel constants rewritten.
- All four designs pass their end-to-end gates on NPU2.
- 5 new lit tests: broadcast, demux, partition-without-header-ownership, bundle-is-not-a-demux, the forwarding chain, a non-forwarding re-stamp hop, contract agreement/violation/unresolvable.
- `check-air-mlir` otherwise unchanged. `aircc/debug_ir.mlir` updated for the inserted pipeline slot.


erwei-xilinx:
Addressed in 3ae983ec5. Two of the findings were real bugs, not just naming.

**1. `kernelWritesHeader()` diverged from `AIRToAIE::channelKernelWritesHeader()`** — correct, and worse than described. It checked `air.strip_pkt_header` (which no builder sets and no pass reads — it survives only in stale dumps) *and* omitted the `packet_ids.size() > 1` case. So a channel shaped like `air.channel`'s own documented example — `packet_ids = [1,4]` with no `keep_pkt_header` — would be classified DMA-stamped, and demux classification plus back-propagation skipped entirely.

Rather than sync the copy, there is now **one** function: `air::channelKernelWritesHeader`, exposed from the dialect and used by both. Two copies drifted apart once already; a shared definition is the same treatment `copyChannelSteeringAttrs` already gets for the marker set.

**2. Destination keying used the whole index tuple** — also correct. `specializeBroadcastShape` keeps `getBroadcastDimension()` at its broadcast extent and pins every other dimension to 1, so the remaining indices select a *bundle instance*, not a destination. A `[2,1]` bundle with `broadcast_shape = [2,2]` looked like four destinations instead of two. Now keyed on the broadcast dimension's coordinate; a broadcast channel whose dimension cannot be resolved is reported unclassifiable rather than guessed at.

Our own `outY` is `[1,1]` with `broadcast_shape = [1,3]`, so the full tuple happened to equal the broadcast coordinate and it worked by accident — exactly the kind of thing that ships and then bites on the next design.

**3. Stale pass names** in both builders' comments — fixed. `air-infer-packet-ids` and `air-verify-packet-id-contract` were merged into `air-annotate-packet-ids` and I missed the Python side.

Both fixes have lit coverage: a channel whose ids alone imply header ownership, and a bundle that also broadcasts (the latter reports 2 destinations where the previous keying reported 4).

Re-validated: the four decode designs' derived channel declarations are still byte-identical to the hand-pinned IR with zero disagreements, `check-air-mlir` unchanged, and all four pass end-to-end on NPU2 (llama-1b, llama-3b, gemma `*** PARIS ***` 17.03 tok/s, qwen `*** PARIS ***` 12.13 tok/s).

PR #1797, opened 2026-08-13
[AIR] Keep the shim bucket column instead of re-deriving it downstream

erwei-xilinx:
`ShimDMAAllocator` groups every L3 flow into a bucket keyed on a column — the far-side column when known, else derived by walking a memtile LTO to its downstream cores (`effectiveColForMemTileLTO`). It then **discarded that column**: the shim `LogicalTileOp` was emitted col-less unless the channel carried an explicit `air.shim_col`, leaving `aie-place-tiles` to re-derive a column from its own centroid.

The two can disagree. Measured on the 1B decode, a weight feed bucketed on **col 1** was independently placed on **col 3** — a cross-column flow that the front end then had to pin back by hand.

So `air.shim_col` was not supplying information the compiler lacked. It was stopping the compiler from throwing away information it had already computed.

### Fix

Stamp the bucket column on the LTO. `bucketCol` already equals `shimColPin` when a pin is present, so the pin branch collapses out — 8 insertions, 2 deletions in the pass.

This is safe by construction on the placer side: under `SequentialPlacer` the column is a **preference, not a constraint** — `findTileWithCapacity` ranks candidates by `(|col − targetCol|, load, col, row)`, so a saturated column drifts to a neighbour rather than failing. Only `AIESAPlacer` enforces `satisfiesConstraints`, and AIR does not drive it.

### 38 of 50 pins removed — measured, not assumed

I compiled each of the four q4nx decode designs twice, with pins honoured and with pins ignored, and diffed per-flow shim placement. All four give the same answer: **exactly three pins per model change anything.**

| model | load-bearing pins | what happens without them |
|---|---|---|
| 1b / 3b / gemma | `appendK`, `appendV`, `layerOut` | pinned to cols 3/4/5; all three collapse onto `2:S2MM0` — same tile *and* same channel |
| qwen | `appendK`, `appendV`, `ropeLUT` | `appendK`/`appendV` 7→1, `ropeLUT` 0→1 |

Those stay. Every `inW*`, `inKV_*`, `rms*` and qwen's `layerOut` now lands on exactly the column it used to be pinned to, with no attribute.

The residue is deliberate readback spreading, which the compiler has no rule for. That is a separate problem from this one.

### Test updates

Five `AIRToAIE` tests change because shim LTOs now print a concrete column instead of `(?, ?)`. Each becomes a **stronger** assertion rather than a weaker one:

- `good_shim_packet_flow_npu_4col` now asserts the channel-*k*-to-column-*k* map that its prose already claimed ("preserving the 1-shim-per-compute-col placement").
- `shim_pkt_l2_l1_col_bucket_merge` now asserts the `col=0` that its own comment describes ("both land in bucket col=0").
- `air_shimcpy_to_aie` asserts the shim follows its compute tile's column — col 5 in the `aie.tile(5, 4)` section, col 2 elsewhere.

### Verification (NPU2, CI toolchain)

- **Gate** — stripped drivers + this compiler vs stock: **0 shim-placement diffs, 0 hw-op diffs** on all four designs. The 34–40 residual lines are only the removed attribute text on the channel declarations.
- `check-air-mlir` **487/487**
- `conv2d_14x14` PASS, `xrt/48` PASS
- `xrt/50` **4/4 PASS at 218 buffers / 292 locks**, matching stock (this test's compiler output is nondeterministic run to run, so the comparison is order-insensitive counts over repeated runs)
- 1b PASS, 3b PASS, gemma PARIS 16.91 tok/s, qwen PARIS argmax 12095 12.13 tok/s
- All **7 production decode `insts.bin` byte-identical** to the pre-change builds

### Note on blast radius

This changes shim placement for every design, not just the q4nx ones — previously they all got col-less LTOs and relied purely on the centroid. That is the point: AIR's bucketing decision and the placer's placement decision now agree instead of being allowed to diverge. The regression net above is chosen to cover that (`conv2d_14x14`, `xrt/48`, `xrt/50` all exercise unpinned designs).

### DIFF
--- a/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
+++ b/mlir/lib/Conversion/AIRToAIESchedulingUtils.cpp
@@ -2019,8 +2019,14 @@ FailureOr<air::allocation_info_t> air::ShimDMAAllocator::allocNewDmaChannel(
     }
     tileLT = AIE::LogicalTileOp::create(
         b, device.getLoc(), AIE::AIETileType::ShimNOCTile,
-        /*col=*/shimColPin >= 0 ? b.getI32IntegerAttr(shimColPin)
-                                : IntegerAttr(),
+        // Stamp the bucket column (already == shimColPin when pinned). This is
+        // the column the allocator grouped the flow into; emitting the LTO
+        // col-less discards it and lets aie-place-tiles re-derive one from its
+        // own centroid, which can disagree -- a weight feed bucketed on col 1
+        // was independently placed on col 3, giving a cross-column flow. Under
+        // SequentialPlacer the col is a preference, not a constraint
+        // (findTileWithCapacity ranks by distance), so a full column drifts.
+        /*col=*/bucketCol >= 0 ? b.getI32IntegerAttr(bucketCol) : IntegerAttr(),
         /*row=*/IntegerAttr(),
         /*allocation_scheme=*/StringAttr());
   }
--- a/mlir/test/Conversion/AIRToAIE/air_multi_launch_to_multi_device.mlir
+++ b/mlir/test/Conversion/AIRToAIE/air_multi_launch_to_multi_device.mlir
@@ -15,7 +15,7 @@
 // AIR emits a ShimNOCTile LTO with column hint 0; compute tile is placed
 // directly. The downstream aie-place-tiles pass resolves the LTO.
 // CHECK: aie.device(npu2) @add_three
-// CHECK-DAG:   %[[SHIM3:.*]] = aie.logical_tile<ShimNOCTile>(?, ?)
+// CHECK-DAG:   %[[SHIM3:.*]] = aie.logical_tile<ShimNOCTile>(0, ?)
 // CHECK-DAG:   %[[TILE3:.*]] = aie.tile(0, 2)
 // CHECK:   aie.lock(%[[TILE3]]
 // CHECK:   aie.buffer(%[[TILE3]])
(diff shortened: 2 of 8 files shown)


erwei-xilinx:
Closing in favour of Xilinx/mlir-aie#3551, which fixes this one layer down.

This PR had AIR stamp the shim bucket column onto the `LogicalTileOp` so `aie-place-tiles` would stop re-deriving a different one. That works, but it is the wrong layer, and reviewer feedback that it amounts to a second shim placer in AIR was correct.

Running down *why* the two disagreed found a real bug in the mlir-aie placer. `computeCentroidColumn`'s peer resolution accepted only `CoreTile` peers, so a memtile with an already-known column was discarded and the search fell through to that memtile's downstream cores. A memtile typically serves several unrelated flows, so that returns the union of every column it touches:

```
LTO[air_inW0c1, air_inW1c1]
  dests = {0 0 0 0  1 1 1 1  6 6 6 6  7 7 7 7  1 1 1 1  2 2 2  5}
  centroid=3  ->  placed (3,0)
```

The destination memtile was pinned to col 1 the whole time. Mean 2.96 gave col 3.

With Xilinx/mlir-aie#3551 the same LTO resolves `dests={1}` and lands on col 1 — reproducing everything this PR achieved, with **no AIR change at all**. Keeping the AIR stamp would also have masked the mlir-aie bug for every AIR-generated design and left any future centroid improvement dead behind `tryGetCol()`.

The 38 redundant `air.shim_col` pins still come out, but as a separate change once #3551 lands and the mlir-aie pin is bumped. Verified against the patched toolchain: 0 shim-placement diffs and 0 hw-op diffs on all four Q4NX decode designs, all 7 production decode `insts.bin` byte-identical, and 1b/3b/gemma/qwen PASS on NPU2.

PR #1798, opened 2026-08-13
Bump mlir-aie to 1.4.2.dev8+gdd18668

erwei-xilinx:
Routine pin refresh, eleven mlir-aie commits (`f501f21..dd18668`). No AIR source change is required — the tree builds clean against the new wheel as-is, which is the main thing this PR is here to prove.

LLVM is unchanged (`56bcc1871734`, DATETIME `2026080106`), so `clone-llvm.sh` needs no edit. Both `latest-wheels-4` and `latest-wheels-no-rtti-2` publish `1.4.2.dev8+gdd18668`, so the no-RTTI leg of the wheel matrix is covered.

Note the series moves 1.4.1 -> 1.4.2.

## What this pulls in

Nothing here targets a known AIR failure — this is a keep-current bump, not a fix. The functional changes:

- **mlir-aie #3547** — lower `aievec.neg` to LLVM for AIE2 and AIE2p.
- **mlir-aie #3538** — BD iteration support on `aie.dma_bd`.
- **mlir-aie #3532** (aie-rt) — zero the `p_memsz - p_filesz` gap when loading `PT_LOAD` segments, so `.bss` is actually cleared on ELF load.
- **mlir-aie #3530** — aie-rt pin moved to `8849e208` (`release/main_aig`), now tracked by Dependabot.
- **mlir-aie #3536** — hardware coverage for same-packet-id routing on npu1.

The rest are CI/infra and dependency bumps (#3518 clang-tidy, #3544, #3543, #3541, #3506, #3534).

## Validation

Built and tested locally against the published wheel, Phoenix/npu1:

| Suite | Result |
| --- | --- |
| `ninja` | clean, no API breakage |
| `check-air-mlir` | 492 passed, 7 XFAIL, 7 unsupported |
| `check-air-cpp` | 1/1 |
| `check-air-python` | 13 passed, 8 unsupported |
| `test/xrt` npu1+peano | 4/4 — `01_air_to_npu`, `05_extern_func`, `08_gemm_extern_vec`, `13_conv2d_i32` |
| `programming_examples` npu1+peano | 3/3 — `softmax`, `herd_dataflow` (mlir + python source) |

### Draft because the hardware coverage is a smoke set, not a sweep

The npu1 runs above are seven hand-picked tests, not the full 43-test `ryzen_ai_npu1 && peano` suite — the local Phoenix host is shared, so I kept the device footprint small. Nothing in this bump is npu2-specific, but nothing here has been run on npu2 either, and #3538 (`dma_bd` iteration) and #3532 (`PT_LOAD` zero-fill) are the two commits most likely to surface something a smoke set would miss. Leaving this as a draft until CI has run the full matrix.

### DIFF
--- a/utils/clone-mlir-aie.sh
+++ b/utils/clone-mlir-aie.sh
@@ -14,8 +14,8 @@
 #
 ##===----------------------------------------------------------------------===##
 
-export HASH=f501f215b5e772b7fba38df35b79251d1de64b97
-WHEEL_VERSION=1.4.1.dev58+gf501f21
+export HASH=dd186682c7f5fcd49d42305f2abf6919702c4e1c
+WHEEL_VERSION=1.4.2.dev8+gdd18668
 
 if [ x"$1" == x--get-wheel-version ]; then
   echo $WHEEL_VERSION


erwei-xilinx:
## The amd8845hs failure is pre-existing on `main`, not from this bump

`Build and Test with AIE tools on Ryzen AI (amd8845hs)` fails here on two tests:

```
FAIL: AIRPYTHON :: torch_mlir_e2e/relu.py (19 of 21)
FAIL: AIRPYTHON :: torch_mlir_e2e/mul.py  (20 of 21)
RuntimeError: DRM_IOCTL_AMDXDNA_EXEC_CMD IOCTL failed (err=-5): Input/output error
```

`main` at `d035da95` — the exact merge base of this PR — fails the **same job, same two tests, same error** ([run 31740851383](https://github.com/Xilinx/mlir-air/actions/runs/31740851383)). So this is not a regression introduced by the pin.

It has been red on every `buildAndTestRyzenAI` run since 2026-08-12 21:20:

| Run | Commit | amd8845hs |
| --- | --- | --- |
| [31631610421](https://github.com/Xilinx/mlir-air/actions/runs/31631610421) | `b40eb7a1` | last green (17/21) |
| [31641854793](https://github.com/Xilinx/mlir-air/actions/runs/31641854793) | `a75ea2e8` | first red — same 2 tests, same IOCTL error |
| … 11 consecutive failures … | | |
| [31740851383](https://github.com/Xilinx/mlir-air/actions/runs/31740851383) | `d035da95` | red (merge base) |

The `amdhx370` (npu2) leg is green both here and on `main`.

### This looks like a wedged device on the runner, not a code defect

`DRM_IOCTL_AMDXDNA_EXEC_CMD ... (err=-5)` is the signature of a Phoenix NPU whose stream network is stalled: the wedge is device/firmware state, so it outlives the process and contaminates every subsequent hardware test until the driver is reloaded. Consistent with the pattern here — the failures are the *last* hardware tests in the run, the same two every time, and they went red mid-day without a corresponding code change to that path.

Suggested fix on the runner: `sudo modprobe -r amdxdna && sudo modprobe amdxdna`. Worth confirming before anyone bisects `a75ea2e8` (#1786), which touches decode shim feeds and has no obvious connection to `torch_mlir_e2e/relu` or `mul`.

### Local npu1 result for the same suite

On my local Phoenix against this exact wheel, `check-air-python` is 13 passed / 8 unsupported — the two `torch_mlir_e2e` tests are among the unsupported ones locally (no torch-mlir in that venv), so my earlier smoke run did not cover them. I can't independently clear them; they need a healthy runner.

erwei-xilinx:
Superseded by #1803, which bumps to `c84915b0` / `1.4.2.dev14+gc84915b`.

`dd18668` is an ancestor of `c84915b0` (`compare` → `ahead_by=6, behind_by=0`), so #1803 contains everything this PR pulls in — including mlir-aie#3547 (`aievec.neg` → LLVM), #3538, #3532, #3530 and #3536 — plus six more:

```
efd45c2ca [aiecc] Stop stripping GEP nuw when downgrading for Peano (#3540)
a9e39faeb [aievec] Don't vectorize scalar arith inside runtime_sequence (#3550)
065cfd93d Make ObjectFIFO fully dynamic, add new unrolling pass (#3298)
a5b47887e Remove Lock ID Allocation out of ObjectFIFO pass (#3377)
3ca0193ce [hrx] Retain executable per kernel handle (#3545)
c84915b06 [AIE] Let a non-core peer's known column decide the centroid (#3551)
```

The last one is the reason #1803 exists rather than being another keep-current bump: it lets the tile placer derive shim columns that the Q4NX decode designs were pinning by hand, so #1803 also removes 38 `air.shim_col` attributes.

One thing this PR has that #1803 does not: **npu1 (Phoenix) validation.** #1803 was validated on npu2 only — no npu1 hardware available locally — so npu1 coverage there rests on CI's `amd8845hs` job. Worth a look at that job on #1803 before merging, keeping in mind it has been failing on `main` independently (`torch_mlir_e2e/relu.py` and `mul.py`, `DRM_IOCTL_AMDXDNA_EXEC_CMD (err=-5)`).

Also note #3547 landing makes the `arith.negf` / `aievec.neg` entry in `CLAUDE.md` ("no LLVM lowering, use `arith.subf(zero, x)` instead") stale once either bump lands.

PR #1799, opened 2026-08-13
[AIR] Derive a packet demux's routing in the compiler, and keep the BD ring in step with its flows

erwei-xilinx:
Follows #1795, which took the four q4nx/q4 decode designs from twelve `packet_ids`
declarations down to four. This removes the last four **and** the hand-written
kernel call that stamped the routing header, so nothing in a design names a wire
packet id any more. Three BD-ring defects found on the way are fixed too.

### Routing a packet demux was a contract written twice

Each design declared an ordered `packet_ids` list on its egress demux and
separately hardcoded the same numbers in the compute kernel, which stamped one
into the payload header to choose a destination. One contract, two languages,
nothing tying the halves together — a mismatch routed a packet to the wrong
consumer and hung the device. A checker existed to catch the disagreement, which
is a fair summary of the problem.

Both halves are now the compiler's. A put states which *destination* a packet is
for, and that is the whole design input:

```mlir
air.channel.put @outA[%c0, %c0] (%buf[%c14] [%c66] [%c1]) dest(%d)
```

`dest` is the index along the channel's broadcast dimension — the same index the
matching get sits at, the numbering the design already uses. From it,
`air-annotate-packet-ids` allocates one id per destination, maps the runtime
destination to its id, and emits the store. Where the header goes was never a
choice: the switchbox reads the first word of what arrives and the DMA sends from
`offsets[0]`, so the header *is* `offsets[0]`. With one spelling there is nothing
to disagree, and the contract checker
(`air.pkt_header_channel` / `air.pkt_header_operand`) is deleted along with
`proj_qmm_write_hdr`.

**No design change and no BD change** — the egress still emits one put per round
from one loop, memtile occupancy is identical at 22/48 peak, and the lowered IR
matches the hand-written build exactly (921 locks, 43 packet flows).

### Three things this got wrong first, each of which hung the device

- **Allocate downward from `kMaxPacketID`.** Claiming the low end pre-empts
  `air-to-aie`'s upward auto-assignment and renumbers every other packet flow.
  Taking 31/30/29 leaves the low end untouched, so adding a demux no longer
  perturbs a tuned floorplan.
- **The header store must share a lock section with the payload write.** The lock
  placer brackets each buffer-touching op on its own — deliberately, so several
  buffers can ping-pong in one loop body — so a store emitted *beside* the payload
  write gets a section of its own: the buffer reaches the DMA with no header in
  it, and the consumer lock is signalled twice per production. On device that is a
  timeout, not corruption (`use_lock` 921 → 953, exactly one extra pair per
  store). Adjacency is not enough either, because at lock time the two ops sit in
  different `air.execute` regions. The store goes *inside* the region that
  produced the payload.
- **That region does not exist in the early pass slot.** So the pass is split:
  `emit-headers=false` allocates ids up front, where the shape the front end wrote
  is still intact, and a second run just after `air-dependency` writes the
  headers. Allocation is idempotent — the second run reads back what the first
  wrote.

One supporting fix outside the pass: `air-dependency` re-instantiates every
channel op with an async token, and that builder does not take the optional
operand, so `dest` was dropped silently — the pass then found nothing to do and
no header was written at all. `dest_survives_dependency.mlir` pins it.

The destination lookup is a compare-and-select per destination in general, but
allocated ids are a descending run, so it collapses to one multiply-add. An
explicit pin need not be affine and keeps the general form.

### Three BD-ring defects, all silent

Found while establishing whether the ids were derivable at all. Each emitted a
design that mis-routes and hangs, with no diagnostic.

- **BD equivalence ignored the channel.** `chansMappedToEquivalentBDs` compared
  the memref and offsets/sizes/strides but not the channel declaration, so puts
  on different channels with the same access pattern folded onto one BD. The
  flows were still emitted, but only the surviving BD's id went on the wire.
  `chansPartOfSameRotation` right below it always keyed on the declaration.
- **BD tasks were bucketed by trip count alone.** Two unrelated loops that
  happened to iterate the same number of times became one task, running together
  at the earlier one's position: `[A x4, B x32, C x4]` emitted as
  `[A x4 + C x4, B x32]`. Only transfers adjacent in program order merge now, and
  the result is a vector rather than a map so a repeat count may recur. Chains
  whose transfers all share one count still collapse to a single task, leaving
  infinite-BD-loop mode unchanged.
- **Choosing a destination with a branch is now rejected.** A ring advances one
  BD per transfer whatever the core branched to, so staying in step needs the arm
  sequence to cycle in lockstep with the ring, which nothing in the IR states.
  `verifyMM2SChains` mirrors `repairS2MMChains` and shares its diagnosis, renamed
  `diagnoseBDChain` since the invariant is direction-neutral. Nothing to repair on
  this side: moving a producer to another MM2S changes which port its flows leave
  from.

The two directions are close but not symmetric, so the shared check takes a flag.
An arm that transmits nothing is a hole for a consumer — the packet still arrives
and lands on whatever BD the pointer sits on — but not for a producer, which
simply does not advance the ring. Rejecting that shape would have failed the
KV-cache append in the fused decode designs.

### Testing

Five new lit tests cover the ring order, the BD folding, the rejected
branch-selected destination, the empty producer arm that must still compile, and
`dest` surviving `air-dependency`; `assign.mlir` gains the header-emission and
in-region-placement cases; the two contract-check tests are deleted with the
checker. `check-air-mlir` shows no new failures.

All four models rebuilt from clean and re-verified on device (NPU2), with zero
skipped builds:

| model | correctness | throughput |
|---|---|---|
| llama32_1b_q4nx | PASS (top-k 2/2) | 48.70 / 48.66 / 48.65 tok/s |
| llama32_3b_q4nx | PASS (top-k 2/2) | 22.3 tok/s |
| gemma3_4b_q4nx | PASS (Paris + coherent text) | 17.15 tok/s |
| qwen25_3b_q4 | PASS (coherent text) | 12.12 tok/s |

llama-1b was A/B'd against this branch's base before the merge with main, rebuilt
and measured back-to-back on an idle box, three runs per side: **47.26 / 47.12 /
47.82** before versus **47.74 / 47.79 / 48.00** after — flat. The higher figures
in the table are post-merge and pick up `PROJ_RC_CACHE` from main.


erwei-xilinx:
Addressed in 62047ef67, and rebased onto main via a merge (915a3467e) to clear the conflict in `fused_decode.py`.

All six review points were written against the state before 6636e2063, which replaced the mechanism they describe: the kernel no longer stamps anything, the put names its destination (`dest(%d)`) and the compiler writes the routing header.

- **Stale id-contract comments** (fused_decode.py:1124, fused_decode_qwen.py:691) — right, and there were four more of the same kind. All six now say the design names a destination and the compiler stores the id; the reference's 1/4/8 are called out as the pinned spelling this replaced, so nobody reads them as live wire values. `DISTINCT_IDS` itself is already gone.
- **`repeat_counts` note at AIRToAIEPass.cpp:6332** (suppressed) — correct and still live. It called `repeat_counts` a map keyed on unique repeat values; it is a program-ordered vector and a count may recur. Reworded, including why a single entry still means infinite-BD-loop mode.
- **Grammar, "the receiving gets use"** — four occurrences, now "the receiving gets sit at".

The merge conflict was one hunk: main's `PROJ_RC_CACHE` block landed next to this PR's `pktv` type change (i32 packet id to index destination ordinal). Both kept.

Re-verified on device after the merge, since `PROJ_RC_CACHE` is a functional change: llama-1b top-k 2/2 PASS, 48.70 / 48.66 / 48.65 tok/s. Lowered IR still shows 43 packet flows and 32 compiler-emitted header stores; the lock count moves 921 to 933, which is main's new cache buffer, not this PR.

PR #1804, opened 2026-08-14
[test] Pin the shared-L1 one-section invariant for adjacent writes

erwei-xilinx:
Two back-to-back writes to one shared L1 buffer must take **one** lock section. Nothing checked that until now.

`allocateSharedL1BufferLocks` brackets each buffer-touching op on its own. That is deliberate: several buffers ping-ponging in a single loop body each need their own acquire around their own access, and hoisting to the loop boundary would serialize ping against pong. But applied literally to two writes that are *adjacent*, it releases the buffer to the DMA after the first one — before the second has happened — and signals the consumer lock twice per production.

The failure mode is a hang, not corruption: the lock counts drift by one per production and the core and the DMA fall out of step. It cost a day to find, on a compiler-emitted packet routing header sitting next to the payload write it belongs with (#1799). The fix for that shipped as the `adjacentRuns` helper, which merges maximal runs of immediate-successor accesses into a single section; the invariant it establishes was only ever verified by counting `aie.use_lock` by hand on a model build — a 15-minute device gate standing in for a lit run.

This adds the lit test. Confirmed load-bearing: with `adjacentRuns` disabled the test fails, with it the test passes. Writes separated by other work are untouched by the merge and by this test.

Test-only; no compiler change.

### DIFF
--- a/mlir/test/Conversion/AIRToAIE/shared_l1_adjacent_writes_one_section.mlir
+++ b/mlir/test/Conversion/AIRToAIE/shared_l1_adjacent_writes_one_section.mlir
@@ -1,0 +1,78 @@
+//===- shared_l1_adjacent_writes_one_section.mlir --------------*- MLIR -*-===//
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// Two BACK-TO-BACK writes to one shared L1 buffer take ONE lock section.
+//
+// The shared-L1 placer brackets each buffer-touching op on its own, which is
+// deliberate: several buffers ping-ponging in one loop body each need their own
+// acquire around their own access, and hoisting to the loop boundary would
+// serialize ping against pong. But that rule, applied literally to two writes
+// that are ADJACENT, releases the buffer to the DMA after the first one --
+// before the second has happened -- and signals the consumer lock twice per
+// production.
+//
+// That is not a corruption bug, it is a hang: the lock counts drift by one per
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Addressed in b15542db8.

The unanchored `CHECK: aie.core` was a real fragility: `@main` is pinned at `y_loc=2` and `@helper` at `y_loc=3`, but nothing in the CHECK lines said which core was meant, and `@helper`'s core opens with the same acquire-then-call pair. The test passed only because `core_0_2` happens to be emitted first — had that order flipped, the match would have landed on `@helper` and failed at the `vector.store`, i.e. a spurious failure indistinguishable from the adjacency merge actually regressing.

Now captures the owner tile and anchors both the lock checks and the core to it. Verified discriminating: repointing the anchor at `tile(0, 3)` makes FileCheck fail at the first lock check rather than pass.

erwei-xilinx:
Addressed in 40945c2e9.

The `CHECK-NEXT` pair made the test stricter than the invariant it exists to pin. Loop plumbing or a scheduling change could legitimately print something harmless between the two writes without touching the lock section, and the test would fail as though the adjacency merge had regressed — which is the one signal it must not fake.

Replaced with a `CHECK-NOT: aie.use_lock` in each gap, so the assertion is exactly the property: no lock op anywhere between the acquire and the release, and nothing claimed about what else is printed.

Confirmed still discriminating in both directions: repointing the tile anchor at `tile(0, 3)` fails, and inserting a lock-taking op between the two writes fails.

(The second review comment, on anchoring `aie.core` to the main herd's tile, was posted at 21:14 UTC against the original commit — that was already fixed in b15542db8 at 21:19.)

PR #1805, opened 2026-08-14
[AIR] Recover packet routing domains as an analysis, and reject the broken ones

erwei-xilinx:
`air-annotate-packet-ids` recovered "which channels carry the same routing header" **four separate times**, each an ad-hoc loop over a `feeds` map keyed on exact memref SSA identity. The algorithm was right. What it was built on was not, and three of its failure modes were silent.

## What was broken

Probed against `air-opt` on merged main, not inferred:

| shape | before | after |
|---|---|---|
| hop link crossing a `memref.subview` | **silent** — hop gets no `packet_ids`, no diagnostic → mid-route filtering → device timeout | link resolved |
| 3-hop chain, *middle* link aliased | **silent**, and the originator's own link being intact means nothing fires at all | link resolved |
| compute between the get and the put | hop **wrongly inherits** the demux's ids (the payload was overwritten, header and all) | not a hop |
| `dest` put with no demux in its domain | errors, but blames the put's own channel for "not being a demux" — when that channel is a *hop* and the fault is downstream | names the domain and where the chain stopped |

## What this does

One `PacketRoutingDomainAnalysis` in `Util`, consumed by all four sites:

- **Links resolve through `air::resolveBufferRoot`** — walks up through view-like ops, `air.execute` terminators and hierarchy arg mappings. Its downward companion `collectBufferAliases` moves out of `AIRToAIEPass` so the lock placer and this agree on what aliases what. (It also gains a null guard: inserting a null `Value` into a `SmallDenseSet<Value>` collides with DenseMap's empty-key sentinel.)
- **Forward vs transform** — a get/put pair is a forward only if nothing between them rewrites the buffer. Channel gets are exempt: several gathering disjoint slices into one buffer before a single put is the ordinary shape of a memtile hop, and counting them as writes would sever exactly the chains being followed.
- **Ids belong to the domain**, allocated once and shared by the demux and every hop, rather than propagated channel by channel.

And the remaining faults become compile errors:

- A demux whose put reads L2/L3 with nothing upstream. A memtile is a pure data mover — it forwards bytes it did not produce and cannot have chosen where they go, which is the entire reason the decision travels in the header. Exempted when the design declares its own `packet_ids`: an explicit list is an assertion, and this fault is about a list the compiler *derived*.
- A channel forwarding into two domains, and a pin disagreeing with its domain — previously last-writer-wins.

Structural checks run only under `assign`. Without it the pass is a reporter, and a reduced module that stops short of the demux's upstream is describing a classification, not proposing a design.

Also renames `channelKernelWritesHeader` → `channelSourceWritesHeader`: who writes the header is a source-side property, and the old name invited confusion with the destination-side `keep_pkt_header`.

## Not an NFC

Fixing the subview and transform cases *is* a behavior change, so this is not labelled NFC. The gate is that it is inert on the designs that exist today.

## Verification

- **Byte-identical IR on both decode builders.** Built an `air-opt` from `origin/main` and diffed the full output of the real two-run sequence (`annotate{emit-headers=false}` → `air-dependency` → `annotate`) against this branch:
  - `fused_decode.py` (llama-1b, llama-3b, gemma): identical, 3124 lines
  - `fused_decode_qwen.py` (qwen): identical, 1671 lines
  - assigned ids unchanged in both: `@outA` / `@toMain`|`@toHub` / `@outY` all `[31, 30, 29]`
- **Device gate, llama-1b**: `make clean` first, `SKIPS=0`, `verify` → PASS (`topk_passed: 2, topk_failed: 0`).
- `check-air-mlir`: all packet-id tests pass. The only failures in the tree are 4 pre-existing `Util/Channel` tests that fail compiling a C++ host harness on a missing `air_tensor.h`, unrelated to this change.

No perf A/B: the generated IR is byte-identical, so a measurement would report noise, not signal.

New lit coverage: `hop_aliases.mlir` (subview hop, middle-link alias, gather-then-forward, herd block arg) and `no_header_source.mlir` (each new diagnostic, plus the two shapes that must stay quiet).

### DIFF
--- a/mlir/include/air/Dialect/AIR/AIRDialect.h
+++ b/mlir/include/air/Dialect/AIR/AIRDialect.h
@@ -172,7 +172,7 @@ namespace air {
 // Single source of truth. Both air-to-aie (which must not stamp such a BD) and
 // air-annotate-packet-ids (which uses it to gate demux classification) read
 // this; two copies drifted apart once already.
-bool channelKernelWritesHeader(ChannelOp chanOp);
+bool channelSourceWritesHeader(ChannelOp chanOp);
 } // namespace air
 } // namespace xilinx
 
--- a/mlir/include/air/Util/PacketRoutingDomain.h
+++ b/mlir/include/air/Util/PacketRoutingDomain.h
@@ -1,0 +1,152 @@
+//===- PacketRoutingDomain.h ------------------------------------*- C++ -*-===//
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef AIR_UTIL_PACKET_ROUTING_DOMAIN_H
+#define AIR_UTIL_PACKET_ROUTING_DOMAIN_H
+
+#include "air/Dialect/AIR/AIRDialect.h"
+
+#include "mlir/IR/BuiltinOps.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/MapVector.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <string>
+
+namespace xilinx {
+namespace air {
+
+//===----------------------------------------------------------------------===//
+// Per-channel facts
+//===----------------------------------------------------------------------===//
+
+/// How one packet channel fans out.
+enum class PacketFanout {
+  Broadcast,  // every destination receives every packet -> 1 id
+  Demux,      // the destinations partition the stream -> one id per destination
+  SingleDest, // nothing to decide
+  Unknown,    // not statically classifiable; say so, never guess
+};
+
+StringRef getPacketFanoutName(PacketFanout f);
+
+struct PacketChannelFacts {
+  PacketFanout fanout = PacketFanout::Unknown;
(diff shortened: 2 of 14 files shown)


erwei-xilinx:
Addressed in 058b8450e.

**Pinned ids compared by value, not count** — this was the substantive one. A hop pinning `{1,2}` for a demux routing `{13,17}` had the right cardinality and entirely the wrong routes: its switchbox admits two ids, neither of which the demux keys on, so every packet is filtered out at the hop and the demux never fires. Counting alone waved it through, and a pin agreeing with itself is the precise shape of "two spellings, no link" that this analysis exists to remove. Now compares as sets and reports both lists.

Order is deliberately still not compared, and the diagnostic says so: a hop is single-destination, so air-to-aie returns its whole list for the one buffer and the sequence carries no meaning. On a demux order *does* matter (destination i routes with `ids[i]`), but a demux's list is the domain's by construction, so there is nothing for it to disagree with.

Neither case had coverage before — added both: a hop pinning the same set in a different order (accepted, silent) and a hop pinning the same count with different values (error).

**Missing includes** — both correct. `Util.h` used `llvm::SmallDenseSet` without `llvm/ADT/DenseSet.h`, and `Util.cpp` used `ViewLikeOpInterface` without `mlir/Interfaces/ViewLikeInterface.h`; both were resolving transitively. Added explicitly, and verified `Util.h` now compiles in a TU that includes nothing else.

Re-gated after the change: full-IR output still byte-identical to an `origin/main` baseline binary on both decode builders.

PR #1806, opened 2026-08-15
[AIR] Fold an over-limit shim BD stride into base offsets

erwei-xilinx:
## Summary

A shim BD encodes each dimension's stride in a 20-bit field, so a stride above
`AIE2_STRIDE_UPPER_BOUND` (1048576) cannot be expressed and `aie.dma_bd` rejects it.
`AIRRtToNpuPass` already legalizes an illegal **wrap** (`violatesAIE2WrapLimit` /
`tileIllegalWrapDim`) but had nothing for an illegal **stride**, and tiling cannot help:
splitting a dim leaves its stride unchanged.

Such a dim can become base offsets instead. A dim (wrap `W`, stride `S`) covers the
addresses `{(off + k) * S : k in [0, W)}`, so the transfer is equivalent to `W` copies
that each carry one `k` in their own base.

The rewrite is **in place, not a rank reduction**, because `DmaToNpuPattern` indexes the
wrap-and-stride lists as a fixed rank-4 layout. The dim is set to wrap 1 with offset
`(off + k)`, which:

- keeps the rank;
- keeps the address, since that pattern derives the BD base from `sum(offsets[i] * strides[i])`;
- drops the dim from the BD, because a middle dim of size 1 fails the
  `size > 1 || i == 3 || (use4thDimInBd && i > 0)` test that selects `BDDimLayout` entries.

The oversized stride is therefore never encoded. That last step is what makes this legal,
so it is checked rather than assumed: only a middle dim of a BD with an unused 4th dim
qualifies.

## This is an opportunity, not a gate

Every "cannot fold" path leaves the op untouched. An airrt-level stride above the field
does not by itself mean the emitted BD is illegal — later steps still reshape the access
pattern — so ops this rewrite cannot help are passed through and downstream decides.

An earlier version of this patch raised an error there and **regressed the Llama-3.2-1B
prefill**, whose `o_ffn` kernel carries a dim-1 stride of 2097152 and compiled fine
before. A wrap above `AIE2_STRIDE_UNROLL_LIMIT` (16) is also left alone: folding it would
emit one descriptor per entry, which is worse than the problem.

## What it unblocks

This is what capped context length in the fused decode. The region-major KV cache strides
its per-group dim by `ATTN_MAXL*REGION_W`, which crosses the field above ctx 8k
(llama-3.2-1B) and 4k (llama-3.2-3B, gemma3-4b), while the dim itself is one entry per KV
group — always wrap 2, so the fold costs 2 descriptors per layer per K/V, **+4%** on a
772-BD sequence.

Decode latency on NPU2, llama-3.2-1B, built at `ATTN_MAXL = ctx`:

| ctx | ms | tok/s | Δms per 1k |
|---|---|---|---|
| 1k | 15.173 | 65.91 | — |
| 2k | 17.023 | 58.74 | 1.850 |
| 4k | 20.725 | 48.25 | 1.851 |
| 8k | 28.175 | 35.49 | 1.862 ← previous ceiling |
| 16k | 43.040 | 23.23 | 1.858 |
| 32k | 72.728 | 13.75 | 1.855 |
| 64k | 132.076 | 7.57 | 1.855 |
| 128k | 250.958 | 3.98 | 1.858 |

The model's full 128k context now builds and runs, and the slope holds at 1.85–1.86 ms/1k
across a 128× range — the extra descriptors cost nothing measurable. Also measured at 16k
on llama-3.2-3B (110.214 ms, 56 BDs folded) and gemma3-4b (99.087 ms, 68 folded).

Above 8k the remaining limit is not the compiler: 3B and gemma **build** at 128k and fail
in XRT host-memory allocation (14.0 / 17.0 GiB KV BOs, `mmap ... err=-11`). qwen2.5-3B was
never affected — its KV-append dim has wrap 1 (`NGRP=1`), so the stride was already dead.

## Validation

- `check-air-mlir`: 494 pass, 0 unexpected
- two new lit tests: the fold, the leave-alone-when-wrap-is-large case, and an
  at-the-bound case proving the pass is inert below the limit
- **no effect below the limit**: shipped llama-3.2-1B decode templates rebuild
  byte-identically (insts md5 `c17c55f2`), 64-token greedy id sequence unchanged
  (`350f4a6049dd`), `make verify` PASS 2/2
- 16k built natively produces insts byte-identical to a hand-edited-IR prototype, and runs
  at 43.009 ms / 23.25 tok/s
- the previously regressed llama-3.2-1B prefill runs again
- `clang-format` clean

Latency figures come from `bench_decode`, which uses synthetic weights — it is not a
numerical gate. Contexts above 8k are therefore latency-validated only; verifying numerics
there additionally needs a prefill engine built beyond `seq_len=2048`. Below the limit the
pass is provably inert, so the shipped configurations are fully covered by the gates above.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -1865,20 +1865,46 @@ LogicalResult unrollIllegalStrideDim(airrt::DmaMemcpyNdOp memcpy_op,
   // Folding costs one descriptor per entry of the folded dim.
   if (*constWrap > AIE2_STRIDE_UNROLL_LIMIT)
     return success();
+  // generateAwaitsFromWaitAllOps matches waits to configure tasks FIFO per
+  // channel -- the Nth wait for a channel awaits the Nth config for it. This
+  // rewrite turns one config into `wrap` of them without adding waits, so on a
+  // channel that already carries a wait the pairing shifts and the wait lands
+  // on the first piece while the rest go unawaited. Leave those alone: a
+  // silently under-synchronised transfer is far worse than an unfolded one.
+  if (auto metadata = memcpy_op->getAttrOfType<FlatSymbolRefAttr>("metadata")) {
+    bool channelIsWaited = false;
+    if (auto func = memcpy_op->getParentOfType<func::FuncOp>())
+      func.walk([&](AIEX::NpuDmaWaitOp wait) {
+        if (wait.getSymbol() == metadata.getValue())
+          channelIsWaited = true;
+      });
+    if (channelIsWaited)
+      return success();
+  }
 
+  // The op's !airrt.event is optional and, unlike tileIllegalWrapDim (which
+  // rewrites in place), this replaces the op -- so the result has to be
+  // carried across. The pieces are issued in program order on one channel, so
+  // the last one completing implies all of them have; give it the event and
+  // let it stand in for the original. This mirrors how coalesceShimDmaOrder
+  // redirects the event tokens of the ops it merges away.
+  SmallVector<Type> eventTy(memcpy_op->getResultTypes());
+  airrt::DmaMemcpyNdOp lastOp;
   for (int64_t k = 0; k < *constWrap; k++) {
     SmallVector<OpFoldResult> newOffsets(offsets), newWraps(wraps),
         newStrides(strides);
     newOffsets[i] = builder.getI64IntegerAttr(*constOff + k);
     newWraps[i] = builder.getI64IntegerAttr(1);
-    auto newOp = airrt::DmaMemcpyNdOp::create(
-        builder, loc, SmallVector<Type>{}, memcpy_op.getId(), memcpy_op.getX(),
-        memcpy_op.getY(), memcpy_op.getMemref(), newOffsets, newWraps,
-        newStrides);
+    bool isLast = k + 1 == *constWrap;
+    lastOp = airrt::DmaMemcpyNdOp::create(
+        builder, loc, isLast ? eventTy : SmallVector<Type>{}, memcpy_op.getId(),
+        memcpy_op.getX(), memcpy_op.getY(), memcpy_op.getMemref(), newOffsets,
+        newWraps, newStrides);
     // Ordering/barrier markers must ride along on every piece, or the split
     // silently drops the append barrier and the shim order constraint.
-    newOp->setAttrs(memcpy_op->getDiscardableAttrDictionary());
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Thanks — this found a real bug, and it is worse than described. Fixed in f068eae9.

**What I found.** I built the adversarial case (a DMA whose `!airrt.event` is consumed by `airrt.wait_all`) and ran it. It does *not* crash on `erase()`: `generateNpuWaitFromAIRRtWaitAll` runs before this rewrite, so the airrt event already has no uses by then. The failure is quieter and worse.

`generateAwaitsFromWaitAllOps` pairs waits to configure tasks FIFO per channel — the Nth wait for a channel awaits the Nth config for it. This rewrite turns one config into `wrap` of them **without adding waits**, so the pairing shifts. Output before the fix:

```mlir
%0 = aiex.dma_configure_task_for @airMemcpyId4 { aie.dma_bd(... offset = 16383 ...) }
aiex.dma_start_task(%0)
%1 = aiex.dma_configure_task_for @airMemcpyId4 { aie.dma_bd(... offset = 4210687 ...) }
aiex.dma_start_task(%1)
aiex.dma_await_task(%0)          // <-- awaits the FIRST piece only; %1 unawaited
```

A consumer could proceed after only part of the transfer had landed.

**Fix.** Treated as one more reason not to fold, which is how every other unhandled shape here is handled: if the channel already carries an `NpuDmaWaitOp`, leave the op alone. An under-synchronised transfer is far worse than an unfolded one, and the over-limit stride then falls through to downstream exactly as before this pass existed. The same input now emits one correctly-awaited strided BD.

I also carried the optional `!airrt.event` across the rewrite (the last piece takes it — the pieces are issued in order on one channel, so its completion implies the rest). That is inert in the current pipeline, but this rewrite is the first in the pass to *replace* the op rather than mutate it in place — `tileIllegalWrapDim` rewrites in place, which is why it never needed this — so it should not silently drop a result if the ordering ever changes.

**On the two test comments:** both correct, and the requested test is what exposed the bug. Added as a third case in `unroll_illegal_stride.mlir`, asserting the transfer stays unfolded *and* keeps its `dma_await_task`. I did not add a `wait_all` to the two existing cases, deliberately — they now cover the unwaited path (which folds) while the new case covers the waited path (which must not), so the pair pins both sides of the new guard.

`Conversion/AIRRtToNpu` 21/21 pass. Full `mlir/test` has 46 failures in `Util`/`aircc`/`Target`, all `posix_spawn failed` from binaries I did not build in this scratch tree — none in `Conversion` or `Transform`.

PR #1807, opened 2026-08-15
[AIR] Derive packet header ownership from `dest`, and drop it from the designs

erwei-xilinx:
The four q4nx/q4 decode designs each declared three things about packet headers: `keep_pkt_header` on two forwarding hops, and `air.src_writes_pkt_header` on the demux. **None of them were choices.** All three are now derived, and the designs state nothing.

## The lever is `dest(%d)`

A put naming a `dest` is a runtime value saying which leaf this packet is for. There is nothing else it can mean — so the packet channel it reaches covers its broadcast dimension over **time** rather than over **space**. That is the whole space-versus-time question, and the design was already answering it.

Worth being precise about what that question is, because `broadcast_shape` alone cannot express it. On a **circuit-switched** channel the stretch is always spatial: the route is static, there is no header, and nothing selects anything. Only on a **packet-switched** channel can the same `broadcast_shape` mean two different things — every packet copied to all N leaves (1 id), or each packet routed to exactly one of them (N ids). That ambiguity was being resolved by summing the volumes the gets consume, i.e. arithmetic standing in for intent.

## Three changes

**Classification stops reading the header attribute.** A partition of the stream across destinations is now reported as *evidence* of a demux; confirmation is separate. Requiring the attribute here is exactly what made it load-bearing for classification, and so impossible for a design to stop declaring — remove it and the channel stopped being recognised as a demux at all.

**Confirmation comes from reachability.** A demux is confirmed when a `dest`-carrying put reaches it forward through the hop graph, and the hops are the channels on that path: forward-reachable from the put *and* backward-reachable from the demux. Intersecting the two keeps an unrelated channel that merely feeds the same buffer out of the domain — the old attribute gate excluded those only incidentally. Designs with no `dest` anywhere fall back to the attribute-gated walk and are unaffected.

**`keep_pkt_header` is injected on the hops.** A hop carries a routing decision made upstream to a switchbox further downstream; strip the header there and the demux has nothing to route on. Forced by topology, not a choice. The demux itself is left alone: at the final consumers, stripping to a pure payload *is* a choice and stays the design's.

Also: a partition with no header writer anywhere no longer silently declines to allocate. It is rejected, via the check added in #1805. Previously air-to-aie would auto-assign one id that half the destinations never match.

## Verification

**IR gate.** Regenerated both decode builders with the attributes deleted and ran the full two-run sequence. Against a baseline binary built from merged main (`53fcac0d7`) on the *attributed* designs, the entire IR is identical except the removed `air.src_writes_pkt_header` line itself — 3124 and 1671 lines, **one line of diff each**.

That removal is provably inert. After the pass `@outY` carries three `packet_ids`, so `channelSourceWritesHeader` is true via its `size() > 1` branch — which is already what the bundle-split path relies on, since `copyChannelSteeringAttrs` does not copy that attribute. Its other consumers key on `keep_pkt_header`, absent on `@outY` both before and after.

**Device gates**, all with `make clean` first and `SKIPS=0`:

| model | builder | result |
|---|---|---|
| llama-3.2-1B | `fused_decode.py` | PASS (`topk_passed: 2, topk_failed: 0`) |
| llama-3.2-3B | `fused_decode.py` | PASS |
| gemma-3-4B | `fused_decode.py` | PASS (`*** PARIS ***`) |
| qwen2.5-3B | `fused_decode_qwen.py` | PASS (coherent, 12.08 tok/s) |

llama-1b was re-gated after rebasing onto current main and moving to the new mlir-aie pin from #1803 (`c84915b0`) — still PASS.

`check-air-mlir`: all packet-id tests pass. The only failures in the tree are 4 pre-existing `Util/Channel` tests that fail compiling a C++ host harness on a missing `air_tensor.h`, unrelated to this change.

No perf A/B: the generated IR is unchanged apart from one removed attribute line, so a measurement would report box noise rather than signal.

New coverage in `derive_from_dest.mlir`: the full derivation with a design that declares nothing, an off-path channel that must *not* be pulled into the domain, and a legacy `dest`-free design that still uses the attribute-gated walk.

### DIFF
--- a/mlir/lib/Transform/AIRMiscPasses.cpp
+++ b/mlir/lib/Transform/AIRMiscPasses.cpp
@@ -3641,9 +3641,17 @@ void AIRAnnotatePacketIDsPass::runOnOperation() {
         attrs.push_back(builder.getI32IntegerAttr(id));
       ArrayAttr idsAttr = builder.getArrayAttr(attrs);
       for (air::ChannelOp member : dom.hops) {
-        if (member.getPacketIDs())
-          continue;
-        member->setAttr(air::attrs::PacketIDs, idsAttr);
+        if (!member.getPacketIDs())
+          member->setAttr(air::attrs::PacketIDs, idsAttr);
+        // A hop MUST preserve the header. It is carrying a routing decision
+        // made upstream to a switchbox further downstream; strip it here and
+        // the demux has nothing left to route on. That is forced by the
+        // topology, not a design choice, so derive it rather than requiring
+        // every design to restate it. (At the demux itself keep is a real
+        // choice -- its consumers generally want pure payload -- so it is left
+        // alone.)
+        if (!member->hasAttr(air::attrs::KeepPktHeader))
+          member->setAttr(air::attrs::KeepPktHeader, builder.getUnitAttr());
       }
       if (!dom.idsDeclared)
         dom.demux->setAttr(air::attrs::PacketIDs, idsAttr);
--- a/mlir/lib/Util/PacketRoutingDomain.cpp
+++ b/mlir/lib/Util/PacketRoutingDomain.cpp
@@ -483,9 +483,26 @@ void PacketRoutingDomainAnalysis::buildDomains() {
         if (upstream.contains(c))
           onPath.insert(c);
     }
-    for (ChannelOp c : packetChans)
-      if (onPath.contains(c.getOperation()))
-        dom.hops.push_back(c);
+    // Collect by walking UPSTREAM from the demux, filtered to the path, so the
+    // order means something. Iterating packetChans here would collect in
+    // declaration order, which the std::reverse below then INVERTS -- in a
+    // simple chain that yields exactly the reverse of the travel order the
+    // field is documented to hold.
+    {
+      SmallVector<ChannelOp> wl{d};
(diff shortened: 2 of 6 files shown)


erwei-xilinx:
Addressed in 058aab8ab. Good catch — this was a real bug.

With the on-path construction, `dom.hops` was collected by iterating `packetChans` in **declaration** order and then unconditionally reversed, so in a simple chain it held exactly the reverse of the travel order the field is documented to have. For the decode designs that was `@toHub -> @outA`.

Fixed as you suggested: walk upstream from the demux via `fedBy`, filtered to the on-path set, as the legacy path already did. The existing `std::reverse` then yields a true furthest-upstream-first order.

One thing worth adding on top of the fix — the reason nothing caught it. Every consumer of `dom.hops` is order-independent (writing ids, injecting keep, the pin cross-check), and `printReport` isn't wired to any flag, so the documented ordering had **no observable surface**. A fix I couldn't test would just have regressed later. So the demux's report remark now spells the chain out:

```
@fanout: demux over 3 destination(s); infers 3 routing id(s); fed by @hopA -> @hopB
```

and the reduced case in `derive_from_dest.mlir` CHECKs it. The models report `@outA -> @toMain` and `@outA -> @toHub`.

No codegen change: the IR still differs from the merged-main reference by only the removed `air.src_writes_pkt_header` line, on both builders.

PR #1812, opened 2026-08-16
[docs] Document `dest`, and the space-vs-time question broadcast_shape cannot answer

erwei-xilinx:
`docs/AIRComputeModel.md` said only this about fan-out:

> The `broadcast_shape` attribute enables one-to-many communication following NumPy broadcasting rules.

That describes the **spatial** case and nothing else. On a packet channel the same attribute equally describes a demux, where each transfer reaches exactly *one* destination. And `dest`, the operand that distinguishes the two, appeared nowhere in the document.

The document splits abstract operation semantics (§2) from backend mapping (§3), so this does too.

**§2.5** gets the `dest(%d)` syntax and its meaning, in the dialect's own terms: `broadcast_shape` declares how many destinations a channel reaches, not whether a given transfer reaches all of them; `dest` names one coordinate along the broadcast dimension at run time, making the fan-out temporal rather than spatial; and it sits on the *originating* put rather than the fan-out channel, because the fan-out is typically performed by a data-mover stage that only relays a buffer it did not produce and so cannot know where the data should go.

**§3** gets the AIE half: circuit-switched types are statically routed and support spatial fan-out only; the packet-switched overlay supports both. It says the compiler realises the mode from `dest` alone and rejects a `dest` it cannot realise — and stops there. The routing ids, header placement and per-stage preservation flags are internal; a user-facing manual that enumerated them would invite designs to write what the compiler now derives.

46 lines added, docs only, no code change.

Follows #1799 / #1805 / #1807.

### DIFF
--- a/docs/AIRComputeModel.md
+++ b/docs/AIRComputeModel.md
@@ -695,6 +695,9 @@ air.channel @name [dim₀, dim₁, …] {channel_type = "npu_dma_stream", depth 
 air.channel.put @name[indices] (src[offsets][sizes][strides]) : (type_src)
 air.channel.get @name[indices] (dst[offsets][sizes][strides]) : (type_dst)
 
+// Put naming a run-time destination — fan-out over time
+air.channel.put @name[indices] (src[offsets][sizes][strides]) dest(%d) : (type_src)
+
 // Asynchronous put/get — return !air.token; transfer completes when token signals
 [%token =] air.channel.put @name[indices] [dependency = [%t₀, …]]
            (src[offsets][sizes][strides]) : (type_src)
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Addressed in 359c9f404.

**Spelling of "runtime"** — correct, and it applied in three places rather than two. The document already uses "runtime" 15 times and nothing else; my additions introduced "run-time" twice and "run time" once. All normalised, so the file is now consistent at 18 occurrences.

**The inline IR example** — that comment, and most of the review body, describes an earlier revision of this branch. The example it cites (line 756, `@outY`'s three consumers), along with the `packet_ids` / `keep_pkt_header` / header-store tables, were removed in 8ef2424e5 before this review ran. They were too much for a user-facing manual: those attributes are internal and now compiler-derived, and enumerating them here would invite designs to write what the compiler generates. §3 now says the compiler realises the mode from `dest` and rejects a `dest` it cannot realise, and stops there.

Current diff is 46 added lines, docs only.

PR #1813, opened 2026-08-17
[AIR] Revert the shim-BD stride fold's wait pairing (#1810)

erwei-xilinx:
Reverts #1810. The nightly LLM benchmark went red at d0e522bb with four new
failures, all traceable to that commit. The fold itself (#1806) stays.

## Cause

#1806 refused to fold on a channel that already carries a wait, because
splitting one `airrt.dma_memcpy_nd` into `wrap` of them shifts the FIFO wait
pairing. #1810 replaced that refusal with hand-emitted `NpuDmaWaitOp`s.

But `enforceAIE2StrideLimit` runs at `AIRRtToNpuPass.cpp:2274` and
`generateAwaitsFromWaitAllOps` at `:2344`. The awaits derived from
`airrt.wait_all` do not exist yet when the fold counts them, and the fold does
not split the `wait_all` that produces them. The 1:1 invariant #1810 means to
preserve is computed against an incomplete picture, and the guard it removed was
the only thing keeping these designs off the path.

## Symptoms

Two distinct ones, both measured on NPU2, both gone at 13e02eb5:

| test | d0e522bb | 13e02eb5 |
|---|---|---|
| `llms/qwen25_3b` compile | BD overflow | pass |
| `llms/qwen25_1_5b` compile | BD overflow | pass |
| `llms/qwen25_3b_q4` compile | BD overflow | (shares the qwen prefill builder) |
| `llms/llama32_1b_int4` verify-prefill bfp16 | overlap 0/10, r = -0.8347, `'endi'` | overlap 9/10, r = 0.9938, `' Paris'` |

**Qwen** — the prefill folds a `wrap=16` dim (stride 1409024 = 128*11008) on four
shim channels. Sixteen configs stay live until one wait point, so tile (0,0)
needs 64 of its 16 BDs:

```
'aiex.dma_configure_task' op Too many simultaneously active buffer descriptors
on tile (0,0), which supports up to 16.
```

`AIE2_STRIDE_UNROLL_LIMIT` is 16 — a descriptor-count heuristic equal to the
entire per-tile budget, so it never fires.

**llama32_1b_int4** — folds `wrap=2`, the same shape #1810 was written for and
verified on. Compiles and runs, then answers `'endi'` instead of `' Paris'`.

That wrap=2 case is why this is a revert and not a tightened limit: bounding the
fold by BD cost would fix qwen and leave llama32_1b_int4 silently wrong.

## What this gives up

#1810's own target, the fused decode above ctx 4k/8k, which needs the fold on
waited channels. No CI test covers it — the nightly was green at 9b037828,
before either commit — so nothing regresses here. Re-landing it needs the wait
split to happen where the await set is final, with these two designs as gates.

## Verification

- `check-air-mlir`: 505 passed, 7 unsupported, 7 XFAIL, 0 failures.
- NPU2: `qwen25_3b` and `qwen25_1_5b` compile clean; `llama32_1b_int4`
  verify-prefill bfp16 PASS (overlap 9/10, r = 0.9938).
- The reverted tree is byte-identical to 13e02eb5 for both changed files.

Note: `llms/gemma3_4b_q4nx/run_npu2_verify.lit` is currently UNSUPPORTED on the
nightly runner (its bf16 reference `unsloth/gemma-3-4b-it` is not yet in the
runner's HF cache). Unrelated to this revert; needs a `downloadLLMWeights.yml`
dispatch.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -2055,22 +2055,20 @@ LogicalResult unrollIllegalStrideDim(airrt::DmaMemcpyNdOp memcpy_op,
     return success();
   // generateAwaitsFromWaitAllOps matches waits to configure tasks FIFO per
   // channel -- the Nth wait for a channel awaits the Nth config for it. This
-  // rewrite turns one config into `wrap` of them, so the channel needs `wrap`
-  // waits or the pairing shifts: the original wait would land on the first
-  // piece, every later transfer on the channel would be awaited one slot
-  // early, and the tail pieces would go unawaited -- on an S2MM channel that
-  // is both a missed completion token and a BD that is never freed. Emit the
-  // extra waits alongside the extra configs so the 1:1 invariant survives.
-  //
-  // Only on a channel that is already waited: adding waits to a fire-and-
-  // forget channel would impose synchronisation the design never asked for.
-  StringRef waitedSym;
-  if (auto metadata = memcpy_op->getAttrOfType<FlatSymbolRefAttr>("metadata"))
+  // rewrite turns one config into `wrap` of them without adding waits, so on a
+  // channel that already carries a wait the pairing shifts and the wait lands
+  // on the first piece while the rest go unawaited. Leave those alone: a
+  // silently under-synchronised transfer is far worse than an unfolded one.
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Reopening from a branch in this repo so nightlyPerfBenchmark.yml can be dispatched against it.

PR #1819, opened 2026-08-18
[AIR] Give host readbacks their own shim channel and tile, and drop 3 more shim_col pins

erwei-xilinx:
A shim readback ends off-chip, so `allocateL3` hands the allocator the **producer
core's** tile and column as the flow's "other side":

```cpp
auto mm2sTile = f.MM2S_alloc[0].getDmaTile();
shim_dma_alloc.allocNewDmaChannel(memcpyOpIf, mm2sTile,
                                  mm2sTile.tryGetCol().value_or(-1), ...);
```

Herds are placed before `air-to-aie`, so that column is concrete and every readback
out of a herd keys on it. Two things then collapse: the bucket already holds the
**feeds into those same cores**, so the readback joins them; and the packet path
reuses "the bucket's existing packet channel", so independent readbacks land on one
channel. llama-3.2-1B put `appendK`, `appendV` and `layerOut` on **(2,0) S2MM 0** —
one tile, one channel — and the pathfinder then could not route an unrelated flow,
`air_xnorm`, out of column 2:

```
'aie.packet_flow' op packet flow source (2, 2) DMA1 could not be routed to
destination (1, 1) DMA2
```

`air.shim_col` was pinning that apart by hand. #1803 kept 3 pins per design for
exactly this reason, having measured they were the only ones that changed anything.

## Fix

Both scoped to packet readbacks on unpinned channels:

- don't reuse a packet channel that already hosts a readback from a **different**
  channel decl (sub-channels of one bundled decl still multiplex — `@outD [2,2]`
  needs that);
- steer the readback to the nearest free shim LTO holding no other readback, ranked
  by (distance from the producer column, channels used), instead of to its
  producer's bucket.

Reusing an existing LTO rather than opening one per readback is deliberate: the
pipeline runs `aie-place-tiles{merge-logical-tiles=false}`, so every LTO costs a
whole shim tile and NPU2 has 8. One-LTO-per-readback wants 9 here and fails to
place. Joining another column's bucket is exactly what a pin did (`appendK` pinned
to col 3 shared the `inKV` LTO).

**No mlir-aie change is needed.** An earlier version made the placer's column choice
congestion-aware; reverting it produced identical placement, because the reused LTOs
take their column from their own buckets and the readback never drives a placement
decision.

## The compiler change is a no-op on everything that exists today

Pinned flows bypass all of it. With the pins still in place the llama-3.2-1B decode
templates rebuild **byte-identical** (insts md5 `6d6ae12e` / `2eb3876a`) and all 19
shim allocations are unchanged.

## Measured on NPU2 — pins on vs pins off, 5 reps each

| model | verify | tok/s | TTFT |
|---|---|---|---|
| llama32_1b_q4nx | PASS | 50.16 → 50.00 (−0.3%) | 952 → 938 ms |
| llama32_3b_q4nx | PASS | 21.85 → 21.38 (−2.2%) | 1880 → 1940 ms |
| gemma3_4b_q4nx | PASS | 17.10 → 16.70 (−2.3%) | 4000 → 4079 ms |
| qwen25_3b_q4 | PASS | 11.79 (pins untouched) | 4236 ms |

`check-air-mlir` 505 passed, 0 failed.

The ~2% on 3b and gemma is the cost of the compiler's column choice versus the
hand-tuned 3/4/5, and is accepted here: the pins were unmaintainable and blocked any
new unpinned design. Recovering it likely means preferring a column already carrying
the same DDR buffer's traffic — which is why 3/4/5 were good — and is left for
follow-up.

`fused_decode_qwen.py` keeps its pins; it is a separate file needing its own
validation, and its `ropeLUT` pin has a documented circuit→packet interaction.

### DIFF
--- a/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
+++ b/mlir/include/air/Conversion/AIRToAIESchedulingUtils.h
@@ -139,6 +139,20 @@ struct allocation_info_t {
   Operation *otherSideLTO = nullptr;
   std::vector<int32_t> dma_id;
   std::vector<Operation *> memcpyOps;
+  // True when the shim is this flow's DESTINATION (S2MM), i.e. a readback whose
+  // data leaves for host DDR. `otherSideLTO`/`col` then describe the PRODUCER
+  // core, because the real far side is off-chip and has no column. Bucketing on
+  // a producer column is wrong -- it says nothing about where a transfer should
+  // leave the array, and it forces every readback out of a herd onto that
+  // herd's column. That is how three llama-1b readbacks plus the feeds to the
+  // same cores landed on one shim tile and one channel, which does not route.
+  // A readback is therefore steered to the nearest free shim LTO holding no
+  // other readback, instead of to its producer's bucket. See the readback
+  // handling in ShimDMAAllocator::allocNewDmaChannel.
+  //
+  // Declared last on purpose: several sites aggregate-initialize this struct
+  // positionally, so a field inserted earlier silently shifts them.
+  bool isHostReadback = false;
   bool valid();
   AIE::TileLike getDmaTile();
   bool foundAlloc(AIE::TileLike tile);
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Fixed in 80d7fd17.

The first point was a real code bug, not just a comment one: `hostsOtherPacketReadback` tested only direction, so a *circuit* readback would also push a new flow off that tile. Only packet flows time-multiplex a shim channel, so only they can collapse; the filter now tests flow type. air-to-aie output is byte-identical before and after on the design it was written for, so the PR's measurements are unaffected.

PR #1830, opened 2026-08-19
[programming_examples] Port relu to air.api, adding maximum/minimum/relu ops

erwei-xilinx:
> ~~Stacked on #1828.~~ **#1828 has merged**; this branch is rebased onto `main` and is now standalone and conflict-free. The redundant axpy commit was dropped in the rebase — verified that this branch's `programming_examples/axpy` is byte-identical to merged `main`, so nothing was lost.

Ports `programming_examples/relu` to the `air.api` DSL and deletes the raw-bindings implementation, continuing the series from #1823 (eltwise_add) and #1828 (axpy).

relu was blocked on exactly one missing operator, so this adds it. The whole compute body becomes:

```python
out_buf[:] = ops.relu(x_buf[:])
```

## New DSL surface: `ops.maximum`, `ops.minimum`, `ops.relu`

`air.api` previously implemented only `+ - * /`. These fill in names the API proposal had stubbed as `NotImplementedError`. They build lazy expression nodes exactly like the arithmetic operators, so they compose in one tree and lower as a single vectorised loop — `python/test/api/relu.py` pins `min(max(x,0),6)*2` doing that.

The float forms are `arith.maximumf`/`minimumf`, deliberately **not** `maxnumf`/`minnumf`: `maximumf` is what the hand-written kernel emitted, and the LLVM 24 bump regressed scalar f32 `maxnumf` specifically.

## Coverage audit: old vs new

Since this deletes the predecessor, here is the side-by-side. Every axis is a gain or unchanged.

| Axis | Before | After | |
|---|---|---|---|
| Shape / tile | 1-D N=65536, 1024 | same flags, same defaults | = |
| dtype | bf16 only (hardcoded) | bf16 + f32 | gain |
| Vector width | 16 | 16 bf16 / 0 f32 (see below) | = |
| Herd | fixed 2 cores | 4 on npu1, 8 on npu2; `--herd-shape` | gain |
| Device target | implicit | `--target auto\|npu1\|npu2`, auto-detected | gain |
| Output format | xclbin / elf | same | = |
| Perf harness | none | `--perf-iters` | gain |
| Correctness scope | 100 sampled indices | all N elements | gain |
| Tolerance | rtol=1e-2, atol=1e-8 (runner default) | **identical** | = |
| Harness / PASS string | `XRTRunner` / `PASS!` | same, plus `report_precision` | gain |
| **Hardware lit invocations** | **1** | **5** (4 peano + 1 elf) | gain |
| lit `REQUIRES` | `ryzen_ai, peano` | same, plus `ryzen_ai_npu2` for elf | gain |
| Makefile | 4 targets, no knobs | 5 targets, 6 knobs | gain |

`--output-format elf` was reachable from the CLI before but exercised by no test; it now has a gate. The full-array check runs at the predecessor's *exact* tolerance, not a relaxed one.

Behaviour changes, none silent: `-p/--print-module-only` → `--print-ir` and `--compile-mode compile-only` → `--compile-only` (loud argparse failures; nothing outside the example's own Makefile referenced the old spellings); the predecessor's `assert tile_n % vector_size == 0` becomes the DSL's documented scalar fallback.

## Verification

Hardware, npu1 (Phoenix) — all `PASS!`, and all **bit-exact**:

| Config | mean_rel_L1 | rel_err max |
|---|---|---|
| bf16 vectorised | 0.000e+00 | 0.000e+00 |
| bf16 scalar | 0.000e+00 | 0.000e+00 |
| f32 scalar | 0.000e+00 | 0.000e+00 |
| bf16 N=131072, tile 2048 | 0.000e+00 | 0.000e+00 |

Zero error is the expected result, not luck: `max(x, 0)` returns either the input or zero, so it introduces no rounding at all.

Compile-only sweep green across dtype × target × vector width on both generations. npu2 is left to CI — no npu2 hardware on the development box.

**Performance A/B**, same session, alternating, 10 paired reps, Default power mode:

| Arm | Config | Median | Spread |
|---|---|---|---|
| A | predecessor, 2 cores | 219.9 µs | 198.7 – 293.0 |
| B | air.api, 4 cores | 202.1 µs | 194.3 – 210.8 |

B ahead in **8/10** paired reps, **+8.1%** on the median, and materially steadier.

One caveat worth recording: a matched-width arm (air.api pinned to 2 cores) came out *slower* than the predecessor at the same core count in 3/4 reps, so relu's gain appears to come from using the whole device rather than from better per-core code. That was only 4 noisy reps — suggestive, not established. At 256 KB of traffic and ~200 µs this kernel is dispatch-bound, which is also why doubling the cores buys only 8%.

Method note: host-observed wall time (`time.perf_counter()` around `run.start()/wait2()`), not device-side bandwidth — relu has no `test.cpp` harness. Adequate for a relative A/B, which is why no absolute bandwidth figure is quoted.

## f32 runs scalar

Pinned by the example, not by the emitter. mlir-aie's `convert-vector-to-aievec` rejects a vector `maximumf` on f32 outright:

```
warning: aievec.max conversion fails due to unsupported element data type
error: failed to legalize operation 'aievec.max'
```

on both generations, while scalar f32 max is fine. The DSL still vectorises f32 on request; it is the example that asks for scalar, since whether a given operator has an aievec form is a backend property the DSL cannot see.

### DIFF
--- a/programming_examples/generate_readme.py
+++ b/programming_examples/generate_readme.py
@@ -137,7 +137,7 @@ EXAMPLES = [
         "category": "Activation/Math",
         "name": "RELU",
         "path": "relu",
-        "datatypes": "bf16",
+        "datatypes": "bf16, f32",
     },
     {
         "category": "Activation/Math",
--- a/programming_examples/relu/Makefile
+++ b/programming_examples/relu/Makefile
@@ -13,14 +13,38 @@ endif
 OUTPUT_FORMAT ?= xclbin
 OUTPUT_FORMAT_FLAG = --output-format $(OUTPUT_FORMAT)
 
+# Extra Python flags (used by LIT tests to override defaults)
+EXTRA_PY_FLAGS ?=
+
 all: run
 
+N ?= 65536
+TILE_N ?= 1024
+DTYPE ?= bf16
+VECTOR_SIZE ?=
+PERF_ITERS ?= 0
+
+VEC_FLAG = $(if $(VECTOR_SIZE),--vector-size $(VECTOR_SIZE),)
+# The device is auto-detected: one invocation has to run on whichever NPU is
+# installed. Set TARGET=npu1|npu2 to compile for a generation that is not
+# plugged in.
+TARGET ?= auto
+PY_FLAGS = --n $(N) --tile-n $(TILE_N) --dtype $(DTYPE) $(VEC_FLAG) \
+	--target $(TARGET) $(OUTPUT_FORMAT_FLAG) $(EXTRA_PY_FLAGS)
+
 print:
-	${powershell} python3 ${srcdir}/relu.py $(OUTPUT_FORMAT_FLAG) -p
+	${powershell} python3 ${srcdir}/relu.py $(PY_FLAGS) --print-ir
+
+compile:
+	mkdir -p $(BUILD_DIR)
+	PEANO_INSTALL_DIR=$(PEANO_INSTALL_DIR) cd $(BUILD_DIR) && \
+		${powershell} python3 ${srcdir}/relu.py $(PY_FLAGS) --compile-only
 
 run:
 	mkdir -p $(BUILD_DIR)
-	PEANO_INSTALL_DIR=$(PEANO_INSTALL_DIR) cd $(BUILD_DIR) && ${powershell} python3 ${srcdir}/relu.py $(OUTPUT_FORMAT_FLAG)
+	PEANO_INSTALL_DIR=$(PEANO_INSTALL_DIR) cd $(BUILD_DIR) && \
+		${powershell} python3 ${srcdir}/relu.py $(PY_FLAGS) \
+		--perf-iters $(PERF_ITERS)
 
 clean:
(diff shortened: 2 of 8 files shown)


erwei-xilinx:
Thanks — good catch, and it reproduced exactly as described:

```
MLIRError: Invalid attribute: error: unknown: expected floating point type
```

Fixed in 808c8e33. `ops.relu` now takes the Python type of its zero from the operand's dtype, so an integer buffer gets `arith.constant 0 : i32` feeding `arith.maxsi`, and float behaviour is unchanged.

I also fixed the general case behind it, since the same failure was reachable without `relu` at all — `int_buf[:] + 1.0` produced the identical message, which names neither the offending scalar nor the expression it came from. The emitter's scalar path now accepts a whole-number float on an integer type (a harmless way to write an integer constant) and rejects a non-integral one with a message that says which value is wrong and why, rather than truncating silently.

`python/test/api/relu.py` gains an integer case pinning `arith.maxsi` against an i32 zero, so this cannot regress unnoticed.

PR #1831, opened 2026-08-19
[programming_examples] Port leaky_relu to air.api without adding select

erwei-xilinx:
> **Stacked on #1830** (#1828 has since merged, so this is now a two-deep stack, rebased onto `main` and conflict-free). This uses `ops.maximum`/`ops.minimum`, added in #1830. Merge that first; this PR's diff then collapses to the leaky_relu-only changes — 6 files.

Ports `programming_examples/leaky_relu` to the `air.api` DSL and deletes the raw-bindings implementation, continuing the series from #1823 (eltwise_add), #1828 (axpy) and #1830 (relu).

## No `select` was added to the DSL

The predecessor expressed the kernel with a comparison and a select:

```python
cmp    = arith.CmpFOp(OGE, x, 0)
result = arith.SelectOp(cmp, x, alpha * x)
```

`air.api` has neither, and this PR does not add them. It uses the identity

```
leaky_relu(x, alpha) = max(x, 0) + alpha * min(x, 0)
```

which needs only operators the DSL already has after #1830. For `x >= 0` the max contributes `x` and the min contributes zero; for `x < 0` the max contributes zero and the min contributes `x`.

This is **bit-exact, not approximate**, and was checked rather than assumed:

- the two forms are bit-identical over 200k unit-normal samples plus the awkward inputs (`+0`, `-0`, `±1e-30`, `±1`), in both bf16 and f32, at `alpha ∈ {0.01, 0.1, 0.2, 0.5}`;
- simulating **both** forms under bf16 intermediate rounding gives bit-identical device semantics, so the rewrite adds no rounding step.

`python/test/api/leaky_relu.py` pins the emitted shape with `CHECK-NOT: arith.cmpf` / `CHECK-NOT: arith.select`, so the rewrite cannot be silently undone.

### Reading the precision line

bf16 reports `rel_err max = 7.8e-03` against an fp32 reference, which is one bf16 ULP. That comes entirely from `alpha` rounding to bf16 (`0.01` → `0.010009765625`) — it is inherent to bf16 leaky ReLU and present in the predecessor **identically**. I verified that rather than inferring it. f32 is bit-exact.

## Coverage audit: old vs new

| Axis | Before | After | |
|---|---|---|---|
| Shape / tile / alpha | 1-D N=65536, 1024, 0.01 | same flags, same defaults | = |
| dtype | bf16 only (hardcoded) | bf16 + f32 | gain |
| Vector width | 16 | 16 bf16 / 0 f32 (see below) | = |
| Herd | fixed 2 cores | 4 on npu1, 8 on npu2; `--herd-shape` | gain |
| Device target | implicit | `--target auto\|npu1\|npu2`, auto-detected | gain |
| Output format | xclbin / elf | same | = |
| Perf harness | none | `--perf-iters` | gain |
| Correctness scope | 100 sampled indices | all N elements | gain |
| Tolerance | rtol=1e-2, atol=1e-8 (runner default) | **identical** | = |
| **Hardware lit invocations** | **1** | **5** (4 peano + 1 elf) | gain |
| lit `REQUIRES` | `ryzen_ai, peano` | same, plus `ryzen_ai_npu2` for elf | gain |
| Makefile | 4 targets, no knobs | 5 targets, 7 knobs | gain |

The reference is computed in the predecessor's own `np.where` form, so the check is against the semantics being replaced rather than against the rewrite.

## Verification

Hardware, npu1 (Phoenix) — all `PASS!`:

| Config | mean_rel_L1 | rel_err max |
|---|---|---|
| bf16 vectorised | 3.666e-05 | 7.812e-03 |
| bf16 scalar | 3.666e-05 | 7.812e-03 |
| f32 scalar | 0.000e+00 | 0.000e+00 |
| bf16 α=0.2, N=131072 | 6.002e-04 | 7.812e-03 |

Compile-only sweep green across dtype × target × vector width on both generations. npu2 left to CI.

**Performance A/B**, same session, alternating, 12 paired reps:

| Arm | Config | Median |
|---|---|---|
| A | predecessor, cmp+select, 2 cores | 232.1 µs |
| B | air.api, max+min, 4 cores | 218.6 µs |

B ahead in **9/12** paired reps, **+5.8%** median, +5.3% on min. The first six reps were genuinely ambiguous (3/6) with a 434 µs outlier in arm A, so I ran six more at 400 iterations where dispatch noise averages out better — those are **6/6**. Worth stating plainly: the extra op per element (max, min, mul, add vs cmp, select, mul) costs nothing measurable here, because at 256 KB of traffic and ~220 µs the kernel is dispatch-bound.

Method: host-observed wall time (`time.perf_counter()` around `run.start()/wait2()`), not device-side bandwidth — leaky_relu has no `test.cpp`. Adequate for a relative A/B, which is why no absolute bandwidth is quoted.

## f32 runs scalar

Pinned by the example, not the emitter: aievec has no f32 max (`aievec.max conversion fails due to unsupported element data type`), and separately a chained f32 multiply-then-add does not legalize at any vector width. bf16 vectorises the whole expression at 16 lanes.

### DIFF
--- a/programming_examples/generate_readme.py
+++ b/programming_examples/generate_readme.py
@@ -143,7 +143,7 @@ EXAMPLES = [
         "category": "Activation/Math",
         "name": "Leaky RELU",
         "path": "leaky_relu",
-        "datatypes": "bf16",
+        "datatypes": "bf16, f32",
     },
     {
         "category": "Activation/Math",
--- a/programming_examples/generate_readme.py
+++ b/programming_examples/generate_readme.py
@@ -137,7 +137,7 @@ EXAMPLES = [
         "category": "Activation/Math",
         "name": "RELU",
         "path": "relu",
-        "datatypes": "bf16",
+        "datatypes": "bf16, f32",
     },
     {
         "category": "Activation/Math",
--- a/programming_examples/leaky_relu/Makefile
+++ b/programming_examples/leaky_relu/Makefile
@@ -13,14 +13,39 @@ endif
 OUTPUT_FORMAT ?= xclbin
 OUTPUT_FORMAT_FLAG = --output-format $(OUTPUT_FORMAT)
 
+# Extra Python flags (used by LIT tests to override defaults)
+EXTRA_PY_FLAGS ?=
+
 all: run
 
+N ?= 65536
+TILE_N ?= 1024
+DTYPE ?= bf16
+ALPHA ?= 0.01
+VECTOR_SIZE ?=
+PERF_ITERS ?= 0
+
+VEC_FLAG = $(if $(VECTOR_SIZE),--vector-size $(VECTOR_SIZE),)
+# The device is auto-detected: one invocation has to run on whichever NPU is
+# installed. Set TARGET=npu1|npu2 to compile for a generation that is not
+# plugged in.
+TARGET ?= auto
+PY_FLAGS = --n $(N) --tile-n $(TILE_N) --alpha $(ALPHA) --dtype $(DTYPE) $(VEC_FLAG) \
+	--target $(TARGET) $(OUTPUT_FORMAT_FLAG) $(EXTRA_PY_FLAGS)
+
 print:
-	${powershell} python3 ${srcdir}/leaky_relu.py $(OUTPUT_FORMAT_FLAG) -p
+	${powershell} python3 ${srcdir}/leaky_relu.py $(PY_FLAGS) --print-ir
+
+compile:
+	mkdir -p $(BUILD_DIR)
+	PEANO_INSTALL_DIR=$(PEANO_INSTALL_DIR) cd $(BUILD_DIR) && \
+		${powershell} python3 ${srcdir}/leaky_relu.py $(PY_FLAGS) --compile-only
 
 run:
 	mkdir -p $(BUILD_DIR)
-	PEANO_INSTALL_DIR=$(PEANO_INSTALL_DIR) cd $(BUILD_DIR) && ${powershell} python3 ${srcdir}/leaky_relu.py $(OUTPUT_FORMAT_FLAG)
+	PEANO_INSTALL_DIR=$(PEANO_INSTALL_DIR) cd $(BUILD_DIR) && \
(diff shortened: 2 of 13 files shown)


erwei-xilinx:
Thanks — addressing the three comments.

**f32 gated at the bf16 tolerance (relu.py:205, leaky_relu.py:213) — valid, fixed.**

Correct, and it was the wrong number for a specific reason: `rtol=1e-2` is the *predecessor's* tolerance, and the predecessor was bf16-only. So it is right to carry it for bf16 and wrong to extend it to f32, which has no predecessor tolerance to inherit and was left an order of magnitude slacker than `XRTRunner`'s own default — on the one path with nothing to fall back on.

Both examples now split it:

```python
rtol, atol = (1e-2, 1e-8) if args.dtype == "bf16" else (1e-3, 1e-8)
```

That also brings them in line with axpy (#1828), which already did this. Each bar is now justified from its own measurement: bf16 leaky_relu genuinely needs 1e-2, because its `rel_err max` of 7.8e-03 is one bf16 ULP arising from `alpha` rounding to bf16 (`0.01` → `0.010009765625`); f32 is bit-exact in both examples. Re-verified on npu1 at the tightened gate — f32 `0.000e+00` against `rtol=1e-3`, bf16 unchanged.

**`--output-format elf` not enforced as npu2-only (leaky_relu.py:189, :214) — no change, and here is why.**

The backend already rejects it, before aircc runs:

```
AirBackendError: output_format='elf' is not supported for npu1 target. ELF output format
is only supported on npu2 and later devices. Use output_format='pdi' for a raw PDI
alongside the NPU instruction sequence.
```

I ran `--compile-only --target npu1 --output-format elf` to confirm rather than assume. That message is more useful than an argparse error would be, since it names the alternative. An example-level guard also could not fire any earlier in the `--target auto` case: the target is only known after `launch.build()` resolves it, which is essentially where the backend check already sits. Duplicating it per example would add churn around a check that already exists centrally, at the right layer. Happy to add it if you would still prefer the earlier failure.

PR #1836, opened 2026-08-20
[llms] Add Qwen3-8B Q4NX, and split the decode weight BO past 4 GiB

erwei-xilinx:
Adds `programming_examples/llms/qwen3_8b_q4nx` — Qwen3-8B (36L, K=4096, GQA 4,
untied Q4NX lm_head) as a Q4NX prefill + fused decode example reproducing
FastFlowLM's mechanism. The prefill is a thin driver over the `qwen3_4b`
config-parameterized builders rather than a self-contained clone.

Measured on NPU2 (Strix), production `LBUILD=2048` templates:

| | |
|---|---|
| `make verify` (2-prompt CI slice) | **2 / 2 PASS** vs `Qwen/Qwen3-8B` |
| `make verify-full` (8 prompts) | **8 / 8 PASS** |
| prefill TTFT | 7.72 s at CTX=2048 (265 tok/s) |
| decode | 10.21 tok/s (98 ms/token), 64 tokens |

FastFlowLM publishes 11.9 tok/s for this model at 1k context, so decode is at
0.86x of the implementation it reproduces.

## The engine change: `DECODE_WGROUP`

Qwen3-8B is the first model on this engine whose weights exceed a single BO's
addressable span. A shim BD's byte offset is a `uint32` end to end
(`aie.dma_bd $offset` → `aiex.npu.address_patch $arg_plus` → `uint32_t
patchedArgPlus` in `AIETargetNPU`), so one buffer reaches at most 4 GiB. 36
layers are 4.04 GiB plus a 0.37 GiB lm-head; past layer 33 the offsets wrapped
and **every logit came back NaN**. The unsplit build shows it directly —
`offset = -2144075776`, already negative.

`DECODE_WGROUP=G` splits the weights over `ceil(UNI_DEC/G)` layer buffers plus a
dedicated lm-head buffer, each addressed from its own base, selected per wave by
an `scf.index_switch` on the wave induction variable. The wave loop is unrolled
before BD assignment, so the switch folds to its single taken arm. Qwen3-8B runs
at G=9: four 1.01 GiB groups, still **one dispatch**.

This is the same layer-invariant addressing FastFlowLM gets by binding one BO per
layer, except FLM needs 36 dispatches and an XRT runlist to do it — their runtime
sequence is one reusable program with one weight arg, while ours is fully
unrolled, so each wave's BDs can name a different arg inline.

Two constraints on that switch, both load-bearing and both documented in the code:

1. The selector may only use ops in `AIRRtToNpuPass`'s post-unroll fold set.
   `iv / G` and `iv % G` are the obvious spelling and do **not** fold —
   `divui`/`remui` are absent from that set — and an unfolded `index_switch`
   cannot parent `aiex.dma_configure_task_for`. The group index is a `select`
   chain instead.
2. It must wrap the whole phase fan, not each column. The cross-channel phase
   barrier opens a new phase group at every block boundary and each switch arm is
   a block; a per-column switch shrinks the group from `2*NCX` channels to 2 and
   **deadlocks the dispatch** — it builds and folds cleanly and then times out.

### Why the shipped models can't regress

- `DECODE_WGROUP=0` (the default) emits **byte-identical** AIR IR for
  llama-3.2-1b, llama-3.2-3b, gemma3-4b and qwen3-8b.
- With the split on, llama-3.2-3b at G=14 has **1372 `dma_bd` split and unsplit
  alike** — the 1008 weight BDs are re-attributed 448/448/112 across args with
  zero duplication, and the max byte offset per weight arg halves.
- For qwen3-8b at G=9, the op sequences are structurally identical; the only
  differences are the `(arg, offset)` pairs on weight BDs, and all 1143 map
  exactly as `global = group*G*W_LAYER + local`.

`DECODE_STACK` (default 10240, 6144 here) is likewise a no-op when unset; at
K=4096 the seven K-wide L1 activation buffers leave no room for the default core
stack.

## Other changes

- `nightlyPerfBenchmark.yml`'s memory note described a ~32 GB runner; the runner
  has 64 GB now. The measured 3B/4B peaks are kept as upper bounds since they
  were taken on the old configuration. Added a hand-derived bound for the 8B and
  flagged it as not yet measured on that runner.
- `generate_readme.py`, `llms/README.md` and `hf_models.txt` get the new model.
  The lit suites, nightly workflow and weight pre-seeding all discover models by
  path glob or read `hf_models.txt` directly, so nothing else needed wiring.

## Notes for review

- Everything above was measured on an `amdryzenai5pro340`-class Strix box, not
  inferred. The one figure I could not verify locally is the CI runner's memory,
  which is why that comment says the 8B peak is unmeasured rather than quoting a
  number.
- 4 lit tests, matching gemma3_4b_q4nx: compile (decode), compile_prefill,
  profile, verify. The profile lit asserts the three parsed metrics, not just the
  model name, so a drift in the driver's print format can't silently blank the
  published dashboard row.


erwei-xilinx:
Thanks — all four were real gemma-clone residue. Fixed in 6fbbb31.

**`Qwen3Config` geometry.** Correct catch, and the comment above it already
claimed the values came from `qwen3_8b_q4nx_weights` — they didn't. It now
actually imports them, so it can't drift from the loader again:

```
Qwen3Config(n_layers=36, emb_dim=4096, n_heads=32, head_dim=128,
            n_kv_heads=8, hidden_dim=12288, vocab_size=151936)
```

One correction to the finding, for the record: this was **latent, not live**.
`verify_runner.py` reads only `config.n_layers` (line 348), which was 36 and
correct. So no diagnosis or probing behaviour was actually wrong. Worth fixing
regardless — a config that misdescribes the model is a trap for the first caller
that reads another field.

**`requirements.txt`** — header renamed, and the tokenizer note no longer points
at the gated `google/gemma-3-4b` checkpoint (also covers `make chat`).

**`Q4nxModel` docstring** — listed Gemma deltas (4-norm (1+w) fold, dual-theta
rope LUT, embed scale) that the surrounding code explicitly does *not* do. Now
describes what it does: 2-norm pre-norm with un-folded weights, per-head qk-norm,
single-theta rope, untied Q4NX lm_head.

The remaining `Gemma` mentions in `qwen3_8b_q4nx_weights.py` are deliberate
contrast notes (`# NOT Gemma: no (1+w) norm fold, no embedding scale, ...`) and
are kept — they mark the places a reader coming from the Gemma sibling would
otherwise expect a fold.

`make verify` re-run after the change: **2/2 PASS**.

PR #1853, opened 2026-08-21
[python] air.api: packed channel puts, and the two vecmat i8 examples

erwei-xilinx:
Converts `vector_matrix_multiplication/i8/single_core` and its block-quantized
sibling onto `air.api`, replacing the raw-bindings versions. They are the two
remaining siblings of the bf16 vecmat that landed in #1852.

## The one piece of new surface: `pack=` on `Channel.put`/`get`

Both examples stage B through L2 flat and micro-tile it on the way into L1, so
the **DMA** performs the pack by walking the staging buffer out of order.
`ops.load` derives that walk from its destination buffer, which it can see. A
channel cannot: put and get are separate ops, and the packed side is at the far
end of the stream, possibly in another scope. So the walk is named where it
happens:

```python
mm = air.micro_tile(1, 16, 8)
b_l2_l1.put(l2_b[kk : kk + tile_k, :], pack=mm.b(tile_k, tile_n, lead=()))
```

It takes a shape from `air.micro_tile(...).a/.b` and reuses `_pack.pack_pattern`
— the same derivation `ops.load` already uses, not a second implementation.

The block-quantized example's B *scale* is that same walk with one scale per
block of `bs` rather than per element, which is exactly a `k=1` micro-tile. It
reuses the derivation with a different argument rather than needing a special
case, which is the main evidence that the knob is the right shape.

`pack=` on a C operand raises: C unpacks the other way round, with the pattern
on the packed buffer itself.

## Verification, on npu1

|  | i8 | block_quantized_i8 |
|---|---|---|
| channel ops vs predecessor, SSA normalised | identical | identical |
| buffers / locks / flows / dma_bds / tiles / shim allocs / dma_starts / use_locks | 10/12/6/13/6/3/9/40, both arms | 18/16/6/31/6/3/9/88, both arms |
| runtime-sequence access patterns | same set | same set |
| hardware | `PASS!` | `PASS!` |
| lits | peano passes; chess unsupported on this box | peano passes; chess unsupported on this box |
| consumers | none | none |

One deliberate difference: the B pack emits **rank 4** — `[6, 6, 16, 8]` over
`[8, 768, 48, 1]` — where the predecessors hand-collapsed it to rank 3,
`[6, 96, 8]` over `[8, 48, 1]`. Same walk, since `768 = 16 x 48`; the identical
resource counts and runtime patterns confirm it lowers to the same thing.

Regression: all 10 `python/test/api` suites green, and the 15 examples already
on `air.api` emit byte-identical IR. Six new tests — one IR case for the pack
and five negative cases (not a packed shape, not a region, a C operand, wrong
rank, indivisible extent).

## Also fixed

Each example declared `linalg_fill_i32_view16x8xi32as2` as `(scalar, memref)`
and passed a zero the callee never reads. `vm.cc` defines it taking the buffer
alone. Harmless under the C ABI, but untrue, and it is now declared as defined —
the same fix the bf16 sibling took.

## Coverage limits, stated plainly

**These two examples are npu1-only by construction, and no npu2 run is
possible.** I originally wrote that npu2 coverage was being collected
separately; that was wrong, and an npu2 box has since confirmed why:

- All four lits gate on `ryzen_ai_npu1` — including the two chess ones, which
  require `ryzen_ai_npu1, valid_xchess_license`. An npu2 machine can never run
  them however it is licensed.
- Both Makefiles carry only `--target=aie2-none-unknown-elf`. The bf16 sibling
  has a real `AIE_TARGET=aie2p` path and three npu2 lits; these do not.
- Forcing the aie2p triple fails in the microkernels, not in this PR:
  `aie::detail::mmul_8_8<1, 16, 8, signed char, signed char, 32>` has no AIE2P
  specialization, and block_quantized's `aie2_fp32_conf_shim.h` does not compile
  for aie2p (`undeclared identifier '__int128_t'`). Both build cleanly for aie2.

All of that is inherited from the predecessors — the conversion neither adds nor
removes a target. But it means the honest statement is:

| lit | who can run it | status |
|---|---|---|
| `i8` peano, `block_quantized_i8` peano | npu1 + peano | **pass** (this box) |
| `i8` chess, `block_quantized_i8` chess | npu1 + Vitis license | **unverified by anyone** |

The chess pair needs an npu1 box with a Vitis license. CI will not close that
gap, and neither will an npu2 machine. Giving these examples npu2 coverage would
mean writing an aie2p microkernel, which is a separate piece of work.

### DIFF
--- a/programming_examples/vector_matrix_multiplication/bf16/single_core/output.mlir
+++ b/programming_examples/vector_matrix_multiplication/bf16/single_core/output.mlir
@@ -1,0 +1,78 @@
+#map = affine_map<()[s0] -> (s0 * 48)>
+#map1 = affine_map<()[s0] -> (s0 * 96)>
+module {
+  air.channel @aL3ToL2 []
+  air.channel @bL3ToL2 []
+  air.channel @aL2ToL1 []
+  air.channel @bL2ToL1 []
+  air.channel @cL1ToL2 []
+  air.channel @cL2ToL3 []
+  func.func private @linalg_fill_bf16(bf16, memref<48xbf16, 2 : i32>) attributes {link_with = "vm.o", llvm.emit_c_interface}
+  func.func private @vecmat_bf16_bf16(memref<96xbf16, 2 : i32>, memref<96x48xbf16, 2 : i32>, memref<48xbf16, 2 : i32>) attributes {link_with = "vm.o", llvm.emit_c_interface}
+  func.func @vecmat_bf16(%arg0: memref<288xbf16>, %arg1: memref<288x48xbf16>, %arg2: memref<48xbf16>) {
+    %c1 = arith.constant 1 : index
+    %c1_0 = arith.constant 1 : index
+    air.launch (%arg3, %arg4) in (%arg5=%c1, %arg6=%c1_0) args(%arg7=%arg0, %arg8=%arg1, %arg9=%arg2) : memref<288xbf16>, memref<288x48xbf16>, memref<48xbf16> {
+      air.channel.put  @aL3ToL2[] (%arg7[] [] []) : (memref<288xbf16>)
+      %0 = affine.apply #map()[%arg4]
+      %c0 = arith.constant 0 : index
+      %c288 = arith.constant 288 : index
+      %c48 = arith.constant 48 : index
+      %c48_1 = arith.constant 48 : index
+      %c1_2 = arith.constant 1 : index
+      air.channel.put  @bL3ToL2[] (%arg8[%c0, %0] [%c288, %c48] [%c48_1, %c1_2]) : (memref<288x48xbf16>)
+      %c48_3 = arith.constant 48 : index
+      %c1_4 = arith.constant 1 : index
+      air.channel.get  @cL2ToL3[] (%arg9[%0] [%c48_3] [%c1_4]) : (memref<48xbf16>)
+      air.segment @vecmat_i8_0  {
+        %alloc = memref.alloc() : memref<288xbf16, 1 : i32>
+        air.channel.get  @aL3ToL2[] (%alloc[] [] []) : (memref<288xbf16, 1 : i32>)
+        %alloc_5 = memref.alloc() : memref<288x48xbf16, 1 : i32>
+        air.channel.get  @bL3ToL2[] (%alloc_5[] [] []) : (memref<288x48xbf16, 1 : i32>)
+        %c0_6 = arith.constant 0 : index
+        %c3 = arith.constant 3 : index
+        %c1_7 = arith.constant 1 : index
+        scf.for %arg10 = %c0_6 to %c3 step %c1_7 {
+          %1 = affine.apply #map1()[%arg10]
+          air.channel.put  @aL2ToL1[] (%alloc[] [] []) : (memref<288xbf16, 1 : i32>)
+        }
+        %c0_8 = arith.constant 0 : index
+        %c3_9 = arith.constant 3 : index
+        %c1_10 = arith.constant 1 : index
+        scf.for %arg10 = %c0_8 to %c3_9 step %c1_10 {
+          %1 = affine.apply #map1()[%arg10]
+          air.channel.put  @bL2ToL1[] (%alloc_5[] [] []) : (memref<288x48xbf16, 1 : i32>)
+        }
+        %alloc_11 = memref.alloc() : memref<48xbf16, 1 : i32>
+        air.channel.get  @cL1ToL2[] (%alloc_11[] [] []) : (memref<48xbf16, 1 : i32>)
+        %c1_12 = arith.constant 1 : index
+        %c1_13 = arith.constant 1 : index
+        air.herd @herd_0  tile (%arg10, %arg11) in (%arg12=%c1_12, %arg13=%c1_13) attributes {link_with = "vm.o"} {
+          %alloc_14 = memref.alloc() : memref<48xbf16, 2 : i32>
+          %cst = arith.constant 0.000000e+00 : bf16
+          func.call @linalg_fill_bf16(%cst, %alloc_14) : (bf16, memref<48xbf16, 2 : i32>) -> ()
+          %c0_15 = arith.constant 0 : index
+          %c288_16 = arith.constant 288 : index
+          %c96 = arith.constant 96 : index
+          scf.for %arg14 = %c0_15 to %c288_16 step %c96 {
+            %alloc_17 = memref.alloc() : memref<96xbf16, 2 : i32>
+            air.channel.get  @aL2ToL1[] (%alloc_17[] [] []) : (memref<96xbf16, 2 : i32>)
+            %alloc_18 = memref.alloc() : memref<96x48xbf16, 2 : i32>
+            air.channel.get  @bL2ToL1[] (%alloc_18[] [] []) : (memref<96x48xbf16, 2 : i32>)
+            func.call @vecmat_bf16_bf16(%alloc_17, %alloc_18, %alloc_14) : (memref<96xbf16, 2 : i32>, memref<96x48xbf16, 2 : i32>, memref<48xbf16, 2 : i32>) -> ()
+            memref.dealloc %alloc_17 : memref<96xbf16, 2 : i32>
+            memref.dealloc %alloc_18 : memref<96x48xbf16, 2 : i32>
+          }
+          air.channel.put  @cL1ToL2[] (%alloc_14[] [] []) : (memref<48xbf16, 2 : i32>)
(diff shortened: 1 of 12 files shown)


erwei-xilinx:
**npu1 perf A/B — block_quantized may regress; asking npu2 to arbitrate before this merges.**

5 alternating reps, 20 in-process iterations each, both arms instrumented
identically with `XRTRunner(n_perf_iters=20)` and sharing one `vm.o`. Power mode
`Default` (turbo needs root, refused on this box).

| arm | best | median | spread |
|---|---|---|---|
| i8 predecessor | 175.2 | 205.9 | 40.4 |
| i8 **port** | 186.5 | 194.7 | 92.1 |
| block_quantized predecessor | 186.7 | 190.5 | 12.1 |
| block_quantized **port** | 198.7 | 214.8 | 40.3 |

`i8` is noise — best says +6.4%, median says −5.4%, and the port's spread is 92 us
on a ~200 us kernel. Parity.

`block_quantized` is less comfortable. Every port rep is slower than every
predecessor rep bar one 198.8/198.7 overlap: best +6.4%, median +12.8%, against a
predecessor spread of only 12 us. I am not claiming parity on that.

**It is not the pack.** The rank-4 pattern `pack=` emits canonicalizes back to
the predecessor's hand-collapsed rank-3 BD, identically:

```
pred:  air.channel.put @bL2ToL1[] (%X[0, 96, 0] [6, 96, 8] [8, 48, 1])
port:  air.channel.put @bL2ToL1[] (%X[0, 96, 0] [6, 96, 8] [8, 48, 1])
```

and the B-scale puts differ only in which dimension carries the offset (`[0,3,0]`
against stride 48 vs `[0,0,144]` against stride 1 — the same 144 elements in).
All lowered resource counts match exactly.

What does differ is the async dependency structure, from the one change the DSL
forces: the L3-side put and get move inside the segment, because outside one
there is no shim DMA allocation to link to. The port's herd carries an extra
dependency (`air.herd async [%X]` vs bare `async`) and the segment-end
`air.wait_all` has 7 operands against 5. That is the first place to look. The
same move went the other way on the bf16 sibling — 251 → 201 us — so it is not a
known cost.

npu2 numbers are being collected. If they show the same one-sided gap I will
chase it before this merges; if they show parity, npu1 goes down as noisy.

erwei-xilinx:
**Correction: the block_quantized regression I reported above does not exist. It was my measurement, not the kernel.**

I re-ran it as a controlled A/B — one driver process, same inputs, same `vm.o`,
same `n_perf_iters=20`, the *only* variable being which `.mlir` is parsed and
run. 8 reps:

| variant | best | median | mean | spread |
|---|---|---|---|---|
| predecessor | 214.3 | 234.6 | 307.1 | **509.8** |
| **port** | 185.3 | 213.3 | 211.9 | 56.4 |
| port, herd operands stripped | 196.6 | 211.3 | 209.2 | 20.5 |

The port is faster in **8 of 8 paired reps** — median −9.1%, best −13.5%. No
regression on either example.

Two things went wrong in the run that produced the earlier table, and both are
worth recording:

1. **The script did not alternate.** It ran `old` then `new` in that fixed order
   every rep, which pairs each arm with a fixed position in the pair rather than
   randomising it. Fixed: the order now swaps on alternate reps.
2. **The predecessor's own distribution is very heavy-tailed** — 509 us of
   spread across 8 reps, including a 724 us outlier, against 56 us for the port.
   Five fixed-order samples of that distribution are simply not enough to
   compare medians, and I treated 5/5 one-sided as signal when it was not.

Also settled in passing: stripping the herd operands the DSL passes changes
nothing measurable here (−0.9%, 5/8 wins), so the extra async dependencies I
pointed at in the earlier comment are not costing anything on this example.

Review comments addressed in 3d115df0 — the `pack=` TypeError no longer suggests
`.c` when C is rejected, and both examples stop describing themselves as the
"passthrough_dma example" in `--help`.

erwei-xilinx:
**Raw output, per the new bar: claims replaced by the commands that back them.**

Also: **the seven scratch files are gone.** `git add -A` under a path in the
review-fix commit swept untracked `.mlir` scratch that was already sitting in
those directories — including one from #1852's example, which this PR does not
touch. That commit is amended; the file list below is the check.

```
$ git log --oneline origin/main..HEAD
47ffcd98 [programming_examples] Address review on the vecmat i8 port
1853f158 [python] air.api: packed channel puts, and the two vecmat i8 examples

$ git show --name-status HEAD
M	programming_examples/vector_matrix_multiplication/block_quantized_i8/single_core/single_core.py
M	programming_examples/vector_matrix_multiplication/i8/single_core/single_core.py
M	python/air/api/_channel.py

$ git show --name-status HEAD~1
M	programming_examples/vector_matrix_multiplication/block_quantized_i8/single_core/single_core.py
M	programming_examples/vector_matrix_multiplication/i8/single_core/single_core.py
M	python/air/api/_channel.py
M	python/test/api/channel.py
M	python/test/api/errors.py

$ git diff --stat origin/main...HEAD
 .../block_quantized_i8/single_core/single_core.py  | 554 +++++++--------------
 .../i8/single_core/single_core.py                  | 415 ++++++---------
 python/air/api/_channel.py                         |  70 ++-
 python/test/api/channel.py                         |  49 ++
 python/test/api/errors.py                          |  80 +++
 5 files changed, 509 insertions(+), 659 deletions(-)

$ grep -H 'REQUIRES:' programming_examples/vector_matrix_multiplication/{i8,block_quantized_i8}/single_core/*.lit
programming_examples/vector_matrix_multiplication/block_quantized_i8/single_core/run_makefile_chess.lit:// REQUIRES: ryzen_ai_npu1, valid_xchess_license
programming_examples/vector_matrix_multiplication/block_quantized_i8/single_core/run_makefile_peano.lit:// REQUIRES: ryzen_ai_npu1, peano
programming_examples/vector_matrix_multiplication/i8/single_core/run_makefile_peano.lit:// REQUIRES: ryzen_ai_npu1, peano
programming_examples/vector_matrix_multiplication/i8/single_core/run_makefile_chess.lit:// REQUIRES: ryzen_ai_npu1, valid_xchess_license
```

```
$ lit -sv ... --filter 'vector_matrix_multiplication.*(i8|block_quantized)'   [per-test lines]
UNSUPPORTED: AIR_PROGRAMMING_EXAMPLES :: vector_matrix_multiplication/block_quantized_i8/single_core/run_makefile_chess.lit (1 of 4)
UNSUPPORTED: AIR_PROGRAMMING_EXAMPLES :: vector_matrix_multiplication/i8/single_core/run_makefile_chess.lit (2 of 4)
PASS: AIR_PROGRAMMING_EXAMPLES :: vector_matrix_multiplication/i8/single_core/run_makefile_peano.lit (3 of 4)
PASS: AIR_PROGRAMMING_EXAMPLES :: vector_matrix_multiplication/block_quantized_i8/single_core/run_makefile_peano.lit (4 of 4)

$ lit -v build/python/test/ --filter=api   [per-test lines]
PASS: AIRPYTHON :: api/leaky_relu.py (1 of 10)
PASS: AIRPYTHON :: api/pack.py (2 of 10)
PASS: AIRPYTHON :: api/segment.py (3 of 10)
PASS: AIRPYTHON :: api/axpy.py (4 of 10)
PASS: AIRPYTHON :: api/relu.py (5 of 10)
PASS: AIRPYTHON :: api/eltwise_add.py (6 of 10)
PASS: AIRPYTHON :: api/activations.py (7 of 10)
PASS: AIRPYTHON :: api/channel.py (8 of 10)
PASS: AIRPYTHON :: api/dot.py (9 of 10)
PASS: AIRPYTHON :: api/errors.py (10 of 10)
```

The two `UNSUPPORTED` are the chess lits — no Chess on this box. As the body
now says, **no machine in reach can run them**: they gate on
`ryzen_ai_npu1, valid_xchess_license`, so an npu2 host is excluded by the
`REQUIRES` line above regardless of licensing. They stay unverified.

Hardware, both examples, on npu1 via `make run PEANO_INSTALL_DIR=...`:

```
i8:                 PASS!
block_quantized_i8: PASS!
```

Perf, block_quantized, 8 reps, one driver process, same inputs, same `vm.o`,
only the parsed `.mlir` differing:

```
                 best   median   mean    spread
predecessor     214.3   234.6   307.1    509.8
port            185.3   213.3   211.9     56.4
```

Port faster in 8/8 paired reps. (An earlier comment in this thread reported a
regression from a fixed-order A/B; that is retracted above.)

erwei-xilinx:
Closing in favour of a clean re-submission. This branch's history carries 793 lines of debug scratch that `git add -A` swept in, and amending it in place leaves a PR whose review thread contradicts its own diff — including a retracted perf regression and a retracted npu2 coverage claim. The replacement carries the same code with clean history and the raw-output evidence bar from the outset. The broadcast-index bug found alongside this work is separately in #1855.

PR #1854, opened 2026-08-21
[llms] Add Qwen2.5-7B Q4NX (prefill + fused decode)

erwei-xilinx:
Qwen2.5-7B-Instruct end to end on NPU2: an AIR prefill over the padded context, and a fused decode that dispatches **all 28 layers plus the LM head in one call**.

Two firsts for the shared [`fused_decode`](programming_examples/fused_decode/) engine: the first `HAS_QKV_BIAS` model, and the first with no FastFlowLM bundle to load — weights are Q4NX quantized on load from the ungated `Qwen/Qwen2.5-7B-Instruct`, which is also the bf16 verify reference.

## Engine fixes

Both are no-ops for the five models already on this engine: **their emitted IR is byte-identical before and after** (checked by re-emitting each model's module on both revisions).

**1. `refeed()` splits counts above 63.** An AIE2/AIE2P lock counter is 6 bits (`AIETargetModel::getMaxLockValue() == 0x3F`), and the refeed count becomes a lock init that nothing range-checks. This model's gate-up refeed is 74, which truncated on the device and deadlocked the re-broadcast. Counts ≤ 63 — every model shipped so far — emit exactly the IR they did before.

**2. The GLU slice granularity adapts so the slice count stays even.** The GLU core's buffer-slot sequence is statically unrolled and restarts every layer, while the BD ring's rotation carries over — so an *odd* slice count leaves the two a slot apart and every other layer reads a stale slice. `INTERMEDIATE_SIZE=18944` is the first that is an odd number of 1024-row slices (37); dropping to one demux packet per slice makes it 74, same total, same interleave stride. An assert now states the invariant.

The failure was reproducible as exact dispatch-parity (`OK,BAD,OK,BAD` on repeated identical dispatches), which is what pinned it to ring phase rather than to the data.

## Also here

- Conditional SwiGLU row split in the shared `qwen25_3b` prefill builder: at hidden=18944 a single launch needs 1024 DMA iterations and exhausts the shim's 16 BDs. Returns the single unchanged chunk for every smaller `hidden_dim`.
- Vectorized affine Q4NX packer (bit-identical to the reference packer, ~1000x faster).
- Five GEMM kernel-registry entries.

## Results

Measured on an **AMD Ryzen AI 9 HX 370** (Strix Point, XDNA2/NPU2), warm:

| | |
|---|---|
| prefill (TTFT) | 10.0 s at CTX=2048 (205 tok/s) |
| decode | 11.46 tok/s (87.2 ms/token), ATTN_MAXL=2048 |

- `make verify` — PASS, top-k token-set gate vs `Qwen/Qwen2.5-7B-Instruct` bf16 (2 prompts x 32 tokens, k=5)
- `make run` — `*** PARIS ***`, reproducible

The example is wired into the nightly LLM perf CI and the GitHub Pages dashboard the same way its siblings are: path-globbed `run_npu2_{verify,profile}.lit`, an `hf_models.txt` entry for the weight pre-fetch, and an `LLM_HF_MODELS` entry for the dashboard link.

### DIFF
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -54,10 +54,14 @@ env:
   # nightly rather than assuming.
   #
   # qwen25_7b_q4nx runs off Qwen/Qwen2.5-7B-Instruct directly (quantized on load --
-  # FastFlowLM publishes no Q4NX bundle for it), so it is the same shape as qwen3_8b
-  # above and its peak is likewise unmeasured here. Bound: ~15GB bf16 reference in
-  # the compare phase, ~6GB dequantized Q4NX weights plus five decode weight BOs
-  # (4x0.95GB + 0.32GB, see llms/qwen25_7b_q4nx) in the NPU phase.
+  # FastFlowLM publishes no Q4NX bundle for it). Its NPU phase peaks at ~33GB RSS,
+  # MEASURED (32.6GB for prefill alone on a 96GB dev box): the bf16 checkpoint, the
+  # Q4NX host dequant and the five resident decode weight BOs (4x0.95GB + 0.32GB)
+  # are all live across the load. That is the largest peak of any model here and
+  # over half this 64GB runner, so it must not run concurrently with another model
+  # -- the -j1 in the lit suites is what guarantees that. The compare phase holds
+  # the ~15GB bf16 reference in a separate subprocess, so it is not additive.
+  # Re-measure this if the weight preload's lifetime changes.
 
 jobs:
   benchmark:
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -52,6 +52,12 @@ env:
   # five decode weight BOs (4x1.01GB + 0.37GB, see llms/qwen3_8b_q4nx). Separate
   # subprocesses, so the peak is the larger, not the sum. Confirm against a
   # nightly rather than assuming.
+  #
+  # qwen25_7b_q4nx runs off Qwen/Qwen2.5-7B-Instruct directly (quantized on load --
+  # FastFlowLM publishes no Q4NX bundle for it), so it is the same shape as qwen3_8b
+  # above and its peak is likewise unmeasured here. Bound: ~15GB bf16 reference in
+  # the compare phase, ~6GB dequantized Q4NX weights plus five decode weight BOs
+  # (4x0.95GB + 0.32GB, see llms/qwen25_7b_q4nx) in the NPU phase.
 
 jobs:
   benchmark:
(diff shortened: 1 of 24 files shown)


erwei-xilinx:
Thanks — addressed in 059209de. Triaged each comment against the code first; some were real, some describe conventions shared by all six sibling examples.

### Fixed

**Weight-source override (moderate).** Correct and a real bug. The Makefile exports `Q4NX_MODEL_SOURCE` but the standalone prefill driver read `MODEL_SOURCE`, so `make prefill MODEL_SOURCE=...` and `make profile-prefill MODEL_SOURCE=...` silently loaded the default while `make run` honored it. This was a divergence from the Qwen3-8B sibling, which already reads `Q4NX_MODEL_SOURCE`. Now matches it, with `MODEL_SOURCE` as a fallback; verified all four precedence cases plus that `make` exports the value.

**Stale bartowski/GGUF provenance (5 sites).** Correct. Left over from an earlier revision that consumed a converted GGUF; the default now quantizes the HF checkpoint on load. Updated in the driver, prefill, verify adapter, verify lit and Makefile.

**Host memory.** Correct, and worse than the review estimated. I measured it rather than estimating: **peak RSS 32.6 GB** for the prefill path alone, against a documented ~5 GB. Fixed in the prerequisites and in the nightly workflow's memory comment, where it matters — that is over half the 64 GB runner, so I also noted that `-j1` is what keeps it from running concurrently with another model.

**Prefill perf inconsistency.** Correct — I had mixed three measurement scopes from two runs. Re-measured in one run: 7.78 s NPU execution, 8.50 s instrumented XRT time, 9.59 s wall TTFT. The section now gives all three and says what each includes.

**fused_decode codec table.** Correct, and introduced by my one-line edit. The table splits by codec, not model family, so listing Qwen2.5-7B above it made the "Qwen2.5" column look like it covered both. Columns are now labelled by builder, with 7B explicitly on the Q4NX side.

**.gitignore.** Correct — `compile-decode-dynseq` writes `decode_dyn_*` here and they were untracked. Added.

### Not changed, deliberately

These are all identical in the sibling examples, so fixing them only here would make this example inconsistent with five others without fixing the class of bug:

- `build_runner` ignoring `model_name` and `npu_attn` — byte-identical to `qwen3_8b_q4nx/verify_adapter.py`. Real sharp edge, but it belongs in a change to the shared adapter contract.
- `attn_maxl` bound not reserving the decode budget — same `>` check in the sibling.
- `DECODE_WGROUP` hardcoded at runtime — the sibling hardcodes `9` the same way; both rely on the `.decode_flags` build stamp.
- Requant cache keyed only on `W_DUAL_CHAN`, not the model source — the sibling's key is the same.

Happy to do any of them as a follow-up across all the q4nx examples if you'd prefer that; they're genuine, just not this PR's scope.

### Re-verified after the changes

`make run` → `*** PARIS ***`, `[12095, 13, 1084, 374, 7407, 304, 279, 18172, 949, 315]`. The prefill run that produced the new perf numbers passed its own Paris gate. All five previously-shipped models still emit byte-identical IR.

PR #1855, opened 2026-08-21
[python] air.api: bound broadcast channel indices by broadcast_shape

erwei-xilinx:
Fixes a bug shipped in #1852.

A broadcast channel is indexed over its **destinations**, not its sources.
`size=[1, 1], broadcast_shape=[3, 1]` is one producer feeding a 3x1 grid, and
the consumers name slots `[0,0]`, `[1,0]`, `[2,0]`. The index check bounded
them by `size`, which admits only `[0, 0]`, so it rejected the canonical
fan-out:

```
air.channel 'ChanIn' index 1 is out of range on axis 0: size is [1, 1]
```

which reads as a contradiction when the three-way broadcast is declared exactly
as intended. Now bounded by `broadcast_shape` when there is one, and the
message names which bound it applied.

`channel_examples/broadcast/single_herd`, converted in #1852, missed this
because its indices are herd coordinates — dynamic, and so never range-checked.
The bug surfaced converting `broadcast/multi_herd`, whose three herds index
with constants.

## Evidence

```
$ git log --oneline origin/main..HEAD
0e3c8928 [python] air.api: bound broadcast channel indices by broadcast_shape

$ git show --name-status HEAD
M	python/air/api/_channel.py
M	python/test/api/errors.py

$ git diff --stat origin/main...HEAD
 python/air/api/_channel.py | 19 +++++++++++++++----
 python/test/api/errors.py  | 17 +++++++++++++++++
 2 files changed, 32 insertions(+), 4 deletions(-)

$ lit -v build/python/test/ --filter=api
PASS: AIRPYTHON :: api/segment.py (1 of 10)
PASS: AIRPYTHON :: api/leaky_relu.py (2 of 10)
PASS: AIRPYTHON :: api/pack.py (3 of 10)
PASS: AIRPYTHON :: api/axpy.py (4 of 10)
PASS: AIRPYTHON :: api/relu.py (5 of 10)
PASS: AIRPYTHON :: api/channel.py (6 of 10)
PASS: AIRPYTHON :: api/eltwise_add.py (7 of 10)
PASS: AIRPYTHON :: api/dot.py (8 of 10)
PASS: AIRPYTHON :: api/activations.py (9 of 10)
PASS: AIRPYTHON :: api/errors.py (10 of 10)
```

New negative case `broadcast_index_bounded_by_broadcast_shape` in
`errors.py` pins the bound and the message. No hardware run: this is a
trace-time check, and the emitted IR for an in-range index is unchanged.

### DIFF
--- a/python/air/api/_channel.py
+++ b/python/air/api/_channel.py
@@ -158,7 +158,7 @@ class Channel:
             )
         self._declared = True
 
-    def _indices(self, indices):
+    def _indices(self, indices, direction):
         if indices is None:
             indices = ()
         if isinstance(indices, int):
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
**Over-correction in the first commit, fixed in `6a4b1305`** — caught in review on the stacked #1857.

`_indices` is shared by `put` and `get`, so widening the bound to
`broadcast_shape` applied it to *both* ends. But the two ends index different
grids:

- a **get** names a destination → `broadcast_shape` is right. With
  `size=[1, 1], broadcast_shape=[3, 1]` the consumers are `[0,0]`, `[1,0]`,
  `[2,0]`, and bounding by `size` admits only `[0,0]` — the bug this PR exists
  to fix.
- a **put** names a slot in the source bundle → `size`. For
  `size=[1, 2], broadcast_shape=[3, 2]`, `put(indices=[2, 0])` is outside the
  source and my first commit accepted it, emitting an invalid bundle index.

Direction is now threaded into `_indices`, and the error names whichever bound
it applied. New negative case `broadcast_put_still_bounded_by_size` covers the
put side; the existing get-side case is unchanged and still passes.

```
$ git log --oneline origin/main..HEAD
6a4b1305 [python] air.api: bound broadcast puts by size, gets by broadcast_shape
0e3c8928 [python] air.api: bound broadcast channel indices by broadcast_shape

$ git show --name-status HEAD
M	python/air/api/_channel.py
M	python/test/api/errors.py

$ lit -v build/python/test/ --filter=api
PASS: 10 of 10   (api/errors.py and api/channel.py among them)
```

PR #1859, opened 2026-08-21
[mlir] Index channel puts/gets instead of re-walking the module

erwei-xilinx:
`air-fuse-channels` dominated compile time on merged multi-launch modules: **114.5 s of a 121.2 s AIR pipeline** for an 84-channel LLM prefill kernel (`o_ffn` in `llms/llama32_1b_q4nx`) — while producing **byte-identical IR**.

## Root cause

`getChannel{Put,Get}OpThroughSymbol` answers "which puts/gets belong to this channel?" by walking the whole scope, and the pass's `O(N²)` channel-pair loop calls them **16× per pair**. Measured with gdb breakpoint counts on that module:

| | |
|---|---|
| `getChannelPutOpThroughSymbol` calls | 27,888 |
| `getChannelGetOpThroughSymbol` calls | 27,888 |
| full module walks | **55,776** |
| channel pairs (84 choose 2) | 3,486 |
| walks per pair | exactly 16.0 |
| op-visits | ~81 M |

All to recompute a mapping that does not change from one pair to the next.

## Changes

1. **`air::ChannelIndex`** — one walk of a scope, bucketing puts/gets by channel symbol. Documented as a snapshot, with `rebuild()`/`clear()` for callers that mutate. Same traversal order as the free functions, so callers indexing `puts[0]`/`gets[0]` are unaffected.

2. **Adopted in `AIRFuseChannels`** via member overloads that shadow the free functions, so the existing unqualified call sites redirect by name lookup with no edits at the call sites. The index is invalidated at all five IR-mutating points (LB/UB merge, dummy-loop wrap, region wrap, op erase, spatial merge), so it cannot go stale.

3. **Dropped `ForwardDominanceIterator`** from both free functions. Collecting ops by symbol name does not depend on dominance order, and that iterator is why a walk of a 1452-op module cost ~2.6 ms rather than the traversal itself. This also helps the other call sites in `AIRToAIEPass`, `AIRMiscPasses`, `AIRToAIESchedulingUtils` and `AIRGpuChannelToCachelinePass`.

## Results

| | before | after |
|---|---|---|
| `air-fuse-channels` | 114.5 s | **0.15 s** |
| AIR pipeline (same kernel) | 121.2 s | **4.2 s** |
| `aircc` end to end | 155 s | **37 s** |

It was 94.5% of the pipeline; it is now 3.5% and 6th-ranked.

## Equivalence

Three independent checks:

- the pass's own output is byte-identical on the repro input;
- **all 62 per-pass IR dumps** of the full `aircc` pipeline are byte-identical to the pre-change baseline, so nothing downstream shifted either;
- `check-air-mlir` unchanged: 510 passed, 7 unsupported, 7 XFAIL, **0 failures**.

## Reviewer note

Change 3 is the one with residual risk: dropping the dominance iterator can change visit order in *multi-block* regions, and some callers index `puts[0]`. For structured control flow (single-block regions) the order is identical, which is consistent with every dump matching — but this IR may simply not exercise the multi-block case. Changes 1–2 are provably order-preserving and capture ~97% of the win on their own (121 s → 4.5 s), so 3 can be dropped if you would rather be conservative.

Repro:
```bash
air-opt <prefill-module>.mlir -air-fuse-channels -o /dev/null   # 145s before, 0.2s after
```

### DIFF
--- a/mlir/include/air/Util/Util.h
+++ b/mlir/include/air/Util/Util.h
@@ -164,6 +164,41 @@ int64_t getRefeedCount(ChannelInterface op);
 // Same, for a value already resolved to its buffer/alloc carrier (memtile L2
 // rendezvous buffer): reads air.refeed_count directly off `op`.
 int64_t getRefeedCount(Operation *op);
+// One-shot index of a scope's channel puts/gets, keyed by channel symbol.
+//
+// getChannel{Put,Get}OpThroughSymbol walks the entire scope on EVERY call, so a
+// pass that queries it inside a loop pays O(queries * scope size). Build this
+// once instead and query it. Measured on an 84-channel prefill module,
+// air-fuse-channels issued 55,776 such walks and spent 114s doing so; with the
+// index that pass drops to 0.15s.
+//
+// The index is a SNAPSHOT: it holds raw Operation pointers into `scope`, so any
+// mutation that adds, removes or reparents a channel put/get invalidates it.
+// Call rebuild() after such a mutation (or use it only over a read-only phase).
+// The visit order matches getChannel{Put,Get}OpThroughSymbol, so callers that
+// index puts[0]/gets[0] see the same op either way.
+class ChannelIndex {
+public:
+  ChannelIndex() = default;
+  explicit ChannelIndex(Operation *scope) { rebuild(scope); }
+
+  void rebuild(Operation *scope);
+  void clear();
+  // True once rebuild() has run for `scope`; cheap staleness guard for callers
+  // that keep one index across a phase.
+  bool isBuiltFor(Operation *scope) const { return builtScope == scope; }
+
+  ArrayRef<ChannelPutOp> getPuts(ChannelOp channel) const;
+  ArrayRef<ChannelGetOp> getGets(ChannelOp channel) const;
+  ArrayRef<ChannelPutOp> getPuts(StringRef channelName) const;
+  ArrayRef<ChannelGetOp> getGets(StringRef channelName) const;
+
+private:
+  llvm::DenseMap<StringRef, std::vector<ChannelPutOp>> puts;
+  llvm::DenseMap<StringRef, std::vector<ChannelGetOp>> gets;
+  Operation *builtScope = nullptr;
+};
+
 // Get ChannelPutOps from ChannelOp
 std::vector<ChannelPutOp>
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Good catch — this was a real bug, fixed in 7468d0ed.

`renameSymbols` rewrites `chan_name` attributes, and those attributes are the `ChannelIndex`'s **keys**, so every cached name→ops bucket is stale afterwards. The aggressive merge loop sits between the two `renameSymbols` calls and queries the index, so it could have consulted a stale mapping.

I invalidated **inside the lambda** rather than at the two call sites, so a future caller can't forget. I also took the second half of your point — the index is a pass member — and now clear it on entry to `runOnFunction` rather than relying on the scope guard to notice a new function.

## Important caveat on this PR's evidence

Your comment exposes a gap in how I validated the parent commit, and I'd rather state it than let the "62/62 dumps byte-identical" line imply more coverage than it has:

**The module I measured performs zero merges.** `air-fuse-channels` returns byte-identical IR on it, which means `chan_merge_map` ends up empty — so `renameSymbols` rewrites nothing, and none of the LB/UB, NFL, spatial-merge or rename paths execute. **Every cache-invalidation path in this PR is unexercised by my equivalence check**, including the one you just found. That is precisely why it was findable by reading and not by my testing.

`check-air-mlir` (510 passed, 0 failures) is also weaker evidence than it looks here: if those tests covered the merge paths at realistic channel counts, this pass could not have been silently burning 114 s.

So the correct reading of this PR today is: the fast path (no merges) is verified byte-for-byte end to end; the merging paths are reviewed but not tested.

**What I think it needs before merge:** run `aircc --debug-ir` over the decode templates and prefill modules of the q4/q4nx models with and without the patch and diff all pass dumps, specifically looking for a model where `pass_NNN_after_air-fuse-channels.mlir` differs from the dump preceding it. That proves merges are live and exercises the invalidation. It is CPU-only, no NPU needed. I have not run it.

Also flagging that the third change here — dropping `ForwardDominanceIterator` — can reorder visits in multi-block regions where callers index `puts[0]`. Structured single-block control flow is unaffected, consistent with the dumps matching, but this IR may simply not contain that case. Changes 1–2 are provably order-preserving and give ~97% of the win (121 s → 4.5 s) if you'd rather split it out.

PR #1865, opened 2026-08-21
[aircc] Lower each device's cores once, not once per core

erwei-xilinx:
[aircc] Lower each device's cores once, not once per core

aiecc can lower a device's cores together into one shared object and link
each core's ELF against it (`--unified`), or lower every core separately
(`--no-unified`, its default). aircc has never asked for either, so every
build takes the per-core path and runs the same pass pipeline once per core.

It is the same pipeline either way. Both modes call `loweringPipeline`, with
the same passes in the same order; the only difference is the core filter in
`AIECoreToStandardPass`:

    if ((tileRow != row && tileRow != -1) || (tileCol != col && tileCol != -1))
      return failure();   // skip this core

Per-core passes a real (col, row) and discards every other core's work;
unified passes (-1, -1), which disables the filter. So the per-core path is
not doing different work, it is doing the same work N times and throwing away
N-1 of each result. On an 8-sub-kernel LLM ELF that is 192 runs where 13 --
one per aie.device -- suffice.

Peano only. The conditional mirrors the -j one above it: Chess keeps the
per-core objects it has always had, since that path is not exercised here.

## Measurements

Ryzen AI 9 HX 370 (Strix, npu2), 24 CPUs, Peano, on top of #1863.

aiecc alone, same npu.in.mlir (o_ffn, 192 cores / 13 devices), two runs each:

| mode | run 1 | run 2 |
|---|---|---|
| per-core (today) | 25.5 s | 25.5 s |
| unified | 9.1 s | 9.0 s |

The saving is where the redundancy was: the lowering node drops from 17479 ms
over 192 cores to 1439 ms over 13 devices.

Clean `make compile`, cumulative with #1863:

| | llama32_1b_q4nx | llama32_3b_q4nx |
|---|---|---|
| before #1863 | 108.8 s | 166.4 s |
| #1863 | 81.1 s | 87.1 s |
| this PR | 47.2 s | 49.0 s |

Per-ELF on the 1B: o_ffn 41.7 -> 18.9 s, rms_gemms_rope 16.8 -> 8.0 s,
lm_head_gemv 11.3 -> 8.6 s.

## Correctness

Output is not merely equivalent, it is the same code. Comparing the two modes
on the o_ffn design:

- `.text` is bit-identical in 192/192 core ELFs
- every per-core ELF is byte-for-byte the same size in both modes
- the only sections that differ are .symtab/.strtab, by two local
  basic-block labels (.LBB9_1, .LBB9_2) -- block numbering shifts when a
  function is compiled beside its siblings. Same FUNC symbol count.
- the packaged full ELF differs only in .note.xrt.UID and the .pdi.* headers,
  the same two fields that two builds of one source already differ in

`--gc-sections` is what keeps each core's ELF to its own code; the shared
object is 5788 B against 2672 B for a single core's, and it is one per device
rather than one per core.

On hardware:

| check | result |
|---|---|
| `check-air-mlir` | 525 tests, 511 passed, 0 unexpected failures |
| `llama32_1b_q4nx` verify, clean -> compile -> verify | PASS 2/2 |
| GEMV bf16, all 4 npu2 lit configs via `make run` | PASS |
| `check-air-e2e-peano` | 26 passed / 42 failed, identical to main |

The e2e suite's 42 failures are pre-existing and environmental -- it invokes
python3.12 against a cpython-313 build and dies in `import air.ir` before
aircc runs. Same 26/42 split on main with this commit stashed.

Not covered: the Chess path (no Vitis on this host, so the
valid_xchess_license lits are UNSUPPORTED) and npu1. The conditional leaves
Chess on --no-unified, which is the default it already got.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/tools/aircc/aircc.cpp
+++ b/tools/aircc/aircc.cpp
@@ -1454,9 +1454,12 @@ static LogicalResult runAieCompilation() {
     aieccCmd.push_back("-j");
     aieccCmd.push_back((xchesscc || xbridge) ? "1" : "0");
     // Lower each device's cores once instead of once per core; every core still
-    // links its own ELF from the shared object. Peano only -- the Chess object
-    // path is untested here.
-    aieccCmd.push_back((xchesscc || xbridge) ? "--no-unified" : "--unified");
+    // links its own ELF from the shared object. Asked for on the Peano path
+    // only, and asked for rather than stated on both: the two drivers disagree
+    // on the default (C++ aiecc per-core, aiecc.py unified), so naming Chess's
+    // half here would change it on one of them.
+    if (!(xchesscc || xbridge))
+      aieccCmd.push_back("--unified");
 
     // bf16 emulation
     if (bf16Emulation)
--- a/tools/aircc/aircc.cpp
+++ b/tools/aircc/aircc.cpp
@@ -1453,6 +1453,10 @@ static LogicalResult runAieCompilation() {
     // aiecc's default, which has flipped before (mlir-aie 1ba5b5c).
     aieccCmd.push_back("-j");
     aieccCmd.push_back((xchesscc || xbridge) ? "1" : "0");
+    // Lower each device's cores once instead of once per core; every core still
+    // links its own ELF from the shared object. Peano only -- the Chess object
+    // path is untested here.
+    aieccCmd.push_back((xchesscc || xbridge) ? "--no-unified" : "--unified");
 
     // bf16 emulation
     if (bf16Emulation)


erwei-xilinx:
The npu1 comment was right, and chasing it down says this should not land as written. Withdrawing to draft.

## What the review caught

**npu1 coverage.** Enabling `--unified` for every Peano target regresses npu1 badly. `matrix_multiplication/bf16 run4x4 AIE_TARGET=aie2 COMPILE_MODE=compile-only`, flag toggled by shim so nothing else moves, two reps each:

| target | per-core | unified | |
|---|---|---|---|
| npu2, 192 cores / 13 devices (LLM o_ffn) | 25.5 s | 9.0 s | 2.8x faster |
| npu2, 16 cores / 1 device (matmul 4x4) | 4.2 s | 4.5 s | 7% slower |
| npu1, 16 cores / 1 device (matmul 4x4) | 4.1 s | **11.2 s** | **2.7x slower** |

## Why

`--unified` bundles two changes, and only one of them is the one I wanted. aiecc node timings for the npu1 case:

| | per-core | unified |
|---|---|---|
| `objects_{0}.o` | 1256 ms (32 inputs, parallel) | — |
| `unifiedObjects_{0}.o` | — | **8473 ms (2 inputs, serial)** |
| lowering | 479 ms | 104 ms |

Lowering once is the win, and it is real here too (479 -> 104 ms). But unified also compiles **one object per device instead of one per core**, which serializes what #1863 had just parallelized. On a design whose per-core codegen is expensive, that loss swamps the lowering saving. The LLM ELFs come out ahead only because their lowering is enormous (17.5 s over 192 cores) and their per-core object compile is cheap.

So the premise in the commit message — that per-core lowering "is doing the same work N times and throwing away N-1 of each result" — is true of the *lowering* and not of the *object compile*, and `--unified` does not let you have one without the other.

## Correctness still holds

Nothing here contradicts the correctness evidence: `.text` bit-identical in 192/192 core ELFs, verify PASS 2/2, GEMV 4/4. The problem is purely that the flag is a performance regression on designs the repo supports.

## Also fixed in fbb5e6fa, from the fallback comment

`runAieCompilation` falls back to `aiecc.py`, and the two drivers disagree on the default: the C++ aiecc lowers per-core, while aiecc.py has `aie_unified_compile = True` in configure.py. Unified was in fact the default for *both* backends until the declarative rewrite (mlir-aie #3271) changed it. That makes the original `--no-unified` on the Chess path a behaviour change on the fallback driver rather than the no-op it claimed. Second commit passes `--unified` on the Peano path only and says nothing about Chess.

That also answers the open question in the original description about why unified is opt-in: it is not a cautious default for a new mode, it is the old default that the rewrite flipped.

## Where this should go

The useful change is upstream, not here: split the two behaviours in aiecc so a device can be lowered once while its cores still compile to separate objects in parallel. That is strictly better than either mode today for every design shape. I would rather file that against mlir-aie than land a flag in aircc that trades npu1 and small npu2 designs for large ones.

Leaving the third comment (`segAieccCmd`) alone deliberately — that path passes no `-j` either, so it is consistent with #1863's scope, and I have no Versal hardware to validate a change there.

erwei-xilinx:
Superseded by #1873.

Closing rather than reusing this branch. The change itself is unchanged -- pass `--unified` on the Peano path -- but the first commit here argues for it on a rationale that this PR then went on to refute, and since we do not force-push, that message is permanent. #1873 is one commit that states what is actually true now.

This stays as the record of how the regression was found, which is what motivated Xilinx/mlir-aie#3580: `--unified` bundled "lower once per device" with "compile the device's cores into one object", and the second half narrowed the object stage from per-core parallelism to per-device. Harmless on a 13-device LLM ELF, fatal on a single-device design -- npu1 `run4x4` 4.1 s -> 11.2 s.

#3580 splits the two. With it applied locally, the air-side change measures 85.5 s -> 47.5 s on `llama32_1b_q4nx` `make compile`, with npu1 4x4 at 4.1 s and npu2 4x4 at 3.9 s -- no longer a regression on either. #1873 is draft until #3580 lands and the pin is bumped.

PR #1866, opened 2026-08-21
[python] air.api: affine index arithmetic, buffer lifetime, and worker_to_worker

erwei-xilinx:
Converts `channel_examples/worker_to_worker` onto `air.api`, replacing the raw-bindings version. It needed two things the DSL could not express.

## 1. `%` and `//` on a tile coordinate

Neither is linear, so neither can be a coefficient on a leaf in `IndexExpr`'s affine form. Both now become a `DerivedLeaf`, which the linear form carries like a coordinate and expands into `affine.mod` / `affine.floordiv` only at materialisation — so the whole expression still lowers to one `affine.apply`, and a `DerivedLeaf` takes no symbol of its own.

The maps are the ones the predecessor built by hand with `AffineExpr.get_mod` and `get_floor_div`:

```
                     predecessor (hand-built AffineMap)   air.api (Python operators)
tile_num          →  ()[s0, s1] -> (s0 * 3 + s1)          ()[s0, s1] -> (s0 * 3 + s1)
th_next           →  ((s0 + (s1 + 1) floordiv 3) mod 2)   ((s0 + (s1 + 1) floordiv 3) mod 2)
tw_next           →  ((s0 + 1) mod 3)                     ((s0 + 1) mod 3)
```

A coordinate divisor is still rejected — affine requires a literal — but the message now names the divisor as the problem rather than reporting the operation as unsupported.

**None of this runs on the NPU.** `air-to-aie` clones the herd body per core with the coordinates folded to constants. In the final `npu.air.mlir`:

```
affine.apply anywhere in the module:  0
mod / floordiv in any affine_map:     0
aie.core bodies:                      6
aie.flow (static routes):            18
```

A core body is loads, stores, one `addi` and lock ops; the tile number arrives as `arith.constant 5`. The affine form exists to carry the expression from trace time — where the body is walked once, symbolically — to specialisation.

## 2. Buffer lifetime

Every dealloc used to be emitted at the end of the body. That overstates the lifetime — it tells the compiler a value is still needed when it is not — and the overstatement is not free. Each dealloc now lands after the buffer's own last use, walked out to the ancestor sitting in the block the alloc was emitted into, so a use nested in a loop still holds the buffer until the loop is done.

`air.dealloc(buf)` is added alongside, for a program that wants to name the point rather than let the tracer infer it. This is not an AIE detail: the in-tree ROCDL path (`AIRToROCDLPass.cpp`) has a real allocator behind `memref.dealloc`. Releasing before the last use, releasing twice, and passing a non-buffer all raise; the first is reported when the body finishes, since whether a later use exists is not knowable at the call.

**No new MLIR op is emitted.** `memref.dealloc` was already emitted, just later.

### How the deadlock was isolated

The conversion timed out while the predecessor passed. Six raw-bindings variants of the predecessor, one change each:

| variant | change | npu1 |
|---|---|---|
| predecessor | — | **PASS** |
| v1 | L3 put/get moved inside the segment | PASS |
| v2 | v1 + unused L3 memrefs as herd operands | PASS |
| v3 | v2 + a 1×1 `air.launch` grid | PASS |
| v4 | v2 + identity `affine.apply` on channel indices | PASS |
| v5 | v3 + **all deallocs deferred to the end of the body** | **TIMEOUT** |
| v6 | v3 + `tile_num` computed inside the loop nest | PASS |

Diffing the two final `npu.air.mlir` files — 562 lines each, identical `aie.buffer` (24) and `aie.lock` (48) counts — the only difference is the position of two `aie.use_lock(..., Release, ...)` per core. Deferring the release makes every core acquire its predecessor's tile before releasing its own to its successor: a six-core circular wait.

## Verification, on npu1 (Phoenix)

- `worker_to_worker`: **`PASS!`**. Its predecessor also passes, so the README's *"This example currently fails for unknown reasons"* is stale and is dropped along with the WIP marker.
- `lit -sv build/python/test/ --filter=api` → `Passed : 12 (36.36%)`, including the two new files.
- The 24 previously converted examples: IR differs from baseline **only** in dealloc position, and **no file changed its dealloc count**. Across the `air.dealloc` addition they are byte-identical (24/24).
- npu1 lit sweep over every converted example: **29 pass, 1 fails.** The failure is `matrix_vector_multiplication/bf16/run_npu1_makefile_peano.lit`, at `PROFILE1`, with `xrt/xrt_bo.h: No such file or directory`. Its three correctness runs (`RUN1`/`RUN2`/`RUN3`) pass, and a rebuild **without** this change reproduces the identical failure — the local build was configured without XRT, so lit passes `XILINX_XRT=XRT_DIR-NOTFOUND` to the C++ profiling harness. Not a regression and not related to this change.

Not verified here: the `chess` and `ryzen_ai_npu2` lits, which this box cannot run.

### DIFF
--- a/programming_examples/channel_examples/README.md
+++ b/programming_examples/channel_examples/README.md
@@ -28,11 +28,9 @@ This example ([worker_to_self/worker_to_self.py](worker_to_self/worker_to_self.p
 
 WARNING: This example currently fails for unknown reasons.
 
-#### WIP: ```worker-to-worker```:
+#### ```worker-to-worker```:
 
-This example ([worker_to_worker/worker_to_worker.py](worker_to_worker/worker_to_worker.py)) is a work-in-progress data passthrough example using the same tiling structure as the [matrix_scalar_add/multi_core_channel](../matrix_scalar_add/multi_core_channel.py) examples, only the each worker trades a tile of input data to another worker in the herd by sending it via channel.
-
-WARNING: This example currently fails for unknown reasons.
+This example ([worker_to_worker/worker_to_worker.py](worker_to_worker/worker_to_worker.py)) uses the same tiling structure as the [matrix_scalar_add/multi_core_channel](../matrix_scalar_add/multi_core_channel.py) examples, only each worker trades its tile of data to another worker in the herd by sending it via channel. A worker stamps its own tile number on its tile before handing it on, so the output records the permutation rather than only the copy.
 
 #### ```broadcast```:
 
--- a/programming_examples/channel_examples/worker_to_worker/worker_to_worker.py
+++ b/programming_examples/channel_examples/worker_to_worker/worker_to_worker.py
@@ -1,17 +1,45 @@
 # Copyright (C) 2024, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
+"""A ring of workers passing tiles to their neighbours, on air.api.
+
+    air.channel @ChanIn          L3 -> L1, one bundle member per worker
+    air.channel @WorkerToWorker  L1 -> L1, worker (th, tw) -> its successor
+    air.channel @ChanOut         L1 -> L3, one bundle member per worker
+
+Each worker stamps its own tile number onto its tile and hands the result not
+to itself but to the *next* worker in row-major order, which then writes it
+out. So the tile that lands at (th, tw) in the output carries the tile number
+of the worker *before* it -- the test's expected output is built from exactly
+that permutation, and a kernel that quietly kept its own tile would fail it.
+
+The successor is where the index arithmetic comes in:
+
+    tw_next = (tw + 1) % W
+    th_next = (th + (tw + 1) // W) % H
+
+Both must reach the IR as affine expressions rather than a chain of arith ops,
+because the AIR dependency analysis and the DMA specialisation passes read
+affine offsets directly. The raw-bindings version this replaces built them by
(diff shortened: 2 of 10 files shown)


erwei-xilinx:
Good catch — that was a real hole, not a wording problem. `nxt - nxt` reuses one `DerivedLeaf`, so the cancellation would succeed under pure object identity and the test proved nothing about the structural key it claimed to cover.

Now built twice, deliberately:

```python
# Built twice, deliberately, rather than computed once and
# subtracted from itself: two calls produce two distinct
# DerivedLeaf objects, so they only collapse into one term
# if the key is the expression. Reusing a single value would
# cancel under object identity too and prove nothing.
row = ((tw + 1) % W) - ((tw + 1) % W)
```

Rather than assert it has teeth, I checked. Temporarily removing `DerivedLeaf.__eq__`/`__hash__` so identity falls back to the object:

```
DerivedLeaf.__eq__/__hash__ removed
RESULT: test FAILS with object identity -> the test now detects it
restored
restore verified: install == source
index.py PASSES with the real implementation
```

Under object identity the two terms stay distinct, `row` is no longer a compile-time constant, and the offset materialises as an `affine.apply` of a `mod` map — which trips the `CHECK-NOT: mod` and the static `[0, %...]` offset check. So the test fails for the right reason.

`lit --filter=api` still 13/13. No non-test code changed, so the emitted IR of every example is untouched.

PR #1869, opened 2026-08-21
[python] air.api: implement ops.dot(transpose_b=True)

erwei-xilinx:
`transpose_b=True` says B is stored `[n, k]` rather than `[k, n]`, so `a`'s last axis contracts against b's last. It previously raised:

> not implemented yet; it needs `linalg.matmul_transpose_b`

That was stale. `linalg.matmul` takes `indexing_maps` directly from the Python bindings, so the transpose rides in the maps on a **named** op:

```mlir
#map  = affine_map<(d0, d1, d2) -> (d0, d2)>
#map1 = affine_map<(d0, d1, d2) -> (d1, d2)>     // B as [n, k]
#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
linalg.matmul indexing_maps = [#map, #map1, #map2] ins(...) outs(...)
```

No deprecated op, no `linalg.generic`, no pipeline change.

## Why not generate-then-specialize

The other candidate was emitting a generic and recovering the named op with `--linalg-specialize-generic-ops`. Measured against this LLVM by round-tripping each named op through `--linalg-generalize-named-ops --linalg-specialize-generic-ops`:

| op | specializes back? |
|---|---|
| `linalg.matmul` | yes |
| `linalg.batch_matmul` | yes |
| `linalg.fill`, `linalg.copy` | yes |
| `linalg.dot` | **no** |
| `linalg.vecmat` | **no** |
| `linalg.matvec` | **no** |

Same on tensor semantics, so it is coverage rather than a memref restriction. That route would have discarded three of the four contractions `ops.dot` dispatches to, in order to re-derive by pattern match what the tracer already knows from the operand ranks. The named-op table stays.

## Nothing downstream needs to change

- `air-to-aie` matches **no** named contraction op (zero hits for `MatmulOp`/`VecmatOp`/`MatvecOp`/`DotOp` in `mlir/lib`, `mlir/include`)
- `lower_linalg_to_func` matches the interface — `fnName = linalgOp.getLibraryCallName()`
- `transform.air.herd_vectorize` is maps-agnostic by construction: `MatchAnyOpTypeTag()` + `linalg::hasVectorizationImpl` + `linalg::vectorize`

Checked by A/B against the untransposed matmul: both produce structurally identical `vector.contract`s, the transposed one with `vector<32x16xf32>` for B.

## Verified on npu1 (Phoenix)

Two arms computing the **same** `C = A @ B` against the same reference; the transposed arm is handed `Bᵀ` and must undo it through the map.

```
CONTROL     B stored [K,N], transpose_b=False   PASS!   exit 0
TRANSPOSED  B stored [N,K], transpose_b=True    PASS!   exit 0
```

The test discriminates rather than passing by luck:

```
elements differing between A@B and A@B.T : 1016 / 1024
max abs difference                        : 161
=> a map that ignored the transpose would mismatch
```

And the transpose survives into the core program — B read at `[n, k]`, A at `[m, k]`:

```mlir
scf.for %arg0 ... scf.for %arg1 ... scf.for %arg2 {
  %0 = memref.load %buf2[%arg0, %arg2]   // A[m, k]
  %1 = memref.load %buf1[%arg1, %arg2]   // B[n, k]
  %2 = memref.load %buf0[%arg0, %arg1]   // acc[m, n]
```

## Two limits, stated in the docstring

- **This is the scalar path** — `convert-linalg-to-loops`, 32×32×32, no external kernel and no transform script. The AIE2 matmul intrinsic is **not** exercised; it wants micro-tiled operands whose layout the DMA pack (`mm.b(...)`) already fixes, which is where a transposed B would more naturally be expressed. An earlier probe could not settle this either way — the untransposed control failed `convert-vector-to-aievec` identically, so the shapes were simply wrong for that path.
- **Under `lower_linalg_to_func`**, an external kernel built for `[k, n]` links happily against an `[n, k]` operand and computes silently wrong results, so the transposed form should be given its own symbol with `kernel=`.

There is also **no consumer in-tree** today — grepping for `transpose_b`/`matmul_transpose` finds only nvgpu, a DMA-transpose example, and air.api's own error strings. This is surface, measured rather than assumed.

## Regression

- All 24 air.api examples **byte-identical to main**
- `lit --filter=api` 11/11 — three maps pinned in `dot.py`, two shape rules in `errors.py`

### DIFF
--- a/python/air/api/ops.py
+++ b/python/air/api/ops.py
@@ -587,6 +587,23 @@ def dot(a, b, acc=None, alpha=1.0, transpose_b=False, dependency=None, *, kernel
     One rule covers all four: ``a``'s last axis contracts against ``b``'s first,
     and ``acc`` keeps what is left of each.
 
+    ``transpose_b=True`` says B is stored ``[n, k]`` rather than ``[k, n]``, so
+    ``a``'s last axis contracts against b's last. It is emitted as a named
+    ``linalg.matmul`` carrying the transpose in its indexing maps -- not the
+    deprecated ``linalg.matmul_transpose_b``, and not a ``linalg.generic``.
+    Nothing downstream needs to know: ``air-to-aie`` matches no named
+    contraction, and both ``lower_linalg_to_func`` and
+    ``transform.air.herd_vectorize`` go through the ``LinalgOp`` interface,
+    which is indexing-map agnostic.
+
+    Two limits worth stating. It is verified through the scalar path -- lowered
+    by ``convert-linalg-to-loops`` and run on npu1 -- and *not* through the AIE2
+    matmul intrinsic, which wants micro-tiled operands whose layout the DMA pack
+    (``mm.b(...)``) already fixes. And under ``lower_linalg_to_func`` an external
+    kernel built for ``[k, n]`` links happily against an ``[n, k]`` operand and
+    computes silently wrong results, so pass ``kernel=`` to give the transposed
+    form its own symbol.
+
     ``kernel=`` is keyword-only, and sits after ``dependency`` so that every
     existing positional binding is unchanged: inserting it earlier would have
     silently rebound a positionally-passed ``dependency`` to it. It names the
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Both addressed.

**`AffineDimExpr.get` vs `AffineExpr.get_dim`** — switched, though the stated risk isn't real: `AffineExpr.get_dim` *is* provided by these bindings and returns the identical object, which is why the hardware run passed.

```
AffineExpr.get_dim(0)  -> d0 AffineDimExpr
AffineDimExpr.get(0)   -> d0 AffineDimExpr
equal? True
```

The consistency point stands on its own, and is sharper than the comment suggests. Tree-wide the two spellings are near-even (7 vs 5), but inside `python/air/api/` — the package this edits — there was exactly one existing dim constructor and it uses the other form:

```
python/air/api/_emit.py:124:  ... [AffineDimExpr.get(rank - 1)] ...
python/air/api/ops.py:723:    m, n, k = (AffineExpr.get_dim(i) for i in range(3))   <- the outlier
```

So mine was the only one out of step with its own package. Now `AffineDimExpr.get`.

**`transpose_b=` in the test comment** — now `transpose_b=True`.

Re-verified rather than assumed, since the first change is in the map construction itself:

```
lit --filter=api : 11/11
CONTROL:    PASS!
TRANSPOSED: PASS!
```

PR #1873, opened 2026-08-22
[aircc] Ask aiecc to lower each device's cores once

erwei-xilinx:
[aircc] Ask aiecc to lower each device's cores once

`aiecc --unified` lowers a device's cores together instead of once per core.
Per-core lowering runs the whole pipeline against a clone of the entire design
and `AIECoreToStandardPass` then discards every core but one, so N-1 of every N
results are thrown away. aircc has never asked for it, so every build pays that.

Peano only. The conditional mirrors the `-j` one above it, and asks rather than
states both halves: the two drivers disagree on the default -- the C++ aiecc
lowers per-core, `aiecc.py` has `aie_unified_compile = True` -- so naming
Chess's half would change it on one of them.

## Blocked on Xilinx/mlir-aie#3580 and a pin bump

Until that lands, `--unified` also merges each device's cores into one shared
object, which narrows the object stage from as many parallel compiles as there
are cores to as many as there are devices. That is a large regression on
single-device designs -- npu1 `matrix_multiplication/bf16 run4x4` goes 4.1 s ->
11.2 s -- and is why the first attempt at this change (#1865) was withdrawn.
#3580 splits the two so unified lowers once per device while the cores keep
their own objects.

So this needs, in order: #3580 merged, a published wheel containing it,
`utils/clone-mlir-aie.sh` bumped to that hash, and only then this commit. The
bump is not free: mlir-aie is already 10+ commits past our pin and #3577
("Add out-of-order S2MM DMA receive channels") changes DMA op builders that
`AIRRtToNpuPass.cpp:873` and `AIRToAIEPass.cpp:6478` call, so it carries those
call-site fixes with it.

## Measured

Ryzen AI 9 HX 370 (Strix), 24 CPUs, Peano, against a local mlir-aie carrying
#3580, on top of #1863's `-j 0`.

Clean `make compile`, llama32_1b_q4nx:

| | without | with |
|---|---|---|
| rms_gemms_rope | 17.3 s | 8.6 s |
| o_ffn | 44.0 s | 20.9 s |
| flash_attn | 6.9 s | 5.6 s |
| lm_head_gemv | 11.3 s | 7.7 s |
| total | 85.5 s | **47.5 s** |

Single-device designs through aircc, which is where the old `--unified` hurt:
npu1 4x4 4.1 s, npu2 4x4 3.9 s -- both at or under the per-core figures they
replace.

| check | result |
|---|---|
| `check-air-mlir` | 527 tests, 513 passed, 0 unexpected failures |
| `llama32_1b_q4nx` verify, clean -> compile -> verify | PASS 2/2 |
| GEMV bf16 npu2 configs on hardware | PASS |

Not covered: Chess (no Vitis on this host), which the conditional leaves alone.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Depends on Xilinx/mlir-aie#3580. Draft until that lands and the pin is bumped.


erwei-xilinx:
Unblocked and out of draft. #1912 landed the pin bump (91867f2c), so aiecc now has the unified-lowering split from Xilinx/mlir-aie#3580 and this can ask for the flag.

Re-measured on the real pinned toolchain rather than a local build of #3580 — mlir-aie checked out at exactly `4c8b56f5db2` (the pinned hash), mlir-air main at `91867f2c`, main merged in as `ff1efdb0`. Still the same seven lines.

Clean `make compile`, `llama32_1b_q4nx`, A/B on the same box back to back:

| | without | with |
|---|---|---|
| rms_gemms_rope | 17.1 s | 8.5 s |
| o_ffn | 44.0 s | 20.6 s |
| flash_attn | 6.7 s | 5.2 s |
| lm_head_gemv | 11.0 s | 7.2 s |
| **total** | **84.1 s** | **46.2 s** |

**1.82x**, and it lines up with the 85.5 -> 47.5 s I measured against a local #3580 build before the bump.

The case that killed the first attempt (#1865) is clean now — single-device designs are what the old `--unified` regressed 2.7x, because it also merged the cores into one object:

```
npu1 matmul 4x4 through aircc: 3.9 s
npu2 matmul 4x4 through aircc: 3.9 s
```

Both at or under the per-core figures they replace, not above.

| check | result |
|---|---|
| `check-air-mlir` | 538 tests, 524 passed, 0 unexpected failures |
| `llama32_1b_q4nx` `make compile` | passes |
| `llama32_1b_q4nx` `make verify` (npu2 hardware) | **PASS 2/2** |
| GEMV bf16, 3 npu2 configs (hardware) | PASS |

Unchanged from the original description: Chess is untouched (the conditional leaves it alone), and it is not exercised here — no Vitis on this host.

erwei-xilinx:
Both addressed (412da69d).

## Duplicated condition — taken

`(xchesscc || xbridge)` picked the backend in three places in this function, so `-j` and `--unified` could drift. Bound once and reused:

```cpp
// Which core backend aiecc will use; the same condition selects it above.
const bool useChess = xchesscc || xbridge;
...
aieccCmd.push_back(useChess ? "1" : "0");
...
if (!useChess)
  aieccCmd.push_back("--unified");
```

No behaviour change — same condition, same two flags.

## Pin bump — already here

The suppressed comment asks for the pin bump or a compatibility gate, on the grounds that this PR doesn't include it. It does, via the merge of main:

```
$ grep '^export HASH' utils/clone-mlir-aie.sh
export HASH=4c8b56f5db25a9a89582bffed3ad933dea5b1d15
```

`4c8b56f5db2` is the merge commit of mlir-aie#3580 itself, so the pinned wheel is the first one that carries the lowering/object split. The comment quotes that same hash as evidence the bump is missing — it is actually evidence it is present. #1912 landed it as `91867f2c` and this branch merged main afterwards.

On the compatibility-gate alternative: not needed, and I would rather not. `utils/clone-mlir-aie.sh` pins an exact hash, so there is no "older pin" a build can pick up — the pin and this flag move together in the same tree. A version probe would add a failure mode (parsing an aiecc version) to guard a state the repo cannot be in.

## Re-verified after the refactor

Measured on the real pinned toolchain: mlir-aie checked out at `4c8b56f5db2`, mlir-air main at `91867f2c`.

| check | result |
|---|---|
| `check-air-mlir` | 538 tests, 524 passed, 0 unexpected failures |
| `llama32_1b_q4nx` `make compile` | 45.0 s (84.1 s without the flag) |
| `llama32_1b_q4nx` `make verify` (npu2 hardware) | PASS 2/2 |

Single-device designs, the case that killed the first attempt (#1865), stay clean: npu1 4x4 3.9 s, npu2 4x4 3.9 s through aircc.

PR #1875, opened 2026-08-23
[programming_examples] air.api: the five plain-DMA examples

erwei-xilinx:
Converts the DMA counterparts of examples already on `air.api`, replacing the raw-bindings versions:

| example | what it shows |
|---|---|
| `shim_dma_2d` | a strided 2-D shim transfer |
| `passthrough/passthrough_dma` | chunked passthrough, 6 element types |
| `segment_alloc` | L2 staging via `seg.private()` |
| `matrix_scalar_add/single_core_dma` | one core walking every tile |
| `matrix_scalar_add/multi_core_dma` | one core per tile |

**No DSL code changes** — these are example conversions only, so the emitted IR of the already-converted examples is untouched.

Three of the five are siblings of converted channel examples, and the pairs are worth reading together: a channel is a stream, so one put can feed many gets and neither end names an offset, while a transfer is addressed and the loop variable supplies the offset explicitly. `multi_core_dma` is where `air.api` pays off most — turning a herd coordinate into an L3 offset needed four hand-built `AffineMap`s and an `affine_apply` per use, because a coordinate is an SSA value; here `tx` is an index expression and `tx * tile_height` is just a slice bound.

`passthrough_dma` keeps all six element types, including `uint8` and `uint16`, now declared as such rather than as their signed siblings. An unsigned tile is movable but not computable in `air.api`, and this kernel only moves and copies one; the emitter keeps those two configs on the scalar path, since a vector copy needs a padding value and that is an `arith.constant`.

### The one behavioural trap worth flagging in review

`multi_core_dma` pins `shape=(rows, cols)` on its herd. `air.api`'s default 2-D herd shape on npu1 is `(1, 4)` — deliberately conservative, since a 2-D herd can exhaust shim DMA capacity before it exhausts columns — and letting it apply here strip-mines the row axis into a temporal `scf.for`, giving **2 cores doing 2 tiles each where the predecessor had 4 doing one apiece**. The output is still correct, so this is only visible by diffing the emitted herd sizes against the predecessor. One core per tile is the point of this example against its `single_core` sibling, hence the explicit shape.

`shim_dma_2d`'s `build_module` now returns the launch rather than a module, matching the other conversions; `run.py` calls `.build(target=...)`. Its module-level shape constants stay, because `run.py` and `test.py` both `import *` from it.

### Verification

17 lits across the five. **6 are runnable on the npu1 box used here, and all 6 pass:**

```
-- Testing: 17 of 383 tests, 16 workers --
Total Discovered Tests: 383
  Excluded   : 366 (95.56%)
  Unsupported:  11 (2.87%)
  Passed     :   6 (1.57%)
```

The 11 unsupported are not silently skipped: 6 need a chess licence (not installed here) and 5 are `ryzen_ai_npu2`. Both groups were covered by compile sweeps instead — all five examples compile for `npu2_4col`, and the four with `OUTPUT_FORMAT=elf` lits also compile via `--output-format elf`.

`passthrough_dma` advertises 6 dtypes while its lits exercise only 2, so all six were compile-swept on npu1:

```
uint8: COMPILE OK (8358 bytes)
int8: COMPILE OK (8486 bytes)
int16: COMPILE OK (8342 bytes)
uint16: COMPILE OK (8342 bytes)
float32: COMPILE OK (8342 bytes)
bfloat16: COMPILE OK (8342 bytes)
```

The four signless types now vectorise where the predecessor ran scalar; `uint8`/`uint16` stay scalar, matching the predecessor exactly.

Not measured: no perf A/B was run on npu2 — the box used here is npu1.

### DIFF
--- a/programming_examples/matrix_scalar_add/multi_core_dma/multi_core_dma.py
+++ b/programming_examples/matrix_scalar_add/multi_core_dma/multi_core_dma.py
@@ -1,146 +1,100 @@
 # Copyright (C) 2024, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
+"""Add each tile's index to that tile, one core per tile, on air.api.
+
+The sibling ``single_core_dma`` walks every tile from one core with a loop nest.
+This one hands each tile to its own core: the loop *is* the herd, and the loop
+variables are herd coordinates rather than induction variables.
+
+    with air.herd([range(rows), range(cols)], name="xaddherd") as h:
+        def _(tx, ty):
+            air.ops.load(tile_in, A[tx*th : ..., ty*tw : ...])
+            tile_out[:] = tile_in[:] + (tx * cols + ty)
+            air.ops.store(tile_out, B[tx*th : ..., ty*tw : ...])
+
+That is the whole difference from the sibling, and it is the one air.api makes
+cheapest. In the raw-bindings version this replaces, turning a herd coordinate
+into an offset meant building an ``AffineMap`` per expression -- four of them,
+each a handful of ``AffineExpr.get_mul``/``get_add`` calls, plus an
+``affine_apply`` per use -- because a coordinate is an SSA value and cannot be
+multiplied with Python's ``*``. Here it can: ``tx`` is an index expression, so
+``tx * tile_height`` is a slice bound and ``tx * cols + ty`` broadcasts into the
+elementwise expression as an ``index_cast``. Roughly sixty lines of map
+construction become three.
+
+Unchanged otherwise, except that the per-tile add is written
+``tile_out[:] = tile_in[:] + n`` rather than a scalar loop nest over every
+(i, j). ``vector=0`` keeps the emitted loop scalar as the predecessor's was,
+whatever ``--tile-width`` is given: a <8 x i32> add is 256-bit and does not
+legalize on AIE2.
+"""
+
 import argparse
+import numpy as np
 
-from air.ir import *
-from air.dialects.air import *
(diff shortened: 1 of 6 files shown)


erwei-xilinx:
Addressed in `54a2431b`.

**On the stale `type_mapper` reference — agreed, that was a real defect.** The conversion dropped the `type_mapper` import, so the comment named machinery that is no longer visible anywhere in the file. Reworded to state what a reader of *this* file can actually check: `air.api` maps `np.uint8` onto the unsigned `ui8` rather than the signless `i8`, so declaring `i8` would emit an argument type that disagrees with the `np.uint8` array handed to `XRTRunner` a few lines below. That is also the more concrete framing you suggested.

**On "signful" — I don't think that one was wrong,** but I've dropped the word anyway. *Signless* / *signed* / *unsigned* are MLIR's three integer signedness semantics, and *signful* is the standard complement of *signless*; it is the term [`python/air/api/types.py`](https://github.com/Xilinx/mlir-air/blob/main/python/air/api/types.py) already uses for exactly this ("maps numpy's unsigned dtypes onto MLIR's signful `ui8`/`ui16`/`ui32`"). So it was accurate and consistent with the DSL it documents. But it is jargon that wasn't earning its place in an example, and "unsigned `ui8`, not the signless `i8`" says the same thing without requiring the reader to know the term — so the change is worth making on clarity grounds, not correctness ones.

Comment-only change, verified rather than assumed: the emitted IR is byte-identical for `uint8` and `bfloat16` against the pre-fix output, and both npu1 `passthrough_dma` lits still pass (2 passed, 2 unsupported — the chess and npu2-elf ones).

PR #1876, opened 2026-08-23
[programming_examples] air.api: four vector_examples primitives

erwei-xilinx:
Converts `vector_add`, `vector_sub`, `vector_mul` and `vector_max` onto `air.api`, replacing the raw-bindings versions. Each is now one line of compute:

```python
c[:] = a[:] + b[:]                  # vector_add
c[:] = a[:] - b[:]                  # vector_sub
c[:] = a[:] * b[:]                  # vector_mul
c[:] = air.ops.maximum(a[:], b[:])  # vector_max
```

**No DSL code changes** — example conversions only. 618 lines deleted, 293 added.

### Why these four are the clearest case yet for the DSL

Each predecessor spelled its vectorised loop by hand: an `scf.for` over the tile in steps of `VECTOR_SIZE`, three `memref.subview`s per trip to carve out the current lane group, three `vector.transfer_read`s with an explicit identity permutation map and a padding constant, the `arith` op, and a `transfer_write`. Roughly forty lines in which exactly one token said what the kernel computes. The emitter builds that loop — and the subviews are not needed at all, because `air.api` reads at an offset directly.

The strip-mine is the DSL's now too. Each predecessor asked for a `[1, NUM_TILES]` herd and then wrote the outer loop itself, computing the L3 offset through a hand-built `AffineMap`. Here the herd's iteration space is the whole tile grid and `air.api` strip-mines it onto `NUM_TILES` cores.

### Behavioural differences, stated rather than buried

- **Tile distribution changes from interleaved to blocked.** The predecessor's core `ty` took tiles `ty, ty+2, ty+4, …`; each core now takes a contiguous run. Coverage is identical — every 1024-element tile of the 65536-element vector, exactly once — and the tiles are independent, so the result is unchanged.
- **The herd becomes `[NUM_TILES, 1]` rather than `[1, NUM_TILES]`.** A 1-D `air.api` herd is laid out along x, which is the orientation that places on both generations.
- `vector_mul` keeps its `--arch` flag. It selects the vector width (`aie2` → 16, `aie2p` → 64) and, despite the name, does **not** pin the device: `--target` stays `auto`, so the module still runs on whichever generation is installed.

### Verification

19 lits across the four. **8 are runnable on the npu1 box used here, and all 8 pass** — including both `--bf16-emulation` configs, which build the f32 variant rather than bf16:

```
-- Testing: 19 of 383 tests, 16 workers --
  Unsupported:  11 (2.87%)
  Passed     :   8 (2.09%)
```

The 11 unsupported are all `ryzen_ai_npu2`. They are covered by compile sweeps instead: all four examples compile for `npu2_4col` in **both** `xclbin` and `elf` form, at the exact vector widths those lits ask for — `vector<32xbf16>` for add/sub/max (`VECTOR_SIZE=32`) and `vector<64xbf16>` for mul (`AIE_TARGET=aie2p`).

Not measured: no perf A/B — the box used here is npu1, and these examples have no `profile` target or `--perf-iters`.

### Note for whoever picks up the rest of this family

The remaining `vector_examples` are deliberately not in this PR. `vector_div` and `vector_reciprocal` are *entirely* npu2-gated (zero lits runnable on npu1) and bf16 vector divide is a known non-legalizing case, so they want an npu2 box rather than a compile-only claim from here. `vector_fma` needs an fma surface, `vector_muladd` is the chained mul→add f32 legalization trap, `vector_broadcast_scalar` needs `collapse_shape`, and the rest need `exp`, `rsqrt`, `select`, bitwise ops, int↔float casts or reductions.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_add/vector_add.py
+++ b/programming_examples/primitives/vector_examples/vector_add/vector_add.py
@@ -1,167 +1,89 @@
 # Copyright (C) 2025, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
-from ml_dtypes import bfloat16
+"""Vector add primitive, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, load, store, subview
-from air.dialects.vector import transfer_read, transfer_write
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+    c[:] = a[:] + b[:]
+
+One line of compute, over a [n] vector cut into [tile_n] tiles and spread across
+``NUM_TILES`` cores. It is the smallest member of the vector_examples family,
+and the family is a showcase: each sibling differs from this one only in the
+operator on that line.
+
+That is the whole argument for writing these on air.api. The raw-bindings
+version this replaces spelled the vectorised loop by hand -- an ``scf.for`` over
+the tile in steps of VECTOR_SIZE, three ``memref.subview``s per trip to carve
+out the current lane group, three ``vector.transfer_read``s with an explicit
+identity permutation map and a padding constant, the ``arith`` op, and a
+``transfer_write`` -- roughly forty lines in which exactly one token, ``AddFOp``,
+said what the kernel computes. Here the emitter builds that loop, and the
+subviews are not needed at all: air.api reads at an offset directly.
+
+The strip-mine is also the DSL's. The predecessor asked for a [1, NUM_TILES]
+herd and then wrote the outer loop itself, computing ``_l_ivx + _ty * tile_n``
+through a hand-built AffineMap. Here the herd's iteration space is the whole
+tile grid and air.api strip-mines it onto ``NUM_TILES`` cores, so a core that
+gets more than one tile loops over them without the kernel saying so.
+
+Two differences from the predecessor worth naming:
+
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Both addressed in `7676ec48`.

### `--target` not reaching the backend — correct, and a real defect

`--target` reached `launch.build(target=...)`, which resolves the generation used while *tracing*, but `XRTRunner`/`XRTBackend` were left to auto-detect independently. Where the two disagree — `--target npu2` on an npu1 box — that traces for one generation and compiles for the other, which is the failure mode that loads without an XRT error and computes nothing.

I've fixed it slightly differently from the suggestion, and I think the difference matters. Rather than forwarding `args.target` and passing `None` when it is `"auto"`, both call sites now take **`target_device=launch.target`** — the *resolved* generation, after `build()` has already turned `"auto"` into something concrete:

```
--target auto  -> launch.target = 'npu1'
--target npu1  -> launch.target = 'npu1'
--target npu2  -> launch.target = 'npu2'
```

With `None`-on-auto, tracing resolves `auto` and the backend then auto-detects *again*; they agree in practice but nothing enforces it. Passing the resolved value makes divergence impossible by construction, and it's the idiom the elementwise conversions already use — `eltwise_add`, `relu`, `gelu`, `sigmoid`, `silu`, `leaky_relu` and `axpy` all pass `target_device=launch.target`. So this also brings these four in line with their closest siblings rather than inventing a third convention.

### `vector_max` copyright year — correct, my error

`vector_max` is a 2026 file, as its Makefile still says. I generated it from `vector_add` and carried that file's 2025 header across with it. Restored to 2026; the other three are genuinely 2025 and are unchanged.

All 8 npu1-runnable lits still pass after the change (11 unsupported are `ryzen_ai_npu2`, as before).

Unrelated heads-up: the red **Build and Test (Windows)** check on this PR is not from these changes — `main` fails it identically. nanobind 3.0.0 shipped and this workflow installs it unpinned while MLIR requires ~2.9. Fix filed separately as #1877.

PR #1878, opened 2026-08-23
[programming_examples] air.api: eltwise_add_with_l2

erwei-xilinx:
Converts the L2-staged elementwise add onto `air.api`, replacing the raw-bindings version. **No DSL code changes** — 154 lines deleted, 96 added.

The sibling `eltwise_add` (already converted) streams tiles from L3 straight into the cores. This one lands the whole vector in a memtile first and the cores read their tiles back out of it:

```python
l2_a = air.alloc([n], f32, scope=seg.private())   # a memtile buffer
  air.ops.load(l2_a, A)                           # L3 -> L2, whole vector
  ...
  air.ops.load(a, l2_a[i0 : i0 + tile_n])         # L2 -> L1, one tile
  c[:] = a[:] + b[:]
  air.ops.store(c, l2_c[i0 : i0 + tile_n])        # L1 -> L2
air.ops.store(l2_c, C)                            # L2 -> L3, whole vector
```

`seg.private()` is what makes the staging buffers memtile-resident, and the herd nested in the segment body carries them in automatically — `air.herd` is `IsolatedFromAbove`, so the predecessor had to list them in `operands=[...]` by hand and pick them back up as body parameters.

### Behavioural differences, stated rather than buried

**The compute is now vectorised.** The predecessor ran a scalar `load`/`addf`/`store` loop over every element. `c[:] = a[:] + b[:]` on a 1024-f32 tile emits `<16 x f32>` (512-bit) — the width f32 has to use, since 8 lanes does not legalize on either generation.

**The herd takes 6 operands where the predecessor took 3.** `air.api` threads every live L3 tensor into the herd whether the body references it or not, and this body only touches the three L2 buffers:

```
old: air.herd @herd_0 ... args(%arg10=%alloc, %arg11=%alloc_0, %arg12=%alloc_1)      # 3
new: air.herd @herd_0 ... args(%arg17=%arg10, ..., %arg20=%alloc, ...)               # 6
```

Unused herd operands are not free — they survive `air-dependency` as async edges. This is inherent to the DSL rather than specific to this example (every converted example has it), but it is the most likely place for a perf delta here, so I am flagging it for measurement rather than assuming parity.

**The herd becomes `[NUM_TILES, 1]`** rather than `[1, NUM_TILES]`, with the tile grid strip-mined onto those cores by the DSL rather than by a hand-written outer loop with an `AffineMap` for the offset. A 1-D `air.api` herd is laid out along x, the orientation that places on both generations.

### Verification

3 lits. The 1 runnable on the npu1 box used here passes; the other 2 need a chess licence or npu2. Both npu2 configurations compile:

```
npu2 xclbin: OK
npu2 elf: OK
```

Not measured: no perf A/B — this box is npu1, and the example has no `profile` target or `--perf-iters`. Given the herd-operand change above, this one is worth an A/B on npu2 before it is assumed neutral.

### DIFF
--- a/programming_examples/eltwise_add_with_l2/eltwise_add.py
+++ b/programming_examples/eltwise_add_with_l2/eltwise_add.py
@@ -38,7 +38,7 @@ import argparse
 import numpy as np
 
 from air import api as air
-from air.api import f32
+from air.api.types import dtype_of
 from air.backend.xrt import XRTBackend
 from air.backend.xrt_runner import XRTRunner
 
@@ -49,7 +49,17 @@ NUM_TILES = 2
 
 def build_module(n, tile_n, np_dtype_in):
     assert n % (tile_n * NUM_TILES) == 0
-    dt = f32
+    # Honoured, not ignored: the host buffers in __main__ are allocated with
+    # np_dtype_in, so hardcoding an element type here would hand f32 memrefs to
+    # non-f32 host arrays -- an interface the caller cannot see is wrong. The
+    # predecessor mapped it through type_mapper; dtype_of is the same table,
+    # reached through the API's own DType so the vector width comes with it.
+    dt = dtype_of(np_dtype_in)
+    if dt is None:
+        raise ValueError(
+            f"unsupported element type {np_dtype_in!r}; air.api knows "
+            f"float32, float16, bfloat16, int8/16/32 and uint8/16/32"
+        )
 
     A = air.tensor([n], dt)
     B = air.tensor([n], dt)
--- a/programming_examples/eltwise_add_with_l2/eltwise_add.py
+++ b/programming_examples/eltwise_add_with_l2/eltwise_add.py
@@ -1,168 +1,100 @@
 # Copyright (C) 2025, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
-from ml_dtypes import bfloat16
+"""Element-wise add staged through L2, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, load, store
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+The sibling ``eltwise_add`` streams tiles from L3 straight into the cores. This
+one lands the whole vector in a memtile first, and the cores read their tiles
+from there:
+
+    l2_a = air.alloc([n], f32, scope=seg.private())   a memtile buffer
+      air.ops.load(l2_a, A)                           L3 -> L2, whole vector
+      ...
+      air.ops.load(a, l2_a[i0 : i0 + tile_n])         L2 -> L1, one tile
+      c[:] = a[:] + b[:]
+      air.ops.store(c, l2_c[i0 : i0 + tile_n])        L1 -> L2
+    air.ops.store(l2_c, C)                            L2 -> L3, whole vector
+
+``seg.private()`` is what makes the staging buffers memtile-resident, and the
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Both addressed — one accepted as a fix, one answered with a measurement.

## `np_dtype_in` ignored — accepted, fixed in f49db6c6

Correct, and worse than cosmetic: the predecessor *did* honour it (via
`type_mapper`), so hardcoding `f32` was a silent regression in the builder's
contract. Fixed by mapping through `dtype_of` — the same table, reached through
the API's own `DType`, so the default vector width comes with it rather than
being pinned separately. An element type `air.api` does not know is now rejected
by name instead of being silently replaced:

```
   float32 -> L3 memref element type: f32
  bfloat16 -> L3 memref element type: bf16
     int32 -> L3 memref element type: i32
np.complex64 -> unsupported element type <class 'numpy.complex64'>; air.api knows
                float32, float16, bfloat16, int8/16/32 and uint8/16/32
```

I kept the parameter rather than removing it, per the second of your two
suggestions being the weaker one here: no caller outside the example imports
`build_module` (checked both the symbol-import and path-reference scans — only
`generate_readme.py` and docs name the directory), so honouring it costs nothing
and keeps the predecessor's interface intact.

## Unused L3 operands threaded into the herd — measured, and they cost nothing

Your premise is exactly right at the AIR level. The herd goes from 3 operands to
6, and the three additions are the L3 tensors the body never mentions:

```
pred: air.herd @herd_0 tile (...) args(%arg17=%alloc, %arg18=%alloc_1, %arg19=%alloc_2)
        : memref<16384xf32, 1 : i32>, ×3

conv: air.herd @herd_0 tile (...) args(%arg17=%arg10, %arg18=%arg11, %arg19=%arg12,
                                       %arg20=%alloc, %arg21=%alloc_1, %arg22=%alloc_2)
        : memref<16384xf32>, ×3, memref<16384xf32, 1 : i32>, ×3
```

The consequence you predicted does not follow, though — they are dead before AIE
IR. Compiling both arms and counting resources in `npu.air.mlir`:

| resource | predecessor | conversion |
|---|---|---|
| `aie.buffer` | 15 | 15 |
| `aie.lock` | 18 | 18 |
| `aie.flow` | 9 | 9 |
| `aie.shim_dma_allocation` | 3 | 3 |
| `aie.memtile_dma` | 3 | 3 |
| `aie.mem` | 5 | 5 |
| `air.insts.bin` | **420 bytes** | **420 bytes** |
| `aie.tile` | 6 | 8 |

Identical everywhere except tile count, and that difference is **not** the
operands — it is the `[1,N]` → `[N,1]` herd orientation the PR already
describes. The predecessor's `[1,2]` herd is one column (cores at `(0,2)`,
`(0,3)`); the conversion's `[2,1]` spans two (cores at `(0,2)`, `(1,2)`), so it
picks up the extra columns' shim tiles. Buffers, locks, flows and DMA
allocations are unchanged, which is the direct answer to "avoidable
scheduling/metadata overhead": there is none reaching the device.

### And the A/B, since you asked for one

npu1, not npu2 — this box has no npu2, and the npu2-gated lit is `peano_elf`
only. Both arms compiled once, then 12 **alternating** reps of 50 iterations
each (not all-of-A-then-all-of-B), same host buffers, correctness asserted on
every rep:

```
 arm_pred: mean    529.0  min    316.5  max   1077.0  std  272.6
 arm_conv: mean    429.7  min    238.4  max   1033.9  std  263.1

conversion vs predecessor: mean -18.8%, min-to-min -24.7%
```

The conversion is faster, and I would not attribute that to the operands either
way — the predecessor's inner loop moves one element per iteration while this
one vectorises to `<16 x f32>`, which is the difference that should dominate.

Two honest caveats on those numbers: `xrt-smi configure --pmode turbo` fails
here with `DRM_IOCTL_AMDXDNA_SET_STATE ... Permission denied`, so this is
Default power mode; and the standard deviations (~270 µs in both arms) are large
because the box is shared, which is why min-to-min is the statistic worth
reading. The resource table above is the load-bearing evidence for your
question; the latency is context.

PR #1879, opened 2026-08-23
[programming_examples] air.api: sine_cosine and softmax via air.extern

erwei-xilinx:
Converts the two extern-kernel examples that have npu1 lit coverage, replacing
the raw-bindings versions. **No DSL code changes.**

Neither computes anything in the DSL: the kernel body is one call into a
hand-written object file. `air.extern` emits the private `func.func`
declaration, stamps `link_with` on both it and the herd, and builds the
`func.call` — which the predecessors spelled out as a `FuncOp` per symbol, two
attribute assignments, and a `CallOp`. The emitted declaration and call are
byte-identical to the predecessors', including softmax's operand order.

Between them they extend proven `air.extern` usage in two directions:

* **`sine_cosine` declares one of *two* symbols from a single object file**,
  chosen by `--mode` at trace time, so the unused one is never emitted. The
  predecessor declared both unconditionally, leaving a dead declaration in every
  build. A herd links against one object; two kernels from the same object are
  fine.
* **`softmax`'s C symbol is `(memref, i32, memref)`** — the length argument sits
  *between* the two buffers. `air.extern` walks the call's arguments in order
  and takes the next entry from `scalars=` at each non-buffer position, so
  `scalars=` says what type the non-buffer argument has, not where it goes.
  Previously only scalar-last was exercised.

Both herds become `[herd_n, 1]` rather than `[1, herd_n]`; a 1-D `air.api` herd
is laid out along x, the orientation that places on both generations. The tile
grid is strip-mined onto those cores by the DSL rather than by a hand-written
outer loop with an `AffineMap` for the offset.

## The `sine_cosine` seed, measured rather than assumed

`sine_cosine` regains `np.random.seed(42)`, which I dropped while converting.
The predecessor carries it too, and it is load-bearing there: inputs come from
`np.random.randn` and the check is relative-only (`rtol=1e0, atol=1e-8`), so any
draw containing bf16's nearest value to −π/2 — where `cos` is 0.000484 — fails
whatever the implementation.

A seed can hide a real numerical difference by pinning the test to one input
set, so this was checked rather than argued. Both versions were compiled once
and run over the **same** 20 draws (seeds 0..19) at the lit's config
(`--mode cos --n 192 --herd-n 4`), comparing **raw device output**, not just
pass/fail:

```
seed  bitwise A==B  A pass  B pass  A bad  B bad   max|A-B|
   9          True   False   False      1      1          0
  15          True   False   False      1      1          0
  16          True   False   False      1      1          0
   (the other 17 seeds: True / True / True / 0 / 0 / 0)

bitwise identical: 20/20
arm A (predecessor) passed: 17/20
arm B (conversion)  passed: 17/20
```

The same three seeds fail in both arms, one bad element each, `max|A-B| = 0` on
every seed: the two designs produce **byte-identical device output** despite
assigning tiles to cores differently (predecessor herd `(1,4)` with
`scf.for 0 to 192 step 96`, offset `arg8 + arg3*24`; conversion herd `(4,1)`
with `scf.for 0 to 2 step 1`, offset `arg2*48 + arg8*24` — same coverage of all
8 tiles, different assignment).

Predicting failure from `min|cos(x)| < 1e-3` alone, with no hardware involved,
reproduces the observed set exactly:

```
predicted set: [9, 15, 16]  observed set: [9, 15, 16]  exact match: True
```

So the seed restores the predecessor's behaviour rather than masking a
difference in the conversion, and the ~15% flake rate it suppresses is
inherited. A small `atol` alongside `rtol=1e0` would make the test robust rather
than merely pinned, but that is a pre-existing weakness and is left alone here.

## Verification

7 lits. The 2 runnable on this box pass **from a wiped output directory**:

```
$ rm -rf build/test/sine_cosine build/test/softmax
$ lit -sv build/programming_examples --filter 'sine_cosine|softmax'
Testing Time: 2.45s
  Unsupported:   5 (1.31%)
  Passed     :   2 (0.52%)
```

The other 5 need a chess licence (2) or npu2 (3). `softmax` compiles for npu2 in
both formats the npu2 lits ask for, against a real `aie2p` kernel object:

```
=== npu2 / xclbin ===  -rw-rw-r-- 34234 air.xclbin  OK
=== npu2 / elf ===     -rw-rw-r-- 91072 air.elf     OK
```

`sine_cosine` has no npu2 path to check: its Makefile has no `AIE_TARGET`
support and its lits are npu1-only.

`black --check`: both files unchanged.

**No perf A/B has been run.** Neither example has a `profile` target or a
`test.cpp` harness, and the npu2 arm is unavailable on this box.

### Two pre-existing warts noticed, deliberately not fixed here

* These lits run `make clean` **without** `PEANO_INSTALL_DIR`, so the Makefile
  selects `BUILD_DIR := build_chess` and cleans the wrong directory —
  `build_peano` is never cleaned between runs. Affects every peano lit with this
  Makefile shape and predates this branch; it is why the run above wipes the
  output dirs explicitly.
* `softmax`'s `make print` target passes `--arch $(AIE_TARGET)`, which neither
  the predecessor nor this conversion accepts. Dormant — no lit uses `print`.

### DIFF
--- a/programming_examples/sine_cosine/sine_cosine.py
+++ b/programming_examples/sine_cosine/sine_cosine.py
@@ -2,11 +2,13 @@
 # SPDX-License-Identifier: MIT
 """Elementwise sine or cosine via a hand-written AIE kernel, on air.api.
 
-    sinf = air.extern("sinf_bf16_24_8", object="sine_cosine.o")
-    cosf = air.extern("cosf_bf16_24_8", object="sine_cosine.o")
+    kernel = air.extern(
+        "sinf_bf16_24_8" if mode == "sin" else "cosf_bf16_24_8",
+        object="sine_cosine.o",
+    )
     ...
     air.ops.load(a, A[i0 : i0 + tile_n])
-    (sinf if mode == "sin" else cosf)(a, out)
+    kernel(a, out)
     air.ops.store(out, C[i0 : i0 + tile_n])
 
 Nothing is computed in the DSL here -- the whole kernel body is one call into
@@ -20,6 +22,13 @@ Both symbols live in one object file, and only one of them is called per build
 -- ``--mode`` picks at trace time, so the other is never declared. A herd links
 against a single object, which is the constraint ``air.extern`` enforces; two
 kernels from the *same* object are fine.
+
+Note how much of this example's interface the object file pins.
+``sinf_bf16_24_8`` instantiates ``sinf_cosf_poly_bf16<24, 8, true>``: both the
+element type and the 24-element extent are compiled into the object rather than
+read from the memref it is handed, so a kernel called with any other tile size
+would read and write 24 elements regardless. ``--tile-n`` and the dtype are
+therefore checked below rather than accepted and silently miscomputed.
 
 Unchanged from the predecessor except that the herd is [herd_n, 1] rather than
 [1, herd_n] -- a 1-D air.api herd is laid out along x, the orientation that
@@ -45,8 +54,30 @@ from air.backend.xrt_runner import XRTRunner
 np.random.seed(42)
 
 
+KERNEL_TILE_N = 24  # the 24 in sinf_cosf_poly_bf16<24, 8, ...>
+
+
 def build_module(n, tile_n, herd_n, mode, np_dtype_in):
     assert n % (tile_n * herd_n) == 0
+    # Checked, not assumed. The object file exports exactly two symbols, and
+    # each one bakes in the element type and the extent -- neither is read from
+    # the memref at the call. Getting either wrong links and runs, and computes
+    # the wrong answer, so refuse here where the cause can still be named.
+    if mode not in ("sin", "cos"):
+        raise ValueError(f"mode must be 'sin' or 'cos', got {mode!r}")
+    if np_dtype_in is not bfloat16:
+        raise ValueError(
+            f"sine_cosine.o exports only bf16 kernels (sinf_bf16_24_8 / "
+            f"cosf_bf16_24_8), so np_dtype_in must be ml_dtypes.bfloat16, "
+            f"got {np_dtype_in!r}"
+        )
+    if tile_n != KERNEL_TILE_N:
+        raise ValueError(
+            f"tile_n must be {KERNEL_TILE_N}: the extent is a template "
+            f"argument of sinf_cosf_poly_bf16<{KERNEL_TILE_N}, 8, ...>, "
+            f"compiled into sine_cosine.o, not taken from the memref. "
+            f"got tile_n={tile_n}"
+        )
     dt = bf16
 
     A = air.tensor([n], dt)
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
All four addressed in ab880235 — three accepted as written, two fixed more
sharply than suggested, plus one more instance of the same defect found while
doing it.

## `mode` ternary silently falls through to cosine — accepted

Right, and it is a regression rather than a new wart: the predecessor ended its
`if/elif` with `else: raise AssertionError`, which I dropped when collapsing it
to a ternary. Restored as a `ValueError` that names the value:

```
mode='tan': mode must be 'sin' or 'cos', got 'tan'
```

## Docstring shows both externs declared — accepted

Also correct, and the misleading part is exactly the bit the PR is about. The
docstring was describing the *predecessor's* shape while the body does the new
thing. It now shows the single trace-time declaration the code actually emits:

```python
kernel = air.extern(
    "sinf_bf16_24_8" if mode == "sin" else "cosf_bf16_24_8",
    object="sine_cosine.o",
)
```

## `np_dtype_in` unused — accepted, but validated rather than renamed

I went with the stronger fix. `_np_dtype_in` documents that the parameter does
nothing; it does not stop a caller passing `np.float32` and getting a bf16
kernel. And bf16 is not an arbitrary internal choice here — it is in the symbol
names the object file exports (`sinf_bf16_24_8`, `cosf_bf16_24_8`,
`softmax_bf16`), so the constraint is real and nameable:

```
dtype=np.float32: sine_cosine.o exports only bf16 kernels (sinf_bf16_24_8 /
                  cosf_bf16_24_8), so np_dtype_in must be ml_dtypes.bfloat16
dtype=np.float32: softmax.o exports only a bf16 kernel (softmax_bf16), so
                  np_dtype_in must be ml_dtypes.bfloat16
```

## `--tile-n` is pinned too — found while fixing the above

Same defect class, not in the review, and sharper than the dtype one. Both
kernels compile the *extent* in as a template argument rather than reading it
from the memref they are handed:

```cpp
// sine_cosine.cc
void sinf_bf16_24_8(const bfloat16 *in, bfloat16 *out) {
  sinf_cosf_poly_bf16<24, 8, true>(in, out);   // 24 is the extent
}

// softmax.cc
void softmax_bf16(const bfloat16 *in, const int pos, bfloat16 *out) {
  zero_vectorized<bfloat16, 1, 256, 16>(out);  // clears 256, whatever out is
  ...
```

So `--tile-n 32` on sine_cosine links, runs, and processes 24 elements of a
32-element tile. On softmax the failure is worse than partial: `zero_vectorized`
writes 256 elements unconditionally, so a smaller tile is written past its end
and a larger one is left partly dirty. Now rejected by name, with the `--tile-n`
help saying why:

```
tile_n=32:  tile_n must be 24: the extent is a template argument of
            sinf_cosf_poly_bf16<24, 8, ...>, compiled into sine_cosine.o,
            not taken from the memref
tile_n=512: tile_n must be 256: the extent is a template argument of the
            zero_vectorized call inside softmax_bf16, compiled into softmax.o
```

Worth saying plainly: **this trap is inherited, not introduced.** The
predecessors built a `[tile_n]` memref of `type_mapper(np_dtype_in)` and handed
it straight to a `_bf16_24_` symbol, so both flags were equally silent there.
The conversion is where it became checkable, so it is checked here.

## Verification after the change

Nothing in the lits moves — both Makefiles already run at the pinned values
(`sine_cosine` tile_n=24, `softmax` tile_n=256, neither passes `--tile-n`). From
wiped output directories:

```
Testing Time: 2.53s
  Unsupported:   5 (1.31%)
  Passed     :   2 (0.52%)
```

Both produced real artifacts (an `air.xclbin` plus four per-core `.o` files
each, matching the 4-core herds), and softmax still compiles for npu2 in both
formats at unchanged artifact sizes — xclbin 34234, elf 91072, the same figures
quoted in the PR body. `black`: both files unchanged.

PR #1880, opened 2026-08-24
[programming_examples] air.api: four more vector_examples primitives

erwei-xilinx:
Converts `vector_muladd`, `vector_reciprocal`, `vector_div` and `vector_tanh`,
replacing the raw-bindings versions. **No DSL code changed** — every operator
these need already exists; all four were probed before a line was written.
391 insertions, 519 deletions.

Each is one line of compute where the predecessor hand-rolled a vector loop: an
`scf.for` over the tile in steps of VECTOR_SIZE, two or three `memref.subview`s
per trip, `vector.transfer_read`s with an explicit identity permutation map and
a padding constant, the arith op, and a `transfer_write`.

```python
out[:] = alpha * b[:] + c[:]      # vector_muladd
c[:]   = 1.0 / a[:]               # vector_reciprocal
c[:]   = a[:] / b[:]              # vector_div
c[:]   = air.ops.tanh(a[:])       # vector_tanh
```

## The emitted arithmetic is op-for-op identical

| example | op | pred | conv |
|---|---|---:|---:|
| vector_muladd | `arith.mulf` | 1 | 1 |
| vector_muladd | `arith.addf` | 1 | 1 |
| vector_muladd | `vector.broadcast` | 1 | 1 |
| vector_reciprocal | `arith.divf` | 1 | 1 |
| vector_reciprocal | `vector.broadcast` | 1 | 1 |
| vector_div | `arith.divf` | 1 | 1 |
| vector_tanh | `math.tanh` | 1 | 1 |
| *(all four)* | `memref.subview` | 2 or 3 | **0** |

The subviews going to zero is the point: air.api reads at an offset directly.
For `vector_div` the vector width was checked too — `vector<64xf32>` and a
step-64 inner loop in both arms — so the op-count match is not a coincidence of
counting.

## Per-example notes

* **`vector_muladd` must keep emitting *two* ops.** It is one half of a matched
  pair with `vector_fma`, which computes the same value through `vector.fma`;
  the two exist to be compared, so anything that fused them would erase the
  distinction. `alpha * b[:] + c[:]` preserves it.
* **`vector_fma` is deliberately not converted.** air.api has no fma primitive,
  and expressing it as a multiply plus an add would silently turn it into
  `vector_muladd`. That is a scope refusal, not an oversight.
* **`vector_reciprocal` and `vector_div`** keep `--arch` selecting the vector
  width only, as before — aie2p is 32 for reciprocal and 64 for div, the widths
  their lits exercise. The device now comes from `--target`, defaulting to
  auto-detect.
* **`vector_tanh` checks `np_dtype_in` rather than honouring it.** f32 tanh does
  not legalize (`G_FTANH`) and scalar bf16 tanh does not either, so the
  emitter's usual scalar fallback is the *unsafe* direction here rather than a
  safety net; the `tile_n % vector_size` assert is load-bearing for the same
  reason. Its `--arch` is documented as accepted-and-unused, which is what the
  predecessor also did — the Makefile passes it.
* The three arithmetic examples honour `np_dtype_in` via `dtype_of` instead of
  hardcoding the element type, and reject an unknown one by name.

## Verification

10 lits. The 2 runnable on this npu1 box PASS from wiped output directories:

```
PASS: vector_reciprocal/run_npu1_makefile_peano.lit (9 of 10)
PASS: vector_muladd/run_makefile_peano.lit (10 of 10)
(8 UNSUPPORTED: npu2-gated)
```

The 8 npu2-gated ones are covered by a compile sweep at each lit's exact config,
with the predecessor **forced to npu2 via `AIR_TARGET_DEVICE`** as a control —
without that it has no `--target` and would silently have been building npu1
artifacts, which is not a control at all. Both arms build all 8:

```
config                    pred      conv
muladd npu2 vec16 elf    16912     18416
muladd npu2 vec32 xclbin 10073     10730
muladd npu2 vec32 elf    17104     18896
div    npu2 xclbin       18090     24698
div    npu2 elf          41136     60784
recip  npu2 xclbin        9753     10714
tanh   npu2 vec16         9401      9929
tanh   npu2 vec32         9401      9929
```

### Why the conversions are 6–48% larger

Worth stating rather than leaving for a reviewer to find. Two plausible causes
were checked and are **not** it:

* not the herd orientation — rebuilding `vector_div` with a `[1, NUM_TILES]`
  herd gives 24810 / 61248, i.e. unchanged;
* not `stack_size` — byte-identical with and without.

It is **ping-pong buffering**. The predecessors hoist their `memref.alloc`s
*outside* the outer loop, so the buffers are allocated once and
`air-ping-pong-transform` leaves them alone. air.api allocates inside the herd
body, so they are double-buffered: `aie.buffer` goes 6 → 12 for `vector_div`,
and the per-core object 5092 → 9212 bytes. That trades binary size and L1 for
DMA/compute overlap, and it is inherent to air.api rather than a per-example
choice — every conversion already merged behaves this way.

L1 stays well inside budget after the doubling. `vector_div` is the worst case
at 3 × 1024 × f32 = 12 KB declared, 24 KB with ping-pong, against 64 KB; and all
eight npu2 configs compile, which is the direct evidence.

**No perf A/B has been run.** These examples have no `profile` target and no
`test.cpp` harness, and the npu2 arm needs a box this is not. Given the
ping-pong finding above, that caveat now has a specific mechanism attached to
it and could plausibly cut either way: pipelining should help, the larger
footprint might not.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_div/vector_div.py
+++ b/programming_examples/primitives/vector_examples/vector_div/vector_div.py
@@ -1,169 +1,101 @@
 # Copyright (C) 2025, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
+"""Elementwise division primitive, on air.api.
+
+    c[:] = a[:] / b[:]
+
+One line of compute. The predecessor built the same expression as a hand-rolled
+vector loop -- an ``scf.for`` over the tile in steps of VECTOR_SIZE, three
+``memref.subview``s per trip, two ``vector.transfer_read``s with an explicit
+identity permutation map and a padding constant, ``arith.divf``, and a
+``transfer_write``. Here the emitter builds that loop, and the subviews are not
+needed at all: air.api reads at an offset directly.
+
+The emitted arithmetic is unchanged: a bare ``arith.divf`` on
+``vector<VECTOR_SIZExf32>``, with no broadcast -- both operands are buffers.
+That is the difference from the sibling ``vector_reciprocal``, whose numerator
+is a scalar and so carries a ``vector.broadcast``.
+
+f32 is deliberate and load-bearing: bf16 division does *not* legalize vectorised
+on either generation, so this primitive only exists in f32.
+
+Two differences from the predecessor worth naming:
+
+* The herd is [NUM_TILES, 1] rather than [1, NUM_TILES]. A 1-D air.api herd is
+  laid out along x, which is the orientation that places on both generations.
+* The strip-mine is the DSL's. The predecessor asked for a [1, NUM_TILES] herd
+  and then wrote the outer loop itself, computing ``_l_ivx + _ty * tile_n``
+  through a hand-built AffineMap. Here the herd's iteration space is the whole
+  tile grid and air.api strip-mines it onto NUM_TILES cores.
+
+``--arch`` still selects only the vector width, exactly as before; it does not
+pin the device. The generation comes from ``--target``, which defaults to
+detecting the installed part.
+"""
+
 import argparse
-from ml_dtypes import bfloat16
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, load, store, subview
-from air.dialects.vector import transfer_read, transfer_write
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.backend.xrt_runner import XRTRunner, type_mapper
+import numpy as np
+
+from air import api as air
+from air.api.types import dtype_of
 from air.backend.xrt import XRTBackend
+from air.backend.xrt_runner import XRTRunner
 
-range_ = for_
+NUM_TILES = 2
+
+# Vector width per architecture. Selects the SIMD width only -- it does not pin
+# which device the design is built for. Note aie2p is 64 here where the sibling
+# vector_reciprocal uses 32; both are the widths their lits exercise.
+ARCH_VECTOR_SIZES = {"aie2": 16, "aie2p": 64}
+
+# f32 scalar fdiv needs a deeper stack than the 1024-byte default.
+STACK_SIZE = 2048
 
 
-@module_builder
 def build_module(n, tile_n, np_dtype_in, arch="aie2"):
-    a_size = [n]
-    b_size = a_size
-    out_size = a_size
-    xrt_dtype_in = type_mapper(np_dtype_in)
-    num_tiles = 2
-    assert n % (tile_n * num_tiles) == 0
-    # Architecture-specific vector size
-    arch_vector_sizes = {
-        "aie2": 16,
-        "aie2p": 64,
-    }
-    VECTOR_SIZE = arch_vector_sizes.get(arch, 16)  # default to 16 if unknown
-    index_type = IndexType.get()
-
-    # L3 MemRefTypes
-    l3memrefTy = MemRefType.get(a_size, xrt_dtype_in)
-
-    # L1 MemRefTypes
-    l1MemrefTy = MemRefType.get(
-        shape=[tile_n],
-        element_type=xrt_dtype_in,
-        memory_space=IntegerAttr.get(T.i32(), MemorySpace.L1),
-    )
-
-    @FuncOp.from_py_func(l3memrefTy, l3memrefTy, l3memrefTy)
-    def vector_div(arg0, arg1, arg2):
-
-        @herd(
-            name="herd_0",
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Both accepted and fixed in 3a0e7feb — they are the same defect in two files, and
a real one rather than a style nit: each docstring *states* the f32-only
constraint and then `build_module` accepted anything `dtype_of()` knows. A
documented contract with no check is worse than no contract, because the caller
gets a backend crash several passes downstream instead of a one-line error.

I verified the claim before encoding it rather than trusting my own docstring —
the failure mode this repo keeps hitting is a legalization rule generalised from
one data point, so "bf16 division does not legalize" needed to be measured.
Compiling `1.0/a` and `a/b` across dtypes and both generations:

```
kind    dtype     target  vec  result
recip   float32   npu1     16  COMPILES
recip   float32   npu2     32  COMPILES
recip   bfloat16  npu1     16  FAILS [legalize]
recip   bfloat16  npu2     32  FAILS [legalize]
recip   float16   npu1/2       FAILS [legalize]
recip   int32     npu1/2       FAILS [legalize]
div     float32   npu1     16  COMPILES
div     float32   npu2     32  COMPILES
div     bfloat16  npu1/2       FAILS [legalize]
div     float16   npu1/2       FAILS [legalize]
div     int32     npu1/2       FAILS [legalize]
```

f32 is the only element type that builds, so the docstrings were right — and
`f16` and `i32` fail identically, which neither docstring mentioned. The check
now names all three, and both files quote the actual error so the next reader
does not have to rerun this:

```
LLVM ERROR: unable to legalize instruction:
%42:_(<16 x s16>) = G_FDIV %37:_, %41:_ (in function: core_0_2)
```

Guards firing, and the f32 path unchanged:

```
vector_reciprocal: f32 OK, vector<32xf32>
  bfloat16: division legalizes only for f32 on AIE (bf16/f16/i32 all fail ...
  float16:  division legalizes only for f32 on AIE (bf16/f16/i32 all fail ...
  int32:    division legalizes only for f32 on AIE (bf16/f16/i32 all fail ...
vector_div: f32 OK, vector<64xf32>
  (same three)
```

Regression check — the npu2 artifacts are byte-identical to before the guard
(div xclbin 24698, div elf 60784, recip xclbin 10714), and both npu1 lits still
PASS from wiped output directories:

```
PASS: vector_muladd/run_makefile_peano.lit (9 of 10)
PASS: vector_reciprocal/run_npu1_makefile_peano.lit (10 of 10)
```

`black --check`: both files unchanged.

erwei-xilinx:
Correction to my previous comment, and to the code comment it described.

I wrote that "`f16` and `i32` fail identically" to bf16. Two of the three are
right; **`i32` is not**, and it is the one dtype whose behaviour this conversion
actually changed. Fixed in d8419b9a (comment only, no code change).

First, the coverage question: **no lit ever exercised any of this.** Both
predecessors hardcode `INPUT_DATATYPE = np.float32` and expose no dtype flag,
and no lit or Makefile passes anything dtype-related — CI covered exactly one
element type in either arm. So none of it was gated; the question is only
whether the conversion changed the failure.

Three-arm A/B — predecessor at `origin/main`, this PR with the guard stripped,
and this PR as shipped:

```
example            dtype     PREDECESSOR            THIS PR (no guard)     shipped
vector_div         float32   COMPILES               COMPILES               COMPILES
vector_div         bfloat16  legalize (<16 x s16>)  legalize (<16 x s16>)  guard
vector_div         float16   legalize (<16 x s16>)  legalize (<16 x s16>)  guard
vector_div         int32     build-time MLIRError   legalize (G_SDIV)      guard
vector_reciprocal  float32   COMPILES               COMPILES               COMPILES
vector_reciprocal  bfloat16  legalize (<16 x s16>)  legalize (<16 x s16>)  guard
vector_reciprocal  float16   legalize (<16 x s16>)  legalize (<16 x s16>)  guard
vector_reciprocal  int32     build-time MLIRError   legalize (G_SDIV)      guard
```

**bf16 and f16 fail identically in both arms — inherited, not introduced.**

**i32 does not:**

```
predecessor:  MLIRError: Invalid attribute: expected floating point type
this PR:      LLVM ERROR: unable to legalize instruction:
              %117:_(<16 x s32>) = G_SDIV %113:_, %116:_ (in function: core_1_2)
```

The predecessor asks for a float divide on an integer memref and cannot build
the module at all. air.api emits a real *integer* divide — the more correct IR —
and AIE then turns out to have no vector integer divide either. So no working
configuration was lost, but the failure moved later and got less legible: a
regression in diagnostics even though it is an improvement in the IR.

Both are moot in the shipped code, which rejects the dtype at the call site with
a named error before either path is reached. The code comment now says this
rather than lumping all three together.

PR #1881, opened 2026-08-24
[python] air.api: comparisons and ops.select, and vector_select converted

erwei-xilinx:
Adds the first expression node in `air.api` whose result type is **not** the
element type, and converts the example that needs it. 6 files, +525/−164.

A comparison yields `i1` — `vector<Wxi1>` when vectorised — where every other
node in the expression tree yields the element type and feeds an arith builder
that infers its result from operand 0. So comparisons get their own node kind
and exactly one legal consumer:

```python
c[:] = air.ops.select(a[:] >= b[:], a[:], b[:])
```

## Surface

* **`<`, `<=`, `>`, `>=`** on a buffer expression build a predicate node. Float
  comparisons use the *ordered* predicates (`oge`, not `uge`): false when either
  operand is NaN, which is C's `>=` and what the hand-written kernel named as
  `CmpFPredicate.OGE`. Integer comparisons use the signed predicates.
* **`ops.equal` / `ops.not_equal` rather than `==` / `!=`.** Defining `__eq__`
  would set `__hash__` to `None` and make every expression unhashable, and would
  change what `expr == expr` means for ordinary Python. The cost of *not*
  overloading it is that `a[:] == b[:]` silently evaluates to a bool — so
  `ops.select` rejects a plain bool by name and points at `ops.equal`. That is
  the one thing here that could have failed quietly, so it fails loudly and is
  pinned by a test.
* **`ops.select(cond, a, b)`** → `arith.select`. A non-comparison condition is
  rejected at trace time, which is the only place the i1-vs-element-type
  mismatch can be reported against the user's own line rather than deep in
  verification.

## One read per buffer per loop body, not one per mention

`select(a >= b, a, b)` names each buffer twice and `gelu` names `x` four times;
the emitter was issuing a redundant `vector.transfer_read` for every extra
mention, so `vector_select` came out with 4 reads against the predecessor's 2. A
per-body cache keyed on the buffer fixes it — **gelu 5 reads → 1, silu 2 → 1**,
and vector_select 4 → 2, which is what brings the conversion to op-for-op parity.

**Be clear what that is worth.** The npu2 artifacts for gelu, silu and sigmoid
are the *same size* before and after (20055 / 19287 / 18778 bytes), so the aircc
pipeline was already CSE-ing those reads. This is an IR-quality change that
earns the parity claim below — not a measured speedup, and I am not claiming one.

## vector_select is op-for-op identical to its predecessor

| op | pred | conv |
|---|---:|---:|
| `arith.cmpf` (oge, `vector<16xbf16>`) | 1 | 1 |
| `arith.select` (`vector<16xi1>`, `vector<16xbf16>`) | 1 | 1 |
| `vector.transfer_read` | 2 | 2 |
| `vector.transfer_write` | 1 | 1 |
| `memref.subview` | 3 | **0** |
| `arith.constant` | 10 | 9 |

`select(a >= b, a, b)` is `max(a, b)`, and `ops.maximum` would lower it to a
single `arith.maximumf`. Both spellings are kept deliberately: `vector_max` and
`vector_select` exist to be compared, so writing this one as a maximum would
erase the distinction the pair is for — the same call already made for
`vector_muladd` vs `vector_fma`.

## Verification

* `python/test/api/select.py` — 8 FileCheck cases: predicate types, all four
  ordering operators, `equal`/`not_equal`, signed `cmpi` for integers, scalar
  broadcast, composition inside a larger tree, and the scalar fallback.
* `python/test/api/errors.py` — 4 negative cases, including the `==` trap.
* **Full api suite 15/15** after the emitter change (14 pre-existing + select).
* npu1 lits, from wiped output directories:

```
PASS: primitives/vector_examples/vector_select/run_makefile_peano.lit
PASS: primitives/vector_examples/vector_select/run_makefile_peano_bf16_emulation.lit
PASS: relu/run_makefile_peano.lit
PASS: leaky_relu/run_makefile_peano.lit
```

`relu` and `leaky_relu` are there as read-cache regression cover, not as scope.

* The 3 npu2-gated `vector_select` configs compile (vec16 elf 18416, vec32
  xclbin 10506, vec32 elf 18224), as do gelu/silu/sigmoid on npu2 — those three
  are the examples the read cache changes most and are npu2-only, so they cannot
  be run on this box.

**No perf A/B has been run.**

### Reviewer note on blast radius

Unlike the preceding conversion PRs, this one touches shared DSL code. The read
cache changes the emitted IR of **every** `air.api` example, not just the one
converted here — that is why gelu/silu/sigmoid are covered above despite being
out of scope, and why the artifact-size comparison is included rather than
asserted.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_select/vector_select.py
+++ b/programming_examples/primitives/vector_examples/vector_select/vector_select.py
@@ -1,170 +1,105 @@
 # Copyright (C) 2026, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
-from ml_dtypes import bfloat16
+"""Elementwise select primitive, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp, cmpf, select, CmpFPredicate
-from air.dialects.memref import AllocOp, DeallocOp, load, store, subview
-from air.dialects.vector import transfer_read, transfer_write
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+    c[:] = air.ops.select(a[:] >= b[:], a[:], b[:])
+
+One line of compute. The predecessor built the same expression as a hand-rolled
+vector loop -- an ``scf.for`` over the tile in steps of VECTOR_SIZE, three
+``memref.subview``s per trip, two ``vector.transfer_read``s with an explicit
+identity permutation map and a padding constant, ``arith.cmpf`` with
+``CmpFPredicate.OGE``, ``arith.select``, and a ``transfer_write``. Here the
+emitter builds that loop, and the subviews are not needed at all: air.api reads
+at an offset directly.
+
+The emitted arithmetic is unchanged: ``arith.cmpf oge`` producing a
+``vector<VECTOR_SIZExi1>`` predicate, feeding ``arith.select``.
+
+A comparison is the first thing in air.api's expression language whose result
+type is *not* the element type, which is why ``select`` is a function rather
+than an operator: the i1 a comparison yields has exactly one legal consumer.
+The ordering comparisons are operators (``<``, ``<=``, ``>``, ``>=``) and map to
+the *ordered* float predicates -- ``oge``, not ``uge``, so the result is false
+when either operand is NaN, matching C and matching what the predecessor named.
+Equality is ``ops.equal`` / ``ops.not_equal`` rather than ``==`` / ``!=``,
+because overloading ``__eq__`` would make every expression unhashable and would
+change what ``expr == expr`` means for ordinary Python.
+
+Note what this kernel computes: ``select(a >= b, a, b)`` is ``max(a, b)``, and
+the reference in ``__main__`` is written as ``max()``. ``ops.maximum`` would
+lower it to a single ``arith.maximumf``. Both spellings are kept because
+``vector_max`` and this example exist to be compared -- writing this one as a
+maximum would erase the distinction the pair is for.
+
+Two differences from the predecessor worth naming:
+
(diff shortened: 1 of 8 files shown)


erwei-xilinx:
CI failure investigated and fixed in 3ffef589. It was my bug, and worse than it
looked: **all 15 api tests failed, not just the new one.**

```
File ".../python/test/api/select.py", line 21, in <module>
    from air.api.types import bf16, f32, i32
ImportError: cannot import name 'bf16' from partially initialized module
'air.api.types' (most likely due to a circular import)
```

reached from:

```
numpy/__init__.py -> numpy/lib/_utils_impl.py -> platform -> subprocess
  -> import select      <- picks up python/test/api/select.py
```

lit puts the test's own directory on `sys.path`, so a test named `select.py` can
win the import of the stdlib module of that name. It then executes in the middle
of numpy's own import, tries to import `air.api.types` while `air.api.types` is
itself still importing numpy, and dies. Every test in the directory goes down
with it.

**Why it passed locally and failed here.** `select` is *statically built in* on
both Pythons on my machine (3.12.10 and 3.10.12), so the path-based finder is
never consulted and the clash is invisible. GitHub's hosted Python 3.10.21 ships
it as a shared extension in `lib-dynload`, where `sys.path` order decides. I
verified both directions rather than inferring: restoring the file and importing
it under the name `select` succeeds — it really would shadow — and after the
rename there is nothing to shadow.

Renamed to `cmp_select.py`, which also describes it better since it covers the
comparison operators as well as `select`.

I also added `python/test/api/naming.py`, which fails if any test file in the
directory is named after a stdlib module. It is only luck that a runner exists
where this fires — a differently-named offender (`types.py`, `signal.py`) might
pass everywhere until it did not, and the failure mode points at numpy rather
than at the real cause. Checked against a planted `select.py`:

```
FAIL: AIRPYTHON :: api/naming.py
stdlib-shadowing test files: ['select.py']
```

Full api suite **16/16** (15 + the guard). The four Linux jobs should go green;
the Ryzen AI hardware jobs were already passing, which is what made this look
narrower than it was.

PR #1882, opened 2026-08-24
[python] air.api: bitwise &, | and ^, and the three vector_examples converted

erwei-xilinx:
Adds the DSL's first **integer-only** operators and converts the three examples
that need them. 7 files, +449/−354.

```python
c[:] = a[:] & b[:]      # vector_andi -> arith.andi
c[:] = a[:] | b[:]      # vector_ori  -> arith.ori
c[:] = a[:] ^ b[:]      # vector_xori -> arith.xori
```

Every other binary operator has both a float and an integer form, so
`_FLOAT_OPS` and `_INT_OPS` are parallel tables. These three exist only in the
integer one: MLIR has `arith.andi` and no floating-point counterpart, and there
is no coercion that would invent one. A float buffer reaching them raises naming
the operator and the reason, rather than falling through the generic "not
supported for dtype float" — `&` on a float buffer is usually a dtype mistake
upstream, and the message should say so.

Reflected forms (`__rand__`/`__ror__`/`__rxor__`) are included, so `0xff & a[:]`
preserves operand order rather than silently commuting.

## Op-for-op parity, all three

| op | pred | conv |
|---|---:|---:|
| `arith.andi` / `ori` / `xori` | 1 | 1 |
| `vector.transfer_read` | 2 | 2 |
| `vector.transfer_write` | 1 | 1 |
| `arith.constant` | 9 | 9 |
| `air.dma_memcpy_nd` | 3 | 3 |
| `memref.alloc` / `dealloc` | 3 | 3 |
| `memref.subview` | 3 | **0** |
| `affine.apply` | 1 | 3 |

Only the incidental subviews go away; `affine.apply` becomes one per DMA rather
than one shared.

## Verification

* `python/test/api/bitwise.py` — 5 FileCheck cases: the three ops, a scalar
  mask, reflected operands, a narrower integer type, and the scalar fallback.
* `python/test/api/errors.py` — 2 negative cases: float (refused by name) and
  unsigned (refused earlier by the existing signless guard, for a *different*
  reason). Both pinned so the two paths do not get conflated.
* **Full api suite 15/15.**
* npu1 lits, from wiped output directories:

```
PASS:  vector_andi/run_makefile_peano.lit
PASS:  vector_ori/run_makefile_peano.lit
XFAIL: vector_xori/run_npu1_makefile_peano.lit
```

### The XFAIL is inherited, and that was checked rather than assumed

Both arms die in the same place, with only the SSA number differing:

```
pred:  unable to legalize instruction: %61:_(<16 x s32>) = G_XOR
conv:  unable to legalize instruction: %118:_(<16 x s32>) = G_XOR
```

That is the AIE2 GISel gap the lit's own comment documents — no legalizer rule
for vector `G_XOR`, no VBXOR instruction. AIE2P legalizes it natively, and the
npu2 build confirms that. So the conversion is not satisfying the XFAIL
coincidentally for some new reason.

### npu2 compile sweep

Predecessor forced to npu2 via `AIR_TARGET_DEVICE` as a control — it has no
`--target`, so without that it silently builds npu1 artifacts and is not a
control at all.

```
config                     pred      conv
vector_andi npu2 xclbin    9945      10538
vector_ori  npu2 xclbin    9945      10538
vector_xori npu2 xclbin    9977      10634
```

The ~6% growth is the ping-pong buffering inherent to air.api — predecessors
hoist their allocs outside the outer loop; air.api allocates inside the herd
body and `air-ping-pong-transform` double-buffers. Same cause as #1880, where it
is documented along with the measurements ruling out herd orientation and
`stack_size`.

**No perf A/B has been run.**

### Merge-order note

This touches the same two files as #1881 (`ops.select`) — `_value.py` adds
operator methods to `BufferExpr`, `_emit.py` adds `_INT_OPS` entries — so expect
a small textual conflict if both land. They are independent features and neither
depends on the other; this branched off `main` rather than stacking.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_andi/vector_andi.py
+++ b/programming_examples/primitives/vector_examples/vector_andi/vector_andi.py
@@ -17,9 +17,13 @@ The emitted arithmetic is unchanged: a bare ``arith.andi`` on
 Bitwise operators are integer-only, and that is a property of MLIR rather than
 of this example: there is ``arith.andi`` and no floating-point counterpart, so
 air.api refuses ``&``, ``|`` and ``^`` on a float buffer by name instead of
-letting them fall through a generic "unsupported operator". i32 is checked below
-for the same reason -- see the sibling ``vector_ori / vector_xori`` examples, which differ
-from this one only in that operator.
+letting them fall through a generic "unsupported operator".
+
+What ``build_module`` checks is that the element type is a *signless integer* --
+not that it is i32 specifically. These operators work at any integer width, and
+air.api's own test suite covers i16; i32 is simply what ``__main__`` passes,
+matching the predecessor. See the sibling ``vector_ori / vector_xori`` examples, which
+differ from this one only in the operator.
 
 Two differences from the predecessor worth naming:
 
--- a/programming_examples/primitives/vector_examples/vector_andi/vector_andi.py
+++ b/programming_examples/primitives/vector_examples/vector_andi/vector_andi.py
@@ -1,131 +1,97 @@
 # Copyright (C) 2026, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
+"""Elementwise bitwise AND primitive, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, subview
-from air.dialects.vector import transfer_read, transfer_write
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+    c[:] = a[:] & b[:]
+
+One line of compute. The predecessor built the same expression as a hand-rolled
+vector loop -- an ``scf.for`` over the tile in steps of VECTOR_SIZE, three
+``memref.subview``s per trip, two ``vector.transfer_read``s with an explicit
+identity permutation map and a padding constant, ``arith.andi``, and a
+``transfer_write``. Here the emitter builds that loop, and the subviews are not
+needed at all: air.api reads at an offset directly.
+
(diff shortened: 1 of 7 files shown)


erwei-xilinx:
Both accepted and fixed in 1942bbbd.

## "i32 is checked below" — the docstring was wrong, not the code

Correct: `build_module` rejects float and unsigned and accepts any signless
integer. Of your two suggested resolutions I took the first, because the second
would have been a regression: these operators work at any integer width, and
air.api's own test suite has an i16 case (`narrower_integer` in
`python/test/api/bitwise.py`). Narrowing the check to i32 would have contradicted
that test in order to make a sentence true.

Reworded to state what is actually enforced, and that i32 is simply what
`__main__` passes. Confirmed against the guard:

```
int16 (should PASS)    accepted
float32                the bitwise operator '&' is integer-only ...
uint8                  the bitwise operator '&' is integer-only ...
```

## `scalar_fallback` exercised the route it did not describe

Also correct, and a better catch than it looks. The emitter vectorises only when

```
bool(shape) and width > 0 and shape[-1] % width == 0
```

so there are **two** independent ways to reach the scalar loop, and the test hit
`width == 0` while its comment described the remainder condition. Split into two:

* `scalar_fallback_no_vector_width` — width == 0, the caller opting out
* `scalar_fallback_tile_not_a_multiple` — 24 % 16 = 8, the one a user hits by
  accident

And checked the new one is not passing vacuously — it emits a real scalar loop,
step 1:

```
scf.for %arg11 = %c0_3 to %c24 step %c1_4 {
  %3 = memref.load %alloc[%arg11] : memref<24xi32, 2 : i32>
  %5 = arith.andi %3, %4 : i32
  memref.store %5, %alloc_2[%arg11] : memref<24xi32, 2 : i32>
```

A test that exercises one route while claiming the other is worse than no test,
because it reads as coverage.

**The same flaw was in `cmp_select.py` on #1881** — I wrote the same pattern
there. Fixed separately in e94cfd5c on that branch rather than smuggled into
this one.

api suite 15/15 here; npu1 lits unchanged (2 PASS, 1 XFAIL inherited).

PR #1883, opened 2026-08-24
[python] air.api: refuse `and`/`or`/`not` on a buffer expression

erwei-xilinx:
`and`, `or` and `not` are the one part of Python's operator surface a library
cannot reach. There is no dunder for them: they coerce the operand through
`__bool__` and short-circuit. `BufferExpr` defined neither `__bool__` nor
`__len__`, so they took default object truthiness — always `True`:

```
a[:] and b[:]  ->  Buffer(...)   the 2nd operand. No error, no IR.
a[:] or b[:]   ->  Buffer(...)   the 1st operand.
not a[:]       ->  False
if a[:]:                         always taken
```

So `c[:] = a[:] and b[:]` compiled, ran, and copied `b` — a kernel that silently
computes half of what was written.

This is the same class as `==` on a buffer expression, and worse: `==` at least
yields an obviously wrong `bool`, where this yields a plausible-looking
`BufferExpr` that has quietly dropped an operand. Fixed the way NumPy fixes it —
`__bool__` raises, naming what to use instead.

## Also: the internal op key no longer reaches a repr

The tree stores `"add"`, not `"+"`, so a binary node printed as
`(Buffer add Buffer)` — which reads like a typo. It would read far worse once
the bitwise operators land, where `a[:] & b[:]` would print
`(Buffer and Buffer)`, naming the very operator this guard rejects.

Infix keys now print their symbol, and keys with no infix form print the call
the user actually wrote:

```
(Buffer(...) + Buffer(...))
(Buffer(...) * 2.0)
maximum(Buffer(...), 0.0)        <- ops.maximum, not the internal "max"
```

The symbol table lists the comparison and bitwise keys too, so it will not need
revisiting when those land.

## Verified

* `python/test/api/truthiness.py` — `and` / `or` / `not` / `if`, plus the repr
  forms.
* **Full api suite 15/15.**
* **No internal code depends on truthiness of a `BufferExpr`** — every
  `if <name>:` in `python/air/api/` tests a list, a bool, a match object or a
  string. Checked rather than assumed, then confirmed on hardware: 7 air.api
  example lits pass from wiped output directories (axpy, eltwise_add,
  eltwise_add_with_l2, leaky_relu, relu, vector_add ×2), 0 failures.

## Scope

Deliberately `BufferExpr` only, not `Buffer`. `a[:] and b[:]` is caught; bare
`a and b` on the unsubscripted buffers is not. `if buf:` and `if buf is not
None` are plausible things to write about a `Buffer`, so the risk/benefit there
is worse. Easy to extend later if it proves needed.

## Merge order

**Please land after #1881 and #1882.** The error message tells the user to use
`&` / `|` / `^` and `air.api.ops.select` — `&` arrives with #1882 and
`ops.select` with #1881. Merged ahead of them, the message would send someone to
surface that does not exist yet, which is worse than the silent bug it replaces.

The bug itself is pre-existing on `main` and independent of both, which is why
this is its own PR rather than folded into either. The three touch `_value.py`
in different regions (both others insert after `__neg__`; this adds `__bool__`
before `leaves()` and a table above the class), so a clean merge is likely
regardless of order.

### DIFF
--- a/python/air/api/_value.py
+++ b/python/air/api/_value.py
@@ -324,6 +324,32 @@ class BufferSlice:
         return f"BufferSlice({self.buffer!r}, sizes={self.sizes})"
 
 
+# How a binary node prints. The tree stores an internal key ("add"), not the
+# source spelling, so without this a repr shows `(Buffer add Buffer)` -- which
+# reads like a typo. Keys with no infix form fall back to a call spelling.
+# Entries are listed for every key the emitter knows, including the comparison
+# and bitwise ones, so the table does not need revisiting when those land.
+_OP_SYMBOLS = {
+    "add": "+",
+    "sub": "-",
+    "mul": "*",
+    "div": "/",
+    "and": "&",
+    "or": "|",
+    "xor": "^",
+    "lt": "<",
+    "le": "<=",
+    "gt": ">",
+    "ge": ">=",
+    "eq": "==",
+    "ne": "!=",
+}
+
+# Keys with no infix spelling print as the call the user actually wrote, so the
+# internal key never reaches a repr either way: ops.maximum, not "max".
+_OP_CALL_NAMES = {"max": "maximum", "min": "minimum"}
+
+
 class BufferExpr:
     """A lazy elementwise expression over buffers and scalars.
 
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Accepted, fixed in 15013c36 — and it was the one thing this PR's description had
to apologise for, so it is good to have it closed properly rather than
documented.

I took the remedy a different way than suggested, though. Hedging every
suggestion with "if supported/available" costs the *useful* case its precision,
and the useful case is the common one: someone who has `&` should be told to use
`&`, not told they might have it. So the redirection is now assembled from what
the build actually has:

```python
hasattr(type(self), "__and__")   -> suggest `&`, `|`, `^`
hasattr(ops, "select")           -> suggest ops.select(cond, a, b)
```

which is exact in every combination and needs no qualification. All four states
are pinned in `truthiness.py`:

```
neither:      (no advice)
& only:       Use the elementwise operators `&`, `|`, `^` for logic.
both:         Use the elementwise operators `&`, `|`, `^` for logic or
              air.api.ops.select(cond, a, b) to choose between values.
select only:  Use air.api.ops.select(cond, a, b) to choose between values.
```

The first clause — what went wrong, and that it would have emitted no kernel
code — is unconditional, so the diagnosis never degrades. Only the redirection
adapts.

**This also removes the merge-order constraint** this PR was carrying. Landing
it ahead of #1881 / #1882 no longer produces a message pointing at surface that
does not exist; it just produces a shorter one. That is still the sensible
order, but it is no longer a correctness requirement.

api suite 15/15.

PR #1885, opened 2026-08-24
[python] air.api: ops.cast, and the two int<->float conversion examples

erwei-xilinx:
Adds `air.api.ops.cast(expr, dtype)` and converts
`primitives/vector_examples/vector_fptosi` and `vector_sitofp` onto it,
deleting their raw-bindings implementations.

A cast is the first node whose operand has a different element type from its
result, so it is the first thing that makes one assignment hold two element
types at once. Everything below it is read and computed in the source type and
only the conversion lands in the destination's, which is observable:
`cast(a[:] * 2.0, i32)` doubles in f32 and converts once, while
`cast(a[:], i32) * 2` converts first and doubles in i32.

The emitter used to carry one element type, one arith table, one vector type
and one padding constant as four arguments threaded down `_eval`. They are now
grouped into a `_Region` and looked up per node.

## Which conversions are allowed, and why

Each pair was run on npu1 hardware against an **exact** numpy reference,
vectorised and scalar, because compiling is a weaker claim than computing — of
the pairs probed, four compiled cleanly and returned wrong values.

| pair | verdict |
|---|---|
| f32↔i32, bf16→f32, i16→i32, bf16→i32 | exact on both routes |
| f32→bf16, i32→bf16 | 1 ULP: AIE rounds toward −∞, numpy to nearest-even |
| i32→i16 | **refused** |

`i32→i16` is refused because the vectorised `arith.trunci` **saturates**
(933033 → 32767) while the scalar one **wraps** (→ 15529), and the emitter picks
between those two paths from the tile size. Accepting it would make the same
source compute two different things depending on a tile shape. The float
rounding is uniform across both routes, so it is documented rather than refused
— and it is not specific to this op, since a plain `c[:] = a[:] + b[:]` on bf16
buffers differs from numpy the same way.

`ops.relu` picked its zero constant from the first buffer leaf, which is the
type *below* a cast rather than the expression's own. It survived by accident
(`0.0` coerces to `0` for an integer region); it now asks the expression, and a
test pins it.

## Verification

- 17/17 api lits; new `python/test/api/cast.py` plus 7 negative cases in
  `errors.py`.
- Both example lits PASS on npu1 hardware. A mutation test confirms they are not
  vacuous: corrupting the golden reference by +1 makes them fail with
  `expected=-577385, actual=-577386`.
- npu2 xclbin and elf compile; the **scalar route runs clean on hardware** for
  both examples, which is the claim the docstrings make.
- The emitted IR of all 86 air.api example modules (npu1 and npu2, 265 KB) is
  **byte-identical to main** — that is what makes the `_eval` refactor safe, and
  the api lits alone would not have caught an ordering change.
- Regression sweep of all 45 directories containing an air.api example:
  58 PASS / 1 FAIL, the failure being matvec's `PROFILE` step, which fails
  identically on unmodified main (this build has no XRT for the C++ harness).

**No perf A/B was run** — no npu2 box in this session and neither example has a
profile target.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_fptosi/vector_fptosi.py
+++ b/programming_examples/primitives/vector_examples/vector_fptosi/vector_fptosi.py
@@ -1,122 +1,124 @@
 # Copyright (C) 2026, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
+"""Elementwise f32 -> i32 conversion primitive, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, subview
-from air.dialects.vector import transfer_read, transfer_write
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+    c[:] = air.api.ops.cast(a[:], air.api.i32)
+
+The predecessor built this as a hand-rolled vector loop -- an ``scf.for`` over
+the tile in steps of VECTOR_SIZE, two ``memref.subview``s per trip, a
+``vector.transfer_read`` with an explicit identity permutation map and a padding
+constant, ``arith.fptosi``, and a ``transfer_write``. Here the emitter builds
+that loop, and the subviews are not needed at all: air.api reads at an offset.
+
+The emitted arithmetic is unchanged: a bare ``arith.fptosi`` from
+``vector<VECTOR_SIZExf32>`` to ``vector<VECTOR_SIZExi32>``.
+
+**This is the first example where the two L1 tiles have different element
+types.** Every air.api expression until now was written in one type, and the
+emitter carried a single element type, arith table and vector type down the
+whole tree. ``ops.cast`` is the node where those change: everything below it is
+read and computed in the source type, and only the conversion lands in the
+destination's. That distinction is visible, and load-bearing:
+``cast(a[:] * 2.0, i32)`` doubles in f32 and converts once, while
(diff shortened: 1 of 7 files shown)


erwei-xilinx:
Both parts of this comment are correct, and one of them was a real regression rather than just a misleading message. Addressed in `89b16c1b`.

**`ops.relu` was a regression.** `ops.select` deliberately permits every buffer to sit in the predicate — its guard is `not a.leaves() and not b.leaves() and not cond.leaves()` — so `relu(ops.select(a[:] > 0, 1, 2))` is a legal expression. Verified against `main` and against this branch, same input:

```
main:            ops.relu over it: works
this branch:     ops.relu over it: ValueError: needs a buffer operand, got a scalar
```

`relu` now falls back to the first leaf's dtype when `element_dtype()` returns `None` but leaves exist, and only rejects a genuinely scalar expression. The fallback is exactly right rather than merely convenient: `_check_region` requires every leaf in a region to match that region's element type, so the leaf's dtype *is* the type the surrounding region will be built in.

**`ops.cast` was a misleading message, not a bug.** The refusal is correct — such an expression has no element type of its own to convert from, and it cannot simply be adopted into the target type either, because the predicate's own operands would then have to be i32 as well, which is not what was written. But "is a bare scalar" is untrue of it. The two cases are now distinguished:

```
no buffer leaves at all  -> "... is a bare scalar, which is already built in
                             whatever element type surrounds it"
buffers only in a cmp    -> "... its buffers appear only in a comparison, and
                             the values it evaluates to are all scalars, so it
                             has no element type of its own to convert from.
                             Cast the operands instead -- for example
                             ops.select(a[:] > 0, ops.cast(x[:], t), ops.cast(y[:], t))"
```

Both now have tests, since neither form had coverage: `relu_over_a_predicate_only_select` in `cast.py` pins that it lowers (`cmpf` → `select` → `maximumf`), and `cast_of_a_predicate_only_expression` in `errors.py` pins the targeted message.

Re-verified: 17/17 api lits, and the emitted IR of all 86 air.api example modules is still byte-identical to main.

PR #1886, opened 2026-08-24
[python] air.api: refuse arithmetic on f16 instead of miscomputing it

erwei-xilinx:
Neither NPU generation has an fp16 instruction, scalar or vector — bf16 is the
16-bit float AIE2 and AIE2P implement. What made this worth a guard is not the
gap but how the gap presented: **nothing in the toolchain refused an f16
kernel**. It compiled, it ran, no legalizer complained, and it returned the
result of having read the f16 bit patterns as bf16.

Measured on npu1 before this change:

```
c[:] = a[:] + b[:]  over f16 buffers   wrong on 2048 of 2048 elements
f16 512.5 is 0x6001; 0x6001 read as bf16 is 1.0078125 x 2^65
the device returned                    3.718172e19
```

A silently wrong kernel with no error is the failure mode this package exists to
remove.

## Movable, not computable

The split is measured rather than assumed:

```
f16   dma  :    0/2048 differ   EXACT
f16   copy :    0/2048 differ   EXACT
f16   add  : 2048/2048 differ
```

A copy emits a read and a write and no arith op, so the bits arrive unchanged.
So f16 now behaves the way unsigned already does — declarable, allocatable and
movable, with only arithmetic refused. Refusing movement too would make f16
undeclarable rather than uncomputable, and a kernel that moves f16 data and
hands it to an `air.extern` kernel is a real thing to want.

`require_computable` mirrors `require_signless` and is called from the same four
sites that build an arith or linalg op: the elementwise emitter, `ops.dot`, the
micro-tiled fill, and `air.extern`'s scalar arguments. The message names the
hardware reason and the type to use instead — "not supported" alone would send
the reader looking for a missing feature in air.api rather than a missing
instruction in the core.

## Verification

- 17/17 api lits, including a new `python/test/api/no_fp16.py` that pins the
  refusals, the still-allowed copy, and that bf16/f32 are untouched.
- No example is affected: nothing under `programming_examples/` declares f16.
- The emitted IR of all 86 air.api example modules (npu1 and npu2) is
  byte-identical to main.

## Related, not fixed here

The same sweep found that **bf16 arithmetic rounds toward −∞ rather than to
nearest-even** — `add`, `sub` and `mul` all match a floor model on 2048/2048
elements. That is a separate defect with a separate cause
(`set.ctrl.reg(i32 6, i32 0)` emitted by mlir-aie's `AIECoreToStandard.cpp`),
and it is not addressed by this change.

### DIFF
--- a/python/air/api/_emit.py
+++ b/python/air/api/_emit.py
@@ -28,7 +28,7 @@ from air.dialects.memref import load as memref_load, store as memref_store
 from air.dialects.scf import for_ as range_, yield_
 from air.dialects.vector import broadcast, transfer_read, transfer_write
 
-from .types import require_signless
+from .types import require_computable, require_signless
 
 __all__ = ["emit_elementwise"]
 
@@ -105,6 +105,18 @@ def emit_elementwise(dst, expr):
                 "buffer used before allocation; air.alloc() must be called "
                 "inside the herd body that uses it"
             )
+
+    # A tile the core has no instructions for -- f16 -- follows the same rule
+    # as an unsigned one: it can be *copied* elementwise, because a copy emits a
+    # read and a write and no arith op, and the bits arrive unchanged. Measured
+    # on npu1: an f16 copy is exact on 2048 of 2048 elements, and an f16 add is
+    # wrong on 2048 of 2048. Only the second is refused.
+    if not dst.dtype.computes and expr.kind != "buffer":
+        require_computable(
+            dst.dtype,
+            "an elementwise operator or broadcast scalar (a plain copy, "
+            "dst[:] = src[:], is)",
+        )
 
     if dst.dtype.is_unsigned:
         # An unsigned tile can be *copied* elementwise but not computed on. The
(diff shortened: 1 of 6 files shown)


erwei-xilinx:
Correct, and it is the "docstring overclaims what the code does" failure again — the sentence lumped f16 in with unsigned when the two behave differently. Fixed in `7b05d55a`.

Measured rather than reasoned:

```
f16 plain copy: default_vector_width=16  vector types emitted=['16']  -> width REACHED
ui8 plain copy: default_vector_width=32  vector types emitted=none    -> never reached
```

Unsigned really is dead surface — an unsigned buffer is forced onto the scalar path even for a copy, because `vector.transfer_read` takes a padding value and that padding value is an `arith.constant`. f16 is not: arithmetic is refused, but a plain copy is allowed and still vectorises to `vector<16xf16>`.

The docstring now separates them and says where the 16 comes from — it is measured, not extrapolated: an f16 copy at 16 lanes is exact on 2048 of 2048 elements on npu1, which is unsurprising, since a copy moves bit patterns and never asks the core to interpret them as floats.

`no_fp16.py`'s copy case now also pins `vector<16xf16>`, so the claim in the docstring has a test behind it rather than being prose that can drift again.

PR #1888, opened 2026-08-24
[python] air.api: ops.fma, and the vector_fma example converted

erwei-xilinx:
Adds a fused multiply-add to the DSL and moves
`programming_examples/primitives/vector_examples/vector_fma` onto it.

## Why the DSL needed a fused spelling

`vector_fma` and `vector_muladd` are a matched pair: both compute
`alpha * b + c`, and they exist so the fused and unfused forms can be
compared. Converting `vector_fma` as `alpha * b[:] + c[:]` would have made it
byte-for-byte identical to its sibling and silently deleted the comparison —
`vector_muladd`'s own docstring already says "vector_fma is the fused half of
this pair". So this is a case where the example needed new surface rather than
a workaround. `python/test/api/fma.py` holds both forms side by side so the
distinction is pinned rather than described in a comment.

`ops.fma` is a new ternary node kind — the second after `select`, and the
first whose arguments are all value-typed — lowering to `vector.fma`.

## Constraints, all measured by compiling

Two of the three turned out narrower than expected, and one contradicted what
I first wrote in the docstring.

* **There is no scalar fma on AIE2.** `math.fma` reaches the backend and fails
  to legalize (`unable to legalize instruction: (s16) = G_FMA`) on npu1 and
  npu2, bf16 and f32, with and without bf16 emulation — five for five. The
  emitter therefore refuses its own scalar fallback instead of emitting an op
  that cannot build. That fallback is normally the safe direction; for fma, as
  for `math.tanh` before it, it is the unsafe one. Unlike tanh this is caught
  where the offending tile shape can be named.
* **A native f32 `vector.fma` is explicitly marked illegal** by the aievec
  conversion. So this does *not* rescue f32 from the chained-multiply-add
  limitation the axpy conversion documents — f32 has no vectorised
  multiply-add by any spelling. It works only under `bf16_emulation=True`,
  which is exactly the mode the predecessor's fifth lit already exercised.
* **Integer fma is refused permanently, not pending.** Integer multiply-add is
  exact, so there is no intermediate rounding to avoid and no `arith.fmai` to
  lower to. The message says so rather than implying one might arrive.

## axpy

Updated in the same commit because it is where the gap was recorded: its
docstring listed "emits arith.mulf + arith.addf where the predecessor
hand-built a vector.fma" as a deliberate difference, which is no longer
necessary for vectorised bf16. The spelling is conditional on the emitter's
own vectorisation test, since the scalar and f32 paths still have nothing to
fuse into.

Also adds `equal`/`not_equal`/`select` to `ops.__all__`, which were omitted
when the comparison operators landed.

## Verification

* `vector_fma`'s two `ryzen_ai` lits pass on npu1 hardware, including the
  bf16-emulation one; the three npu2-only configurations compile clean.
* `axpy` passes on npu1, and `vector_muladd` still passes as a control.
* All 18 `python/test/api` suites green.
* The IR tests were checked against two mutations — expanding the fma back
  into a mul and an add, and disabling the per-buffer read cache — and both
  are caught, so the CHECK-NOTs are not vacuous.
* Byte-identity sweep over every air.api example on npu1 and npu2: 92
  identical, 2 differ, 2 new. The two that differ are axpy, and the diff is
  exactly the intended four-line substitution.

## Prior art

Checked against the obvious neighbours, since a fused elementwise MAC is a
recurring API question. NumPy has no `fma` (proposed in 2013, rejected as
scope creep), PyTorch has no `torch.fma` (`addcmul` is defined by allocation
behaviour rather than rounding), and TensorFlow leaves it to XLA. All three
rest on the premise that a downstream compiler will fuse `a*b+c` anyway,
which is false here — measured, `arith.mulf` feeding `arith.addf` stays two
ops through the whole pipeline. Triton, the nearest neighbour
architecturally, has exactly this split: `tl.dot(a, b, acc=...)` for the
contraction and a separate elementwise `tl.fma(x, y, z)`, which is the layout
air.api already had and this completes.

### DIFF
--- a/programming_examples/axpy/axpy.py
+++ b/programming_examples/axpy/axpy.py
@@ -7,18 +7,14 @@ Computes out = alpha * x + y over a 1-D [N] vector.
 
 The DSL emits the herd, the L1 allocations, the affine tile offsets, the DMAs,
 and the vectorised compute loop; what it hands to the backend is an ordinary AIR
-module. `alpha * x_buf[:] + y_buf[:]` is a lazy expression tree that is lowered
-once, as a single vectorised loop, when it is assigned into out_buf.
+module. The right-hand side is a lazy expression tree that is lowered once, as a
+single loop, when it is assigned into out_buf.
 
-Two differences from the raw-bindings implementation this replaces, both
-deliberate:
-
-  - It emits arith.mulf + arith.addf where the predecessor hand-built a
-    vector.fma.
-  - The predecessor pinned a 2-core herd and cycled tiles across it. Here the
-    logical tile grid (N // tile_n) is strip-mined onto whatever the target
-    provides -- 4 cores on npu1, 8 on npu2 -- with each core owning a
-    contiguous block of tiles.
+One difference from the raw-bindings implementation this replaces, deliberate:
+the predecessor pinned a 2-core herd and cycled tiles across it, while here the
+logical tile grid (N // tile_n) is strip-mined onto whatever the target provides
+-- 4 cores on npu1, 8 on npu2 -- with each core owning a contiguous block of
+tiles.
 
 f32 runs scalar, and that is not a tuning choice. A *chained* f32
 multiply-then-add on 512-bit vectors does not legalize in the AIE2 backend:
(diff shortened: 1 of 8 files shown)


erwei-xilinx:
Fixed in 0df45619 — `i32` was imported by symmetry with the sibling test files and never used. `fma.py` pins what `ops.fma` emits; the integer refusal is a negative, so it lives in `errors.py` alongside the other ones.

Swept the other seven files in the PR for the same defect with an AST check rather than assuming this was the only one — none.

PR #1889, opened 2026-08-25
[python] air.api: reductions, and the two vector_examples converted

erwei-xilinx:
Adds `ops.reduce_add` and `ops.reduce_max`, and moves
`programming_examples/primitives/vector_examples/vector_reduce_{add,max}` onto
them.

## The first shape-changing node

Every other node in the expression language is elementwise, which is what lets
the emitter check each leaf's shape against the destination. A reduction
collapses the innermost axis, so it is dispatched *before* that check rather
than being taught to it, and it has to be the whole right-hand side — an
elementwise op surrounding it would be operating on two different shapes.

Its *operand* may still be any elementwise expression, so a row-wise dot
product needs no extra surface:

```python
out[:] = air.ops.reduce_add(a[:] * b[:])
```

## No accumulator, deliberately

The reduced axis is read as a single vector of its full extent rather than
stepped in vector-width chunks. That is what the hand-written kernels do, and
the reason matters: stepping would need a loop-carried vector accumulator,
which is exactly the construct `ops.dot`'s docstring already documents as
unlegalizable on AIE2 (LLVM splits it into sub-512-bit pieces). The cost is
that the reduced extent has to be a vector length the backend accepts — 16 and
32 are both exercised here, being what the two examples default to.

## Both keepdims spellings

`[.., n] -> [.., 1]` and `[.., n] -> [..]` are both accepted, because the
predecessor uses both **at once**: its L1 tile is `[tile_m, 1]` while its L3
output is `[m]`. Whichever the DSL refused would have forced the example to
change shape somewhere.

The converted examples drop the axis, which keeps the L1 tile the same rank as
the L3 output. Worth stating plainly: `ops.store` squeezes *leading* unit
dimensions but not trailing ones, so a kept-axis `[tile_m, 1]` tile cannot
currently be stored into a rank-1 `[m]` slice — the DMA the predecessor writes.
`api/reduce.py` pins both spellings and records that limitation at the point
where someone would meet it. Fixing it is a trailing-unit squeeze in
`_check_pair`, a separate change.

## Kinds

`maximumf` on floats — matching the elementwise `ops.maximum`, and the form
known to legalize on AIE2 — and `maxsi` on integers.

## Verification

* **npu1 hardware**: all three `ryzen_ai` lits across the two examples pass,
  matching the baseline measured *before* any change. `vector_add` and
  `vector_max` still pass as controls.
* Both examples compile clean for npu2.
* 21/21 `python/test/api` suites; `black` clean.
* The new IR tests were checked against three mutations — collapsing
  `reduce_max`'s kind onto `add`, fixing the vector length instead of taking it
  from the reduced axis, and dropping the destination-shape check — and all
  three are caught, so the CHECKs are not vacuous.

## A note on scoping

These two examples were previously assessed as *not* a single-feature batch, on
the grounds that they also need `collapse_shape` and `exp`. Neither holds:
`exp` is a dead import with zero call sites in both files, and
`collapse_shape`/`subview` are only how the raw code manufactures a rank-1 row
— air.api reads at an offset directly and needs neither. The assessment had
been made from the import lines rather than the call sites.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_reduce_add/vector_reduce_add.py
+++ b/programming_examples/primitives/vector_examples/vector_reduce_add/vector_reduce_add.py
@@ -1,172 +1,94 @@
-# Copyright (C) 2025, Advanced Micro Devices, Inc.
+# Copyright (C) 2026, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
-from ml_dtypes import bfloat16
+"""Row-wise sum reduction, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, load, store, subview, collapse_shape
-from air.dialects.vector import (
-    transfer_read,
-    transfer_write,
-    reduction,
-    extract,
-    CombiningKind,
-    broadcast,
-)
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.dialects.math import exp
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+    out[m] = np.sum(a[m, :])
+
+The predecessor hand-rolled the row loop: an ``scf.for`` over the tile, two
+``memref.subview``s and two ``memref.collapse_shape``s per trip to turn a
+``[1, n]`` slice into a rank-1 ``memref<n>``, a ``vector.transfer_read`` with an
+explicit identity map and padding constant, ``vector.reduction``, and a scalar
+store. Here that is ``air.ops.reduce_add``: the emitter walks the rows, reads each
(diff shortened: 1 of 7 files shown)


erwei-xilinx:
Both fixed in 207894c3 — same defect twice, and correct. The docstrings were written before the drop-the-axis spelling was added (that happened mid-implementation, when a probe hit the `ops.store` limitation) and were never revisited.

The `ops.py` one was worse than reported. Its single worked example allocated `[tile_m, 1]`, which is exactly the pairing that does **not** work: `ops.store` squeezes leading unit dimensions but not trailing ones, so a kept-axis tile cannot be stored into a rank-1 `[m]` slice — the DMA the hand-written kernel writes. The one example a reader would copy steered them into the unsupported case. It now shows both spellings and says which to prefer and why.

Since this is the same class of defect I've hit twice recently, I grepped for others rather than fixing only what was flagged, and found a third: `ops.py`'s module docstring still listed `reduce` among the API-proposal ops that "are not implemented". Fixed in the same commit.

Docstrings only; no functional change; 21/21 api suites still pass.

PR #1893, opened 2026-08-25
[python] air.api: the scalar_examples family, and the shift surface it needed

erwei-xilinx:
Converts all three of `programming_examples/primitives/scalar_examples` onto `air.api`, and adds the two pieces of surface the third one needed.

Each of these examples is one line of compute next to a `vector=0`, and that `vector=0` is what they are *for*: it sends the emitter down its scalar path, which is the same `scf.for` over the tile in steps of one, `memref.load`, one arith/math op, `memref.store` that the raw-bindings predecessors wrote by hand. The emitted compute is unchanged in all three.

| example | compute | lit |
|---|---|---|
| `scalar_reciprocal` | `c[:] = 1.0 / a[:]` | `ryzen_ai` — **passes on npu1 hardware** |
| `scalar_invsqrt` | `c[:] = ops.rsqrt(a[:])` | `ryzen_ai_npu2` |
| `scalar_shift_saturate` | `ops.cast(ops.cast(clamp(a[:] >> n, -128, 127), i8), i32)` | `ryzen_ai_npu2` |

The first two need no new surface — `ops.rsqrt` and the scalar path both landed in #1890.

## New surface

**`<<` and `>>`.** New `_INT_OPS` entries (`arith.shli` / `arith.shrsi`) and the matching dunders, integer-only and refused by name on a float buffer exactly as the bitwise operators beside them are.

`>>` is arithmetic, and there is no signedness choice being made here rather than a convention being picked: `arith` requires signless operands, so an unsigned buffer is refused before it reaches *any* operator, and every buffer that can reach a shift is signed. That is also what Python does — `-8 >> 1` is `-4` in both.

The shift *count* is where the resemblance to Python stops, silently, so it is checked:

| | Python | MLIR |
|---|---|---|
| `x >> 100` (i32) | `0` / `-1` | poison |
| `x << 40` (i32) | widens | poison |
| `x >> -1` | `ValueError` | poison |

Poison is not an error — it is a value the optimiser may assume never occurs, so it surfaces as a wrong answer several passes later rather than as a diagnostic. A **constant** count out of range is now refused where it is written, which is the closest this gets to Python's own `ValueError`; the boundaries are per-dtype (i32 allows 31, refuses 32; i8 allows 7, refuses 8). A count computed at runtime — `1 << a[:]` — has no constant to inspect, still reaches the backend as poison if out of range, and is documented rather than guarded.

**A narrowing int → int cast, when the operand is provably in range.** `ops.cast` has refused i32 → i8 since #1885 on measured evidence: the vectorised `arith.trunci` saturates while the scalar one wraps, and the emitter picks between them by tile shape, so the same source would compute two different things depending on a tile size. That refusal ends by noting in-range values agree and nothing can check it.

`_clamped_into` is the case where something can. When the operand is `min(max(x, lo), hi)` with **constant** bounds inside the target's range, no value the two paths disagree about can reach the cast, both agree, and the result is a property of the program again. Either clamp order is accepted, since clamping is commutative. Everything else still raises — an unclamped narrow, a clamp to bounds wider than the target, and a single-sided clamp are each pinned as still-refused in `errors.py`.

This is a structural check on the expression tree, not a promise from the caller. The clamp has to be there.

## Why `scalar_shift_saturate` needs the op order preserved

It exists to check that mlir-aie's `LowerScalarShiftClampTruncToSRS` pattern (mlir-aie #2894) still matches a *decomposed* `shrsi + maxsi + minsi + trunci + extsi` chain and fuses it into a hardware SRS. So the conversion has to emit those five ops in that order or the example stops testing anything.

The kernel never rounds; the reference does (`(x + (1 << (shift-1))) >> shift`). They agree only because `AIECoreToStandard` sets rounding mode 9 (positive_inf) for integer SRS, so the rounding comes from the fused instruction and not from the IR. That is why this test checks at `rtol=0, atol=0`, and why a missed pattern match would show up as an off-by-one rather than as a build failure.

## Gates

- `scalar_reciprocal`'s lit **passes on npu1 hardware**, and passes non-vacuously: corrupting the reference to `2.0/x` fails with `actual` exactly half of `expected`, on both the predecessor and the conversion.
- The two npu2-only examples cannot run on a Phoenix box, so they are gated by compilation — and for `scalar_shift_saturate` that is gated hard: the generated **core LLVM IR is byte-identical** to the predecessor's, with the same three `@llvm.aie2p.I256.v32.acc32.srs` calls. The pattern demonstrably fires and the numerics cannot differ. The arith chain matches op for op at the air level too.
- **Blast radius: `-p` output for all 45 buildable `air.api` examples is byte-identical before and after.** Re-measured after the shift-count check landed in `_binary`, which is on the path of every operator.
- `python/test/api` 24/24, including a new `shift.py` pinning the operators, the scalar path, and the SRS chain order — worth a PR-gated test of its own, since the example that depends on it only runs on npu2.

## Differences from the predecessors, shared by all three

* The herd is `[NUM_TILES, 1]` rather than `[1, NUM_TILES]`. A 1-D `air.api` herd is laid out along x, the orientation that places on both generations.
* The strip-mine is the DSL's, so each core gets a contiguous run of tiles where the predecessors' hand-built `AffineMap` interleaved them. Every tile is still computed exactly once by exactly one core, and the tiles are independent.
* The L1 buffers now allocate and free together. **The predecessors allocated both outside the temporal loop and called `DeallocOp` on them inside it**, freeing each buffer once per trip — 32 times at the default size.

`scalar_invsqrt` additionally refuses npu1 with a message instead of building a module that dies later in the backend: `math.rsqrt` on a plain f32 is an instruction on npu2, and on npu1 lowers to a call into `getRsqrtBf16`, which has no f32 scalar form. Its lit was already `REQUIRES: ryzen_ai_npu2` for that reason.

## Note

After conversion these are close to their `vector_examples` siblings with one flag changed — same size, same seed, same tolerance. They are kept separate because they are separately tracked examples with their own lits, but the duplication is now visible in a way it was not before.

### DIFF
--- a/python/air/api/_value.py
+++ b/python/air/api/_value.py
@@ -539,6 +539,44 @@ _OP_SYMBOLS = {
 _OP_CALL_NAMES = {"max": "maximum", "min": "minimum"}
 
 
+def _check_shift_amount(amount, value, op):
+    """Reject a constant shift count that MLIR would turn into poison.
+
+    ``>>`` matches Python on the part people actually reason about -- it is
+    arithmetic, so a negative value floors rather than filling with zeros --
+    but it does *not* match Python on the shift count, and the difference is
+    silent. Python's ints are arbitrary precision: ``x >> 100`` is 0 or -1, and
+    ``x >> -1`` raises. MLIR inherits LLVM's rule instead, where a shift of the
+    operand's own width or more, or by a negative amount, is **poison**, and
+    poison is not an error -- it is a value the optimiser may assume never
+    happens, which is how it turns into a wrong answer several passes later
+    rather than a diagnostic.
+
+    So a constant out-of-range count is refused here, at the call site, which
+    is the closest this can get to Python's own ValueError. A count that is not
+    a compile-time constant cannot be checked and is documented rather than
+    guarded -- see the shift operators on ``BufferExpr``.
+    """
+    if amount.kind != "scalar" or not isinstance(amount.scalar, (int, bool)):
+        return  # runtime amount: nothing to check, see the docstring
+    count = int(amount.scalar)
+    dtype = value.element_dtype()
+    spelling = "<<" if op == "shl" else ">>"
+    if count < 0:
+        raise ValueError(
+            f"negative shift count {count} in '{spelling}': Python raises here "
+            f"and MLIR would make it poison, which is silent. Shift by a "
+            f"non-negative amount, or use the opposite operator"
+        )
+    if dtype is not None and count >= dtype.itemsize * 8:
+        raise ValueError(
+            f"shift count {count} is not less than the width of {dtype} "
+            f"({dtype.itemsize * 8} bits) in '{spelling}': Python would give "
+            f"{'0 or -1' if op == 'shr' else 'a wider integer'} but MLIR makes "
+            f"it poison, which is silent. Shift by less than the width"
+        )
+
+
 class BufferExpr:
     """A lazy elementwise expression over buffers and scalars.
 
@@ -641,6 +679,8 @@ class BufferExpr:
     def _binary(self, other, op, reverse=False):
         other = BufferExpr.coerce(other)
         args = (other, self) if reverse else (self, other)
+        if op in ("shl", "shr"):
+            _check_shift_amount(args[1], args[0], op)
         return BufferExpr("binary", op=op, args=args)
 
     def __add__(self, o):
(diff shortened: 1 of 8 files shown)


python/air/api/_value.py
```diff
@@ -351,6 +351,57 @@ def __repr__(self):
...
+    rather than a diagnostic.
+
+    So a constant out-of-range count is refused here, at the call site, which
+    is the closest this can get to Python's own ValueError. A count that is not
+    a compile-time constant cannot be checked and is documented rather than
+    guarded -- see the shift operators on ``BufferExpr``.
+
+    "Constant" has to include an IndexExpr that folds to one. Index arithmetic
+    over herd coordinates is ordinary in a kernel body, and ``tx - tx + 32``
+    reaches the emitter as a literal ``arith.constant 32`` -- indistinguishable,
```
erwei-xilinx, line 562: Good catch, and it reproduced: `a[:] >> (tx - tx + 32)` was not refused and reached MLIR as a literal `arith.constant 32 : i32` feeding a shrsi on i32 — by then indistinguishable from having been written as `32`, and just as much poison.

Fixed in e1d87495 by reading the count through `as_const()`. This is the same trap as the offset comparison in #1892: IndexExpr implements neither equality nor int conversion against a Python int, so an isinstance test alone classifies every index expression as runtime and lets the folded ones past.

Boundaries verified per dtype — folded 32 on i32, folded -1, and folded 8 on i8 all refuse; folded 31 on i32, a genuinely runtime `tx`, and a plain `4` all still pass. Pinned as `shift_by_a_folded_index_expression` in errors.py. A genuinely runtime count remains unchecked, which is the documented hole and not something as_const() changes.

PR #1895, opened 2026-08-25
[test] Migrate lit configs to the internal shell for lit 23

erwei-xilinx:
`lit 23.1.0` reached PyPI on **2026-08-25T19:24:31**, and `utils/requirements_dev.txt` does not pin `lit`, so CI installs it. The previous release was 18.1.8 from June 2024 — five major versions in one `pip install`. It removes two things this repo relies on, and the first takes the whole suite down before a single test runs.

This is currently failing every PR and every push to `main`. PRs showing green are showing runs from before 19:24.

## What broke

**1. `ShTest(execute_external=True)` now raises at config-parse time.**

```python
if execute_external and not force_execute_external:
    raise ValueError("execute_external=True is deprected as of LLVM-23 ...")
```

All four `lit.cfg.py` files used the old LLVM idiom `ShTest(not llvm_config.use_lit_shell)`. Despite the name, `use_lit_shell` means "use lit's *internal* shell" and is `False` everywhere except win32 — so on Linux it evaluates to `ShTest(True)`, exactly the removed option. `check-air-mlir` dies while parsing the config.

**2. `%T` was dropped from lit's substitution table.**

`getDefaultSubstitutions` lost its `("T", tmpDir)` entry, so `%T` is passed through literally and commands receive a path named `%T/...`. This is not visible from the error in (1); it only appears once (1) is fixed. `split_devices.mlir` fails with an `air-opt` assertion in `raw_fd_ostream`.

## Approach

Migrating, rather than `force_execute_external=True` — an explicit stay of execution that LLVM-24 removes anyway — and rather than pinning `lit<23`, which defers the same work. **Every change here works under both old and new lit, so no pin is needed in either direction.**

* **`ShTest()`** in all four configs: the internal shell, unconditionally.

* **`%ld_lib_path` gains a leading `env`.** It is used as a command prefix (`// RUN: %ld_lib_path %t.test.exe`). A bare `VAR=value` prefix is a shell feature — bash reads it as an environment assignment; lit's internal shell takes it as the name of the program to run and fails with `command not found`. The internal shell implements the `env` builtin (`TestRunner.py`) and `env` is also a real binary, so this spelling works under either shell.

  This was the only bare env prefix in the repo. The ~500 `PEANO_INSTALL_DIR=` / `AIE_TARGET=` occurrences are `make VAR=value` *arguments*, which sit after the command name and are unaffected.

* **`%T` → `%t`, at 198 sites.** `%t` is a unique per-test path base, so `%T/x` becomes `%t.x` and needs no directory. `%T` was per-*directory* where `%t` is per-*test*, which is safe here because every directory involved contains exactly one test — checked, not assumed. The two sites where a pass writes into a directory it does not create keep a real one, `%t.d`, with an explicit `rm -rf` + `mkdir -p`: `split_devices.mlir` had been relying on lit having created `%T` for it.

## Verification — parity in both directions

| suite | lit 18.1.8 | lit 23.1.0 | pre-change baseline |
|---|---|---|---|
| `mlir/test` | 513 pass / 1 fail | **513 pass / 1 fail** | 513 / 1 |
| `python/test` | 33 pass / 11 unsupported | 33 / 11 | 33 / 11 |
| `programming_examples` | passes on npu1 hardware | passes on npu1 hardware | passes |

lit 23.1.0 was installed in isolation (`pip install --target`) to reproduce CI exactly; the config-parse failure reproduces locally on the unmodified tree and is gone after.

The single `mlir/test` failure, `Util/Runner/multi_segment_herds.mlir`, is **pre-existing** — it fails identically on the unmodified tree under the external shell, and is untouched here.

`programming_examples` exercises the parts of the internal shell most likely to differ: `cd` across RUN lines, pipes into FileCheck, `mkdir -p`, and `flock`. All work. I also confirmed the `Util/Channel` tests genuinely recompile, link and execute a fresh binary under the internal shell rather than passing trivially.

## Not verified here

`test/airhost` accounts for 146 of the 198 `%T` sites. It requires a VCK5000, is off by default (`ENABLE_RUN_AIRHOST_TESTS=OFF`), and is not PR-gated, so **I could not run it**. It is migrated by the same mechanical rule, which I verified is safe for it (one test per directory; the same write-then-read-within-a-single-test pattern as the `mlir/test` cases). Leaving it on `%T` would have left it broken for whoever next enables it, but this part is unverified and reviewers should treat it as such.

## Follow-up worth considering separately

`utils/requirements_dev.txt` lists `lit` unpinned. That is what turned an upstream release into an unannounced repo-wide breakage overnight. Worth pinning or bounding regardless of this PR.

### DIFF
--- a/test/airhost/42_air_add_one_place_random/run.lit
+++ b/test/airhost/42_air_add_one_place_random/run.lit
@@ -6,27 +6,32 @@
 //
 //===----------------------------------------------------------------------===//
 
-// RUN: aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a
+// The bash -c wrapper is load-bearing under lit's internal shell, which
+// performs no shell expansion at all: $(...), ${VAR} and even $VAR are
+// passed through literally. Handing the line to a real shell keeps the
+// expansion, and lit's own %s/%S/%t substitutions still apply inside the
+// quotes because lit substitutes textually before anything is executed.
+// RUN: bash -c "aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a"
 // RUN: cat air_project/aiecc.graph_0.mlir | grep tile | head -n 1
 // RUN: %CLANG %S/test.cpp -I%HSA_DIR%/include -L%HSA_DIR%/lib -lhsa-runtime64 -I%LIBXAIE_DIR%/include -L%LIBXAIE_DIR%/lib -lxaiengine -I%AIE_RUNTIME_DIR%/test_lib/include -L%AIE_RUNTIME_DIR%/test_lib/lib -ltest_lib %airhost_libs% -Wl,--whole-archive %t.air.a -Wl,--no-whole-archive -rdynamic -o %t.test.elf
 // RUN: %run_on_board %t.test.elf
 
-// RUN: aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a
+// RUN: bash -c "aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a"
 // RUN: cat air_project/aiecc.graph_0.mlir | grep tile | head -n 1
 // RUN: %CLANG %S/test.cpp -I%HSA_DIR%/include -L%HSA_DIR%/lib -lhsa-runtime64 -I%LIBXAIE_DIR%/include -L%LIBXAIE_DIR%/lib -lxaiengine -I%AIE_RUNTIME_DIR%/test_lib/include -L%AIE_RUNTIME_DIR%/test_lib/lib -ltest_lib %airhost_libs% -Wl,--whole-archive %t.air.a -Wl,--no-whole-archive -rdynamic -o %t.test.elf
 // RUN: %run_on_board %t.test.elf
 
-// RUN: aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a
+// RUN: bash -c "aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a"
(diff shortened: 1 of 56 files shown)


test/airhost/42_air_add_one_place_random/run.lit
```diff
@@ -6,27 +6,32 @@
 //
 //===----------------------------------------------------------------------===//
 
-// RUN: aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %T/air.a
+// The bash -c wrapper is load-bearing under lit's internal shell, which
+// performs no shell expansion at all: $(...), ${VAR} and even $VAR are
+// passed through literally. Handing the line to a real shell keeps the
+// expansion, and lit's own %s/%S/%t substitutions still apply inside the
+// quotes because lit substitutes textually before anything is executed.
+// RUN: bash -c "aircc -row-offset=$(shuf -i 1-8 -n 1) -col-offset=$(shuf -i 0-49 -n 1) %S/air.mlir -o %t.air.a"
```
erwei-xilinx, line 14: Confirmed, and it is broader than reported — thank you. A probe under lit 23.1.0 shows the internal shell performs **no** shell expansion at all: `$(echo HELLO)`, `${CARDANO}` and even a plain `$CARDANO` are each passed through literally as text.

Scanned every suite. There are exactly 7 such RUN lines, all in `test/airhost`:

| count | expansion | file |
|---|---|---|
| 5 | `$(shuf ...)` | 42_air_add_one_place_random |
| 1 | `${CARDANO}` | 50_air_mmult_2x2_extern |
| 1 | `${CARDANO}` | 26_air_beefmaker_extfn |

`mlir/test`, `python/test`, `programming_examples` and `test/xrt` have none — so those suites were genuinely clean rather than lucky, which matters because they are where this PR's parity numbers come from.

Fixed in e0e90d0f by wrapping each line in `bash -c`, per your first suggestion. I kept the randomness rather than substituting fixed offsets: the test is called `place_random` and the random placement is what it exercises.

Verified by reproducing the two exact wrapped RUN lines under lit 23.1.0's internal shell, with `echo` standing in for aircc/xchesscc:

```
xchesscc -p me -P /opt/cardano/data/cervino/lib -c /tmp/.../chess/extern_kernel.cc
aircc -row-offset=7 -col-offset=28 /tmp/.../air.mlir -o /tmp/.../exact42.txt.tmp.air.a
```

`${CARDANO}` expands from the environment, both `shuf` calls evaluate within their ranges, and lit's own `%S`/`%t` still substitute inside the quotes — lit substitutes textually before anything executes.

As with the rest of `test/airhost`, this verifies the shell mechanics rather than the tests themselves: they need a VCK5000 and are off by default.

PR #1896, opened 2026-08-25
[python] air.api: squeeze trailing unit axes in transfers, and convert average_pool

erwei-xilinx:
`ops.store` already ignored a *leading* unit axis, because a per-core L2 staging tile is spelled `staged[tx, j:j+m, :]` and lands in an `[m, k]` L1 buffer. This does the same for a trailing one, for the symmetric reason: `ops.reduce_add` collapses the innermost dimension, so reducing an `[m, n]` tile gives an `[m, 1]` buffer — a column of m values — and the L3 array it belongs in is `[m]`. Both spellings describe the same m contiguous elements.

Only *unit* axes are dropped, and only from the ends. Nothing here can turn `[m, n]` into `[n, m]` or excuse a differing extent: a trailing axis of 4 against a trailing axis of 1 still fails, and is pinned as such in `errors.py`.

## The example

This was the only thing standing between `average_pool` and air.api. The compute needed no new surface:

```python
c[:] = air.ops.reduce_add(a[:] * (1.0 / n))
```

One line, against a per-row `scf.for` containing two `memref.subview`s, two `memref.collapse_shape`s with an explicit dynamic `StridedLayoutAttr`, a `vector.transfer_read` with an identity permutation map and a padding constant, a `vector.broadcast`, an `arith.mulf` and a `vector.reduction`. Op counts are identical to the predecessor's.

**The scale goes inside the reduce, and that is load-bearing.** Reducing first and scaling the result is one *scalar* bf16 multiply per row, and the predecessor carries a comment that a scalar bf16 multiply "can produce corrupted output on AIE2". Keeping the multiply inside puts it on the vector, matching the predecessor's IR and its reference — which also scales per element before summing, an order that differs from scale-after-sum in bf16 regardless of the AIE2 hazard.

## A comment that was documenting the limitation

`python/test/api/reduce.py` carried this:

> Note the L3 output here is [M, 1], not [M]. ops.store squeezes *leading* unit dimensions but not trailing ones, so a [tm, 1] L1 tile cannot currently be stored into a rank-1 [m] slice -- which is exactly what the hand-written reduce kernel does in its DMA. That is why the converted examples drop the axis instead.

That is now false, so it is replaced by a test exercising the pairing rather than a comment explaining the workaround.

## Gates

- **`average_pool` passes on npu1 hardware**, and non-vacuously: doubling the reference fails, and so does a 20% perturbation, which is just outside its `rtol=1e-1`.
  - Worth recording: a `+5.0` perturbation *passes*. That is the control being too weak rather than the test being loose — the row averages run to ~1e6, so +5.0 is far inside a 10% tolerance. The proportional controls are the meaningful ones.
- Op-count A/B against the predecessor: identical.
- `python/test/api` 23/23.
- **Blast radius: `-p` output for all 45 buildable air.api examples is byte-identical**, measured against a baseline rebuilt from current `origin/main`'s DSL after rebasing past #1892.

## Differences from the predecessor

* The herd is `[NUM_TILES, 1]` rather than `[1, NUM_TILES]`. A 1-D air.api herd is laid out along x, the orientation that places on both generations.
* The strip-mine is the DSL's, so each core gets a contiguous run of tiles where the predecessor's hand-built `AffineMap` interleaved them. Every tile is still computed exactly once by exactly one core, and the rows are independent.
* `--target` is new and defaults to detecting the installed part, which is what the predecessor did implicitly by having no device flag at all.

## Note on ordering

Rebased onto `main` past #1892, which touched the same `_check_pair` function; the code merged cleanly and only the docstring needed reconciling, so it now describes both the view case (#1892) and the unit-axis case (this PR).

### DIFF
--- a/programming_examples/average_pool/average_pool.py
+++ b/programming_examples/average_pool/average_pool.py
@@ -1,184 +1,100 @@
 # Copyright (C) 2026, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
+"""Average pool over the rows of an [M, N] array, on air.api.
 
-"""Vectorized AveragePool Example
+    c[:] = air.ops.reduce_add(a[:] * (1.0 / n))
 
-Implements 1D average pooling on a 2D input [M, N]:
-  output[i] = mean(input[i, :]) for each row i
+One line of compute. ``ops.reduce_add`` collapses the innermost dimension, so an
+``[tile_m, n]`` tile reduces to an ``[tile_m, 1]`` buffer -- a column of row
+averages -- which is what the predecessor built by hand out of a per-row
+``scf.for`` containing two ``memref.subview``s, two ``memref.collapse_shape``s
+with an explicit dynamic ``StridedLayoutAttr``, a ``vector.transfer_read`` with
+an identity permutation map and a padding constant, a ``vector.broadcast``, an
+``arith.mulf`` and a ``vector.reduction``.
 
-Each row of N elements is scaled by 1/N (vectorized multiply) and then
-reduced to a single scalar using vector.reduction with ADD.
+The emitted arithmetic is unchanged: broadcast 1/n to a vector, multiply, then
+``vector.reduction <add>``.
 
-Uses a 1x2 AIE herd with DMA transfers between L3 and L1 memory.
+**The scale happens before the reduction, and that is load-bearing.** Writing it
+the other way round -- reduce, then scale the result -- is one scalar bf16
+multiply per row instead of one vector multiply per row, and the predecessor
+carries a comment saying a scalar bf16 multiply "can produce corrupted output on
+AIE2". Putting the multiply inside the reduce keeps it on the vector, matching
+both the predecessor's IR and its reference, which also scales each element
+before summing. In bf16 those two orders do not agree to the last bit even in
+exact arithmetic, so this is not only about the AIE2 hazard.
+
+The ``[tile_m, 1]`` shape of the output tile is what a reduction produces, while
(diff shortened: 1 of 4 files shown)


programming_examples/average_pool/average_pool.py
```diff
@@ -1,184 +1,107 @@
...
-from air.backend.xrt_runner import XRTRunner, type_mapper
+from air import api as air
+from air.api.types import bf16
 from air.backend.xrt import XRTBackend
+from air.backend.xrt_runner import XRTRunner
 
-import numpy as np
-
+# The rows checked against the reference are drawn at random, so the seed is
+# what makes a failure reproducible: without it a mismatch reports row indices
```
erwei-xilinx, line 56: Good catch — this is a real regression I introduced, not a style point. The predecessor set `np.random.seed(42)` at module level and my rewrite dropped it.

It matters most for debugging rather than pass rates: a mismatch reports the row indices it failed on, and without a seed the *next* run does not look at those rows — so a failure cannot be reproduced from its own output.

Restored in 72c44485, at import and before any draw, which is where the predecessor and the rest of this directory (`vector_add`, `vector_reduce_add`) set it.

Verified it reproduces the predecessor's **exact** indices rather than merely being stable — seeding at the same point relative to the first `randint` gives the identical 100-element draw:

```
indices identical to predecessor: True
first 8: [56422 15795   860 38158 62570 54343 44732 11284]
stable across separate processes: True
```

The emitted IR is byte-identical before and after: the seed reaches verification only, never the kernel. Hardware lit still passes, `python/test/api` 23/23.

PR #1897, opened 2026-08-25
[air-runner] Replace path enumeration in hasPath with reachability

erwei-xilinx:
`hasPath` enumerated every path from start to end and concatenated their vertex lists, with no visited set and no memoisation. That is exponential in *reconvergent branching* rather than in graph size.

A wide diamond — one token fanning out to N ops and rejoining at a barrier — multiplies the path count by N. A model built from L such stages has N^L paths. At N=12 that is 20k paths for four stages, 4×10⁸ for eight, and 10¹² for twelve, at which point the vector requested 4.5 billion entries and LLVM aborted:

```
LLVM ERROR: SmallVector unable to grow. Requested capacity (4495263186)
is larger than maximum value for size type (4294967295)
 #13 runnerNode::hasPath(...)
 #14 runnerNode::hasPath(...)
 #15 runnerNode::hasPath(...)
 #17 runnerNode::resetGraphBetweenTwoVertices(...)
 #18 AIRRunner_impl::scheduleLaunch(...)
```

Graph size was not the problem: the same op count in a shape without reconvergent async dependencies simulates fine, because dependency edges are what build the diamonds. It appears once ops fanning out from a common token also rejoin — which is what any pipelined or barrier-synchronised model does.

### Fix

The sole caller only iterates the result to reset vertices, which is idempotent and order-independent, so the *set* is all that is needed. The vertices on some path from a to b are exactly those reachable from a that also reach b: one forward traversal over `adjacentVertices`, one backward over `inverseAdjacentVertices`, and their intersection. Linear in the graph.

### Testing

`check-air-mlir` passes with no failures. A 432-segment model that previously simulated reports an identical latency before and after, so this is behaviour-preserving wherever the old code terminated.

No new lit test: reproducing the abort needs on the order of 10⁹ paths, which is a several-thousand-line input, and any smaller case passes under both implementations because the old one merely gets slow rather than failing. The existing runner tests cover the behaviour that had to stay the same.

### DIFF
--- a/mlir/lib/Util/Runner/RunnerNode.cpp
+++ b/mlir/lib/Util/Runner/RunnerNode.cpp
@@ -1303,18 +1303,24 @@ private:
   // also reach b, which is two linear traversals.
   bool hasPath(Graph::VertexId start_v, Graph::VertexId end_v, Graph &G,
                SmallVector<Graph::VertexId, 1> &vec) {
-    std::set<Graph::VertexId> forward, backward;
+    llvm::DenseSet<Graph::VertexId> forward, backward;
     SmallVector<Graph::VertexId> worklist;
+    // Forward-reachable vertices in traversal order. The result is built by
+    // filtering this rather than by iterating the set, so the order does not
+    // depend on the hash function: the caller may push reset vertices onto the
+    // wavefront candidate list, and that order reaches the trace.
+    SmallVector<Graph::VertexId> reached;
 
     forward.insert(start_v);
     worklist.push_back(start_v);
     while (!worklist.empty()) {
       auto v = worklist.pop_back_val();
+      reached.push_back(v);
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Good catch — fixed in the follow-up commit.

Switched `forward`/`backward` to `llvm::DenseSet`. One wrinkle worth noting: I build the result by filtering the **forward traversal order** rather than by iterating the set, because iteration order over a `DenseSet` depends on the hash function, and `resetGraphBetweenTwoVertices` can push the reset vertices onto the wavefront candidate list — where the order reaches the emitted trace. Filtering a `SmallVector` recorded during the forward walk keeps that deterministic while still getting O(1) membership.

`check-air-mlir` unchanged at 510 passed / 7 xfail / no failures.

PR #1898, opened 2026-08-25
[air-runner] Diagnose a stalled simulation instead of reporting a time

erwei-xilinx:
The scheduling loop stops as soon as no runner node can make progress. That is the normal end of a simulation, but it is also exactly what a stall looks like: an op that can never satisfy its dispatch condition — a `channel.get` whose matching put count is unreachable, say — simply stops being retried, and every op behind it is abandoned. The runner then printed whatever time it had reached and exited 0.

That is the expensive failure mode, because the output says nothing is wrong. A model that lost most of its work to an unsatisfiable pairing reported `0.025us` against a true `10.5us`, and the only way to notice was to count events in the trace and find the work missing.

### Fix

Check that the launch terminator was actually reached, and report a stall if not.

Runs that stop because a hierarchy could not be allocated have already given a more specific reason and are expected not to terminate, so a `ScopedDiagnosticHandler` notes whether anything was reported and keeps this quiet when something was — which is what `Util/Runner/bad_launch` asserts.

`runner_assertion` moves to the public section so the driver reports through the same path.

### Scope

This is a backstop, not a deadlock detector. It catches stalls on the terminator's dependency path, which is where they matter, but an abandoned op whose token nothing consumes still passes unnoticed.

### Testing

`check-air-mlir`: 511 passed, 7 xfail, no failures. The new test builds a put/get spatial-factor mismatch that can never pair, places it on the terminator's dependency path, and checks for the diagnostic. It reports a plausible latency and exits 0 without this change.

### DIFF
--- a/mlir/lib/Util/Runner.cpp
+++ b/mlir/lib/Util/Runner.cpp
@@ -438,6 +438,18 @@ public:
   void scheduleLaunch(runnerNode &launch, device &device_resource_node,
                       uint64_t &time) {
 
+    // Note whether anything reported an error while this launch ran, so the
+    // stall check at the end can stay quiet when a more specific diagnostic
+    // has already been given. Returning failure leaves the diagnostic for the
+    // default handler to print.
+    bool diagnosticEmitted = false;
+    mlir::ScopedDiagnosticHandler diagHandler(
+        launch.ctrl_g->hierarchyOp->getContext(), [&](mlir::Diagnostic &diag) {
+          if (diag.getSeverity() == mlir::DiagnosticSeverity::Error)
+            diagnosticEmitted = true;
+          return mlir::failure();
+        });
+
     auto start_v = launch.ctrl_g->start_vertex;
     // Reset launch graph
     launch.processed_vertices.clear();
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Pushed a follow-up: `test/airrunner/matmul_bf16_vectorized` is now `XFAIL`.

That test asserted `423.824us`, which turns out to be the point at which the simulation gave up rather than a latency. The run stalls at 26489 cycles and the test uses single-iteration mode over 16 iterations — `26489 * 16 = 423824 ns`, exactly the asserted figure. It was the stall point.

**This change does not cause the stall.** On unmodified `main` the trace already has `LaunchTerminator: 0` and `SegmentTerminator: 0` and stops at the same cycle; it just exited 0 and printed a time, so nothing showed. This PR only adds the check that says so.

`XFAIL` rather than an expectation matching the error, so the signal points the right way: matching the error would go green now and red the day someone fixes the stall. As an `XFAIL` it becomes an `XPASS`, which is the prompt to drop the marker and pin a real number. The retained check matches the latency line without its value, since the old one was the stall point and the real one isn't known.

What is actually missing: all sixteen herds terminate, the C write-back over `@channel_10` completes one get of sixteen, neither terminator is reached, and the four L3→L2 input channels hold a dispatched put with no matching get. So the number was compute time without the output drain — an underestimate, which is why it looked plausible.

Ruled out, each costing a run:
- **not scale** — 1x1 stalls at 8057, 2x2 at 14201, 4x4 at 26489; 1x1 is the smallest repro
- **not ping-pong** — dropping both ping-pong passes still stalls, at 6362
- **not granularity** — `"herd"` behaves as `"core"`
- **not resource starvation** — no op is ready-but-blocked; every unstarted op still has an incomplete predecessor, so it is an unsatisfiable dependency state

`test/airrunner/matmul`, which builds its IR through the transform dialect instead of parsing the air.api example, reaches both terminators and is unaffected.

Whether the generated IR is at fault or air-runner cannot model it is still open — I could not root-cause it, and did not want to hold this behind a separate bug it merely revealed.

PR #1899, opened 2026-08-25
[air-runner] Fix silently truncated simulation of multi-instance segments

erwei-xilinx:
An `air.segment` with an iteration space greater than one, containing a herd fed by a channel, dropped the herd's entire body. The run exited 0 with no diagnostic and reported `0.025us` where the correct answer was `10.5us` — the custom op never appeared in the trace and neither terminator was reached.

### Cause

`tokenSpatialFactorForResource` stopped its parent walk at the innermost enclosing hierarchy. A `channel.get` inside a herd therefore saw only the herd, never the segment iteration space enclosing it, while the matching `channel.put` at segment level *did* see it. The put dispatched N instances against a get expecting one, so the completion test in `executeOp(ChannelGetOp)` —

```
put_processed * bcast == total_count && get_processed == total_count
```

— could never hold. The get never retired and the herd hung with its body unexecuted.

### Fix

Walk all the way out instead, multiplying every enclosing spatial factor. `air.launch` stays excluded, as before, because launch iterations are simulated separately. A hierarchy with no iteration space contributes a factor of one, so nothing changes for IR that does not use one — which is every existing test.

The template parameter only ever terminated this walk and was always instantiated with `air::HierarchyInterface`, so it and the unused `position` argument are dropped rather than left to mislead.

Separately, `getResourceCost(air::SegmentOp)` ignored the iteration space entirely, so a 2x1 segment was billed a single DU and fit an arch that cannot hold it. All instances are co-resident for the lifetime of the segment, so their demands sum; multiply by the trip count. A 2x1 segment on a one-DU arch now reports *"isn't allocated with enough resources to run"* instead of a plausible number.

### Why it survived

The construct is well supported elsewhere in the compiler — `air-to-aie` unrolls a 2x1 segment across devices, with tests for metadata ordering, packet flow ids and `scf.if` interaction. None of those is a runner test.

### Testing

`check-air-mlir`: 511 passed, 7 xfail, no failures. The new test fails against the unfixed runner and passes against the fixed one (verified both ways on this branch).

### DIFF
--- a/mlir/lib/Util/Runner/RunnerNode.cpp
+++ b/mlir/lib/Util/Runner/RunnerNode.cpp
@@ -521,6 +521,11 @@ private:
     if (num_rows && num_cols) {
       usage_count *= llvm::divideCeil(*num_rows, du_size_x);
       usage_count *= llvm::divideCeil(*num_cols, du_size_y);
+      // All instances of a segment's iteration space are co-resident for the
+      // lifetime of the segment, so their resource demands sum rather than
+      // alias. Without this a 2x1 segment is billed one DU and fits an arch
+      // that cannot hold it.
+      usage_count *= this->canonicalizer.getTripCountInHierarchyOp(op);
       return usage_count;
     } else {
       op->emitOpError("Segment has no placed AIE cores");
@@ -802,8 +807,7 @@ private:
 
     // Check how many evnets need to be dispatched in this op
     unsigned total =
-        this->tokenSpatialFactorForResource<air::HierarchyInterface>(
-            Op.getOperation(), {});
+        this->tokenSpatialFactorForResource(Op.getOperation());
     this->allocateEventToResources(chan_interface, reserved_resources,
                                    "outbound", dispatched);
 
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Both addressed.

**Test coverage for `getResourceCost`** — correct, and a fair catch: the arch in `segment_iteration_space.mlir` is large enough that resource allocation never binds, so nothing there exercised the second defect. Added `bad_launch/bad_segment_iteration_space.mlir` as suggested, using the existing three-DU arch: four co-resident instances of a one-DU segment need four. It reports **no error at all** against the unfixed runner and the expected diagnostic with the fix, so it gates that half directly. Cross-referenced from the first test so neither header overclaims.

Worth noting I got this wrong once while verifying — the fix was already committed on the branch, so stashing the working tree only removed the clang-format edits and the "without the fix" run was still using the fixed binary. Re-checked against `origin/main`'s actual file: 0 diagnostics without, 2 with.

**clang-format** — applied; it produced exactly the five suggested rewrites.

`check-air-mlir`: 512 passed, 7 xfail, no failures.

PR #1900, opened 2026-08-25
[programming_examples] air.api: the two data_transfer_transpose variants that do not use channels

erwei-xilinx:
Converts `data_transfer_transpose/dma` and `data_transfer_transpose/dma_bf16` onto air.api. **No new DSL surface** — both use what #1892 and #1890 already landed, and no file under `python/air/api/` is touched.

## dma — the example the transpose view was built for

There is no compute in it at all. The transpose is a property of the *access pattern* the second DMA walks:

```python
air.ops.load(t, A[:, :])
air.ops.store(t.transpose(1, 0), B[:, :])
```

against a tile whose own sizes/strides are `[m, k]` / `[k, 1]`, that emits a descriptor with sizes `[k, m]` and strides `[1, k]`. The predecessor spelled the same thing by passing literal lists to `dma_memcpy_nd`:

| | sizes | strides |
|---|---|---|
| predecessor | `[1, k, m]` | `[1, 1, k]` |
| this | `[k, m]` | `[1, k]` |

Identical modulo the leading unit axis `ops.store` already squeezes. `transpose` takes a full permutation, numpy-style, rather than a bare `.T` — on an nd machine reversing every axis is rarely what is meant, and naming the permutation is what would make a rank-3 or rank-4 version of this readable.

The L1 tile is `[m, k]` rather than the predecessor's flat `[m * k]`. Same bytes, same transfers; the shape is now the shape of the thing in it, which is what lets the permutation be written at all.

## dma_bf16 — why this one needs a kernel and its sibling does not

That route is closed for bf16: **on AIE the innermost DMA stride must be 1 for data narrower than 32 bits**, so a bf16 transpose cannot be expressed as an access pattern. The matrix is DMAed into L1 contiguously and a scalar C++ kernel does the transpose in place of the descriptor — `air.extern("transpose_bf16", link_with="transpose.o")`.

So the two variants are not duplicates. They are the two halves of one fact about the hardware, and the element type selects between them. Both docstrings now say so, which the predecessors only said in one of the two.

Its buffers stay flat (`[m * k]`), because the kernel computes indices from its own `DIM_M`/`DIM_N` defines — the flatness is the contract with `transpose.cc`, not an artifact to tidy.

## channel is deliberately left alone

It is the third variant in this directory and it uses `air.channel` put/get, which air.api does not implement. Converting the other two does not strand it — it is untouched and still passes on hardware.

## Gates

**dma** — its peano lit passes on npu1 hardware, and exercises *both* element types the Makefile ships: `run_int` (uint32) and `run_float` (float32) are two separate FileCheck invocations in the same lit.

Non-vacuous: perturbing the expected matrix by 1 fails with **2028 mismatches out of 2048**, so the whole matrix is checked rather than sampled.

The chess lit is UNSUPPORTED here (no xchess license) and the elf lit is npu2-only.

**dma_bf16** — `REQUIRES: ryzen_ai_npu2`, and this was developed on a Phoenix box, so it is gated by compiling for npu2 with its aie2p kernel object built and linked exactly as the Makefile does (`compile-kernel`, then copied into `air_project/`). The predecessor was compiled the same way as a control; both produce an xclbin.

Its emitted IR matches the predecessor's op for op — same `func.func private` declaration, same `air.herd` with `link_with = "transpose.o"`, same two `air.dma_memcpy_nd` and the `func.call` between them. The only delta is the `llvm.emit_c_interface` attribute that `air.extern` stamps on every declaration it makes.

`python/test/api` 23/23. No blast-radius sweep was needed since no DSL file is touched.

### DIFF
--- a/programming_examples/data_transfer_transpose/dma/transpose.py
+++ b/programming_examples/data_transfer_transpose/dma/transpose.py
@@ -56,17 +56,20 @@ def build_module(m, k, np_dtype):
 
         @launch.body
         def _():
-            with air.segment(name="seg"):
-                with air.herd([range(1)], name="herd", shape=(1,)) as h:
+            with air.segment(name="seg") as seg:
 
-                    @h.body
-                    def _(tx):
-                        t = air.alloc([m, k], dt, scope=h.private())
-                        air.ops.load(t, A[:, :])
-                        # The whole example: the same tile, walked with its axes
-                        # swapped. A view, not a copy -- nothing moves until the
-                        # store's descriptor walks it.
-                        air.ops.store(t.transpose(1, 0), B[:, :])
+                @seg.body
+                def _():
+                    with air.herd([range(1)], name="herd", shape=(1,)) as h:
+
+                        @h.body
+                        def _(tx):
+                            t = air.alloc([m, k], dt, scope=h.private())
+                            air.ops.load(t, A[:, :])
+                            # The whole example: the same tile, walked with its
+                            # axes swapped. A view, not a copy -- nothing moves
+                            # until the store's descriptor walks it.
+                            air.ops.store(t.transpose(1, 0), B[:, :])
 
     return launch
 
--- a/programming_examples/data_transfer_transpose/dma/transpose.py
+++ b/programming_examples/data_transfer_transpose/dma/transpose.py
@@ -1,15 +1,43 @@
 # Copyright (C) 2024, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
+"""Transpose an [M, K] matrix on the way out of L1, on air.api.
+
+    air.ops.load(t, A[:, :])
+    air.ops.store(t.transpose(1, 0), B[:, :])
+
+There is no compute here at all. The transpose is a property of the *access
+pattern* the second DMA walks: the tile is read back with its two axes swapped,
+so the descriptor carries sizes [k, m] and strides [1, k] where the buffer's own
+are [m, k] and [k, 1]. The hand-written predecessor spelled that by passing
+``src_sizes=[1, k, m], src_strides=[1, 1, k]`` to ``dma_memcpy_nd`` as literal
+lists; ``transpose`` derives the identical descriptor from the permutation,
+which is the same thing said in terms of what it means rather than what it
+computes to.
+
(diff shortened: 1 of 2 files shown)


programming_examples/data_transfer_transpose/dma/transpose.py
```diff
@@ -18,60 +46,38 @@
...
-                    tensor_in = AllocOp(tensor_type, [], [])
+        @launch.body
+        def _():
+            with air.segment(name="seg") as seg:
 
-                    dma_memcpy_nd(
-                        tensor_in,
-                        a,
-                    )
+                @seg.body
```
erwei-xilinx, line 61: Exactly right, and it was worse than the comment says — the launch went too. What I shipped was:

```
func.func @transpose(...) {
  air.herd @herd ...      <- bare, no launch, no segment
}
```

against the predecessor's `air.launch` → `air.segment @seg` → `air.herd`. Entering the context manager without registering a body emits nothing at all.

Fixed in 1528077f by registering `@seg.body`, which is how every other converted example spells it. Both files.

**Two reasons this got past me, both worth naming:**

1. **The hardware test passed anyway**, on both element types. A 1×1 herd works without the enclosing ops, so correctness said nothing about structure.
2. **My IR A/B did not catch it** because I diffed only the `air.dma_memcpy_nd` lines. The descriptors were right the whole time; what was wrong was everything around them. Comparing the op hierarchy is now part of the check rather than just the ops I happened to be looking for.

After the fix:

| | hierarchy | op counts |
|---|---|---|
| dma | identical | −1 `memref.alloc`/`dealloc` pair |
| dma_bf16 | identical | identical |

That one remaining pair is a dead buffer the predecessor allocates and immediately frees without ever using (`tensor_in` is re-assigned and dealloc'd with no transfer in between) — a copy-paste leftover, dropped deliberately.

Re-gated: dma's peano lit passes on npu1 hardware across both dtypes and still fails when the expected matrix is perturbed; dma_bf16 still compiles for npu2 with its aie2p kernel linked; `python/test/api` 23/23.

Separately — this is a sharp edge in the DSL, not just in my usage: `with air.segment(...)` whose body is never registered is a **silent** no-op that produces structurally different IR which still runs. That seems worth making raise. I have not done it here since it is a DSL change and this PR touches no DSL file, but I will raise it separately.

PR #1901, opened 2026-08-25
[air-runner] Tell the entries of a channel bundle apart

erwei-xilinx:
Channel progress was keyed on the symbol alone, so every entry of a bundle shared one put/get counter. The entries are independent links, so a get on one could be satisfied by a put on another.

It shows up as soon as a pipeline is seeded and drained on the two ends of one bundle — `launch put @io[0]` … stages … `launch get @io[N]`, the natural way to write a chain when the stage index selects the link. The drain becomes ready the moment its own seed completes, long before any stage has produced anything, and consumes that seed. The first stage's get then starves and the run stalls.

### Fix

Key on the symbol plus the indices. Two details matter:

**Resolvability is decided per symbol, not per op.** If one user resolves to `@c[0]` while another on the same link cannot resolve its index, the two land in different key spaces and never pair — which stalls the run just as surely as merging them. So a bundle is index-keyed only when every user resolves.

**An index may arrive through a hierarchy boundary.** `air.segment` and `air.herd` are `IsolatedFromAbove`, so a constant defined outside reaches the body as a block argument bound to a kernel operand, and `getConstantIntValue` sees only the block argument. That is what an index looks like after a loop over segments is unrolled: the constant is real, but one binding away. Walk back through `getKernelArguments`/`getKernelOperands` before giving up.

Anything still unresolvable — a herd induction variable, say — falls back to the symbol alone, which is the previous behaviour, so existing IR is unaffected.

### How narrow the failure is

Worth recording, because it took several attempts to reproduce. A serial chain does *not* expose it: there each stage both receives and sends, so the dataflow orders the pairings and the counters never overlap. Nor do concurrent independent pairs, where the counts stay symmetric and balance out. It needs a producer and a consumer of **different entries ready at the same time**, which is exactly what a seed-and-drain pair creates and what the test reproduces in fifteen lines.

### Testing

`check-air-mlir`: 511 passed, 7 xfail, no failures. The new test stalls against the old keying and completes in 10.5 µs with the new one (verified both ways on this branch).

### DIFF
--- a/mlir/lib/Util/Runner.cpp
+++ b/mlir/lib/Util/Runner.cpp
@@ -37,6 +37,7 @@
 
 #include <algorithm>
 #include <numeric>
+#include <optional>
 #include <string>
 
 #include "./Runner/Resource.cpp"
--- a/mlir/lib/Util/Runner/RunnerNode.cpp
+++ b/mlir/lib/Util/Runner/RunnerNode.cpp
@@ -822,14 +822,14 @@ private:
     }
     bool found_entry = false;
     for (auto &entry : *channel_token_counts_ptr) {
-      if ((!found_entry) && entry.first == Op.getChanName().str()) {
+      if ((!found_entry) && entry.first == this->getChannelInstanceKey(Op)) {
         entry.second += spatial_factor;
         found_entry = true;
       }
     }
     if (!found_entry) {
       channel_token_counts_ptr->push_back(
-          std::make_pair(Op.getChanName().str(), spatial_factor));
+          std::make_pair(this->getChannelInstanceKey(Op), spatial_factor));
     }
   }
   void allocateEventToResources(air::ChannelGetOp Op,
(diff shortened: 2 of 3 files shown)


erwei-xilinx:
Both addressed.

**`<optional>`** — right, it was relying on a transitive include. Added explicitly to `Runner.cpp`, which is the translation unit that pulls in `RunnerNode.cpp`.

**Pinned latency** — agreed, and dropped. The assertion that actually matters is that the stage runs and the launch terminates; with the bundle entries merged neither happens, so the structural CHECKs carry the whole test. The numeric value would break on any unrelated cost-model change without saying anything about which put a get pairs with.

Re-verified both ways against `origin/main`'s `RunnerNode.cpp` after relaxing it — still fails without the fix, passes with it — so the check is no weaker as a gate.

`check-air-mlir`: 511 passed, 7 xfail, no failures.

PR #1902, opened 2026-08-25
[air-runner] Model fixed link latency separately from occupancy

erwei-xilinx:
A link has two costs: how long it is busy carrying a payload, and how long the payload is in flight. The runner only had the first — `getTransferCost` is bytes over bandwidth — so an interconnect whose cost is dominated by time-of-flight rather than throughput could not be expressed at all.

### Why it can't just be added to the existing cost

Resources are reserved in `pushToWavefront` and freed at `end_time`, so occupancy is exactly `modelOp`'s return value. Adding flight time there holds the link for the whole duration and caps its throughput at `1/(occupancy + latency)`. A deep pipeline then cannot fill — the opposite of how a high-latency link actually behaves.

### Approach

Give each event two timestamps:

```
release_time = start + occupancy       ports freed here   (bytes / bandwidth)
end_time     = release_time + latency  data arrives here  (time of flight)
```

`end_time` keeps its meaning, so `is_done`, dependency fulfilment and the trace are unchanged, and consumers still wait for the data. Only the *port release* moves earlier. `getTimeStampsFromWavefront` also has to offer `release_time` as a wake-up point, or the simulator sleeps straight past it.

Latency is declared per memory-space interface in the arch JSON, reusing the existing `interfaces` map, and is additive across the pair: src outbound + dst inbound.

### Backward compatibility

With no `latency` in the JSON, `release_time == end_time` and `has_link_latency()` is false, so the early-release path is unreachable — which is the case for every existing arch model.

Only the get side of a channel is costed, so the latency goes there. `air.dma_memcpy_nd` reserves no ports, so for DMA the split is inert and latency simply extends `end_time`.

### Testing

Two new tests. `channel_latency` checks a single transfer completes at `occupancy + latency`. `link_reuse` is the one that matters: two transfers contending for one port over a 500-cycle link finish in `64 + 64 + 500`, not `2 × (64 + 500)` — the second claims the port one *occupancy* after the first, not one flight time.

`check-air-mlir`: 512 passed, 7 xfail, no failures.

### DIFF
--- a/mlir/include/air/Util/Dependency.h
+++ b/mlir/include/air/Util/Dependency.h
@@ -168,6 +168,15 @@ struct dependencyNodeEntry {
   std::vector<dependencyGraph *> nextDependencyGraphs;
   uint64_t start_time;
   uint64_t end_time;
+  // Time at which the event releases the hardware resources it reserved. For a
+  // data movement event this is the end of the link's *occupancy*
+  // (bytes/bandwidth), which can be earlier than end_time when the link model
+  // also carries a fixed time-of-flight latency. Ports model link bandwidth,
+  // not flight time, so holding them for the latency term would cap link
+  // throughput at 1/(occupancy + latency). When no latency is modelled this
+  // equals end_time and the early-release path is inert.
+  uint64_t release_time;
+  bool resources_released;
   std::vector<std::pair<uint64_t, uint64_t>> start_end_time_log;
   // Token count is used to synchronize operations which consumes/produces
   // multiple async tokens.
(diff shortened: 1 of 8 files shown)


erwei-xilinx:
Fixed — and thank you, that was exactly backwards.

Latency is what makes `release_time < end_time`, which is what `has_link_latency()` tests and what opens the window where the ports are free but the payload has not landed. Equality is the *backward-compatible* case, in which there is no window and the path is inert. The comment stated the no-latency case as though it were the latency one.

Rewritten to say which direction the inequality goes and to tie it to `has_link_latency()` explicitly, since that predicate is the thing a maintainer will be reasoning against.

I checked the rest of the change for the same inversion: the header's `has_link_latency()` comment and the commit message both had it right, so this was the only place.

`check-air-mlir` unchanged at 512 passed / 7 xfail / no failures.

PR #1903, opened 2026-08-25
[python] air.api: raise when a segment or herd body is never registered

erwei-xilinx:
`with air.segment(...)` and `with air.herd(...)` are pure bookkeeping — `__enter__` and `__exit__` are no-ops and every op comes from the `body` decorator. Omitting the decorator therefore emits **nothing** for that scope, and the ops inside it are traced into the enclosing one.

That is the worst available failure mode: the IR is structurally different, it still builds, and on a small grid it still runs and still passes — so neither a hardware test nor an op-count diff notices.

It reached review once already, in #1900. A missing `@seg.body` dropped both the segment *and* the launch, leaving a bare `air.herd` where the predecessor had `air.launch` → `air.segment` → `air.herd`. The npu1 lit passed on both element types anyway, and the IR diff I ran compared only the `air.dma_memcpy_nd` lines — which were correct the whole time.

## Where the holes actually are — measured, not assumed

| context | body omitted | verdict |
|---|---|---|
| `air.launch` | `RuntimeError: launch has no body` | genuinely guarded |
| `air.herd` | `RuntimeError: kernel writes no output` | **incidental** |
| `air.segment` | builds fine, segment vanishes | unguarded |

The `herd` row looked safe, so I tested it rather than trusting it. That error is the *no-output* check, not a missing-body check, and any **other** herd that stores satisfies it:

```python
with air.herd(..., name="h1") as h:
    @h.body
    def _(tx): ...   # stores
with air.herd(..., name="h2"):
    pass             # body forgotten
```

builds cleanly and drops `h2` silently. So both holes are real; only `launch` is genuinely covered.

## The fix

```python
def __exit__(self, exc_type, exc_val, tb):
    if exc_type is None and not self._registered:
        raise RuntimeError(f"{self._what} was opened but its body was never registered; ...")
    return False
```

on `SegmentContext` and `HerdContext`.

**Why `__exit__` rather than a `build()`-time check:** the error fires at the `with` statement that caused it. Deferring would need global "entered but unregistered" state and would report far from the cause — which is exactly the weakness of launch's existing message.

**`exc_type is None` is load-bearing.** If the body itself raised, `__exit__` must stay quiet: a body that failed is more interesting than a body that is absent, and at that point the failure is the *only* reason it never registered. Pinned as `a_failing_body_is_not_masked`.

This mirrors the guard already present for registering a body **twice** — same invariant, other end.

`air.launch` is deliberately left alone: it already errors with a good message, and moving its check to `__exit__` would change which error existing users see, for no benefit. The asymmetry is commented rather than refactored away.

## Gates

- `python/test/api` **24/24**, with three new tests: one per hole, plus the not-masked case.
- **All 45 buildable air.api examples still build**, and their `-p` output is **byte-identical** to the unguarded baseline — no correct program is affected.
- Nothing in the tree relied on a body-less context: I scanned every converted example, and the only apparent hit was inside a docstring.

118 lines, of which 78 are tests. No behaviour change for correct code.

### DIFF
--- a/python/air/api/_trace.py
+++ b/python/air/api/_trace.py
@@ -695,10 +695,14 @@ class SegmentContext:
         # replacing it here would bury the real error.
         if exc_type is None and not self._registered:
             raise RuntimeError(
-                f"{self._what} was opened but its body was never registered; "
-                f"decorate one with @<name>.body inside the `with` block. "
-                f"Without it nothing is emitted for this scope and the ops "
-                f"inside it are traced into the enclosing one"
+                f"{self._what} was opened but its body was never registered, so "
+                "nothing was emitted for this scope and the ops inside it were "
+                "traced into the enclosing one. The body is a function decorated "
+                "inside the `with`, which means the scope needs a name to "
+                f"decorate:\n\n    with {self._what}(...) as scope:\n"
+                "        @scope.body\n        def _():\n            ...\n\n"
+                "`as scope` is easy to leave off, and without it there is no "
+                "variable to hang the decorator on."
             )
         return False
 
@@ -943,10 +947,14 @@ class HerdContext:
         # replacing it here would bury the real error.
         if exc_type is None and not self._registered:
             raise RuntimeError(
-                f"{self._what} was opened but its body was never registered; "
-                f"decorate one with @<name>.body inside the `with` block. "
-                f"Without it nothing is emitted for this scope and the ops "
-                f"inside it are traced into the enclosing one"
+                f"{self._what} was opened but its body was never registered, so "
+                "nothing was emitted for this scope and the ops inside it were "
+                "traced into the enclosing one. The body is a function decorated "
+                "inside the `with`, which means the scope needs a name to "
+                f"decorate:\n\n    with {self._what}(...) as scope:\n"
+                "        @scope.body\n        def _():\n            ...\n\n"
+                "`as scope` is easy to leave off, and without it there is no "
+                "variable to hang the decorator on."
             )
         return False
 
--- a/python/air/api/_trace.py
+++ b/python/air/api/_trace.py
@@ -649,6 +649,8 @@ class SegmentContext:
     reason every hand-written staging example in the tree nests the two.
     """
 
+    _what = "air.segment"
+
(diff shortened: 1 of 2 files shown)


python/air/api/_trace.py
```diff
@@ -680,7 +682,28 @@ def __init__(self, grid=None, name=None):
...
+        # Not raised while another exception is propagating: the body itself
+        # failing is far more interesting than the body being absent, and
+        # replacing it here would bury the real error.
+        if exc_type is None and not self._registered:
+            raise RuntimeError(
+                f"{self._what} was opened but its body was never registered, so "
+                "nothing was emitted for this scope and the ops inside it were "
+                "traced into the enclosing one. The body is a function decorated "
+                "inside the `with`, which means the scope needs a name to "
+                f"decorate:\n\n    with {self._what}(...) as scope:\n"
```
erwei-xilinx, line 702: Both fair, fixed in 4a3152b8.

**The `@<name>.body` advice.** You are right that it is useless in exactly the case that produces this error most often — and that case is not hypothetical, it is the shape of the bug in #1900. The message now shows the spelling *including* the binding, and names the omission directly:

```
air.segment was opened but its body was never registered, so nothing was
emitted for this scope and the ops inside it were traced into the enclosing
one. The body is a function decorated inside the `with`, which means the
scope needs a name to decorate:

    with air.segment(...) as scope:
        @scope.body
        def _():
            ...

`as scope` is easy to leave off, and without it there is no variable to hang
the decorator on.
```

That is the verbatim output for `with air.segment(name="s"):` with no `as`.

**The dead f-prefixes.** Three of the four literals had one with nothing to interpolate. Only the two using `{self._what}` keep it now.

I did **not** touch six other no-placeholder f-strings elsewhere in `_trace.py`. They are pre-existing — the count is 6 on `origin/main` and 6 on this branch, so this PR introduces none — and cleaning them up belongs in its own commit rather than widening a guard PR. Happy to do that separately if you would like it.

Gates unchanged: `python/test/api` 24/24 (the CHECK lines match the first line of the message, which is stable), all 45 examples build, and their `-p` output is byte-identical to the unguarded baseline.

PR #1904, opened 2026-08-25
[python] air.api: broadcast an operand axis of extent 1, and vector_broadcast_scalar

erwei-xilinx:
`air.api` had no buffer-to-buffer broadcast. Only a Python scalar could be
stretched across a tile — `_check_region` demanded that every operand leaf have
exactly the destination's shape — so the one kernel in the tree whose whole point
is a broadcast had to be written by hand.

## The rule

numpy's, right-aligned and **one-sided**: each operand broadcasts *to* the
destination rather than the two being broadcast against each other, because the
destination of an `air.api` assignment is a buffer that already exists. That is
numpy's own rule for an explicit `out=`. Every operand axis must match the
destination's or be 1; leading axes the operand does not have are the 1 case
written by omission.

## Which code comes out depends on which axis is stretched

* An **outer** axis leaves the innermost one intact, so the read is the ordinary
  `vector.transfer_read` — at the *operand's* own rank, pinned at 0 on the
  stretched axes.
* The **innermost** axis has no contiguous run to read at all, so the emitter
  takes the single element with `memref.load` and splats it with
  `vector.broadcast`.

The second is what `vector_broadcast_scalar` needs; the first is the bias vector
in `mnist_fc/broadcast_bias_add`, which is **not** converted here — see below.

`_eval`'s `minor` parameter becomes a `load` callable, because how a leaf is read
is now a property of the leaf rather than of the nest.

A reduction still refuses to broadcast, and now says why: its innermost extent is
the thing being collapsed, so stretching it would decide how many terms the sum
has rather than making an operand fit.

## The example

`vector_broadcast_scalar` becomes

```python
c[:] = a[:]
```

with a `[tile_m, 1]` source and a `[tile_m, n]` destination. The predecessor said
it with two `memref.subview`s, two `memref.collapse_shape`s carrying an explicit
dynamic `StridedLayoutAttr`, a `memref.load` and a `vector.broadcast`. The
subviews existed only to get a rank-1 view the vector ops would accept, and
indexing a rank-2 buffer at `[j, 0]` needs neither. The emitted `memref.load` /
`vector.broadcast` / `vector.transfer_write` is unchanged, and so is the op
hierarchy — one `air.herd`, no segment.

## Gating

* `python/test/api/` 25/25, including a new `broadcast.py` covering the innermost
  axis, a missing leading axis, an explicit leading 1, a middle axis, the scalar
  path, and four new refusals in `errors.py`.
* **All 45 pre-existing `air.api` examples emit byte-identical IR.** When nothing
  is stretched, `_leaf_index` returns the destination's own induction variables
  unchanged and no pinned-zero constant is emitted at all. `no_broadcast_emits_no_pinned_index`
  pins that, and it is control-tested: forcing the constant on makes exactly that
  test fail.
* `run_makefile_peano.lit` **PASSes on npu1 hardware**, and the gate is not
  vacuous — perturbing the kernel by `+ 1.0` fails it with diff exactly 1.0.
* npu2 xclbin and elf both compile clean. npu1/elf fails, but the backend refuses
  that combination for the predecessor identically, which is why the elf lit is
  `REQUIRES: ryzen_ai_npu2`.

## Not converted: `mnist_fc/broadcast_bias_add`

It needs only the outer-axis half of this feature, but converting it ran into a
separate gap: **`air.api` cannot use a launch coordinate inside a herd body.**
`_trace.py:978` threads a *segment's* own grid coordinates into the herd but not
the *launch's*, so the coordinate crosses the `IsolatedFromAbove` boundary and the
module fails verification with `'affine.apply' op using value defined outside the
region`.

The fix is one line — also threading `current_launch().leaves` — and it works, but
it changes IR for 6 of the 45 examples (matmul i16/i8, matvec bf16, all three
vecmat), because an unused herd operand is not free: it survives `air-dependency`
as an async edge, which is why that line was gated in the first place. That is its
own change with its own hardware gating, so this PR reports the gap rather than
part-converting the example.

### DIFF
--- a/programming_examples/primitives/vector_examples/vector_broadcast_scalar/vector_broadcast_scalar.py
+++ b/programming_examples/primitives/vector_examples/vector_broadcast_scalar/vector_broadcast_scalar.py
@@ -1,181 +1,98 @@
 # Copyright (C) 2025, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
-import argparse
-from ml_dtypes import bfloat16
+"""Broadcast one scalar per row across a whole row, on air.api.
 
-from air.ir import *
-from air.dialects.affine import apply as affine_apply
-from air.dialects.air import *
-from air.dialects.arith import ConstantOp
-from air.dialects.memref import AllocOp, DeallocOp, load, store, subview, collapse_shape
-from air.dialects.vector import (
-    transfer_read,
-    transfer_write,
-    reduction,
-    extract,
-    CombiningKind,
-    broadcast,
-)
-from air.dialects.func import FuncOp
-from air.dialects.scf import for_, yield_
-from air.dialects.math import exp
-from air.backend.xrt_runner import XRTRunner, type_mapper
-from air.backend.xrt import XRTBackend
+    c[:] = a[:]
+
+with ``a`` shaped ``[tile_m, 1]`` and ``c`` shaped ``[tile_m, n]``. That is the
+entire kernel. The right-hand side is a plain copy and the shapes are what make
+it a broadcast: numpy's rule, right-aligned, an operand axis either matches the
+destination's or is 1 and gets stretched.
+
+The predecessor said the same thing as a per-row ``scf.for`` containing two
+``memref.subview``s, two ``memref.collapse_shape``s with an explicit dynamic
+``StridedLayoutAttr``, a ``memref.load`` and a ``vector.broadcast``, plus a
+``vector.transfer_write``. The emitted arithmetic is unchanged -- when the
+broadcast axis is the *innermost* one there is no contiguous run to read, so the
+emitter loads the single element and splats it, which is exactly the
+``memref.load`` + ``vector.broadcast`` pair above.
+
+The subviews and collapse_shapes were only ever there to get a rank-1 view the
+vector ops would accept; indexing a rank-2 buffer at ``[j, 0]`` needs neither.
+
+Two differences from the predecessor worth naming, both shared with the
+``average_pool`` conversion this example is the mirror image of:
+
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Good catch, and correct — fixed in `31fc4624`.

`_is_broadcast` asked whether the shapes *differ*, which is true for a `[16]` bias against an `[8, 16]` tile. But that operand is short of a **leading** axis, and a missing axis contributes no index at all, so nothing is pinned and the hoisted constant was dead. Reproduced before fixing:

```
%cst = arith.constant 0.000000e+00 : bf16
%c0  = arith.constant 0 : index          <-- never referenced
%c0_3 = arith.constant 0 : index         <-- loop lower bound
...
%0 = vector.transfer_read %alloc_1[%arg6, %arg7], %cst ...
%1 = vector.transfer_read %alloc_2[%arg7], %cst ...
```

The predicate now asks the narrower question you suggest — does an axis the operand *actually has* get pinned:

```python
def _pins_an_axis(shape, dst_shape):
    offset = _broadcast_offset(shape, dst_shape)
    return any(extent != dst_shape[i + offset] for i, extent in enumerate(shape))
```

That is deliberately the negation of the condition in `_leaf_index`, and the docstring says so: the two have to agree, or the constant is either dead or — worse — used before it is emitted.

`a_missing_leading_axis_reads_at_its_own_rank` now pins the absence with the same shape of check the no-broadcast case already used, and it is control-tested: restoring the over-broad predicate fails exactly that check and nothing else.

Gates re-run: `python/test/api` 25/25, all 45 pre-existing examples still byte-identical against a baseline rebuilt from the current `main`, and the npu1 hardware lit still passes.

PR #1906, opened 2026-08-26
[programming_examples] air.api: rope_sincos, and ops.load filling a view

erwei-xilinx:
Converts `rope_sincos` — RoPE with on-chip sin/cos — onto `air.api`, plus the one
small DSL fix it needed.

## The example

The arithmetic all lives in `rope.cc`, so what this example *is* is five kernel
calls on one head of L1: move it in, copy it (V passes through), compute
freq*position, sin and cos by Chebyshev polynomial, rotate Q and K in place,
move it out.

```python
air.ops.load(l1_in.reshape(3 * head_size), IN[o : o + 3 * head_size])
vector_copy(l1_in, l1_out)
freq_pos(1, l1_freq_pos)
sinf(l1_freq_pos, l1_sin)
cosf(l1_freq_pos, l1_cos)
shuffle_apply_rope(0,         l1_cos, l1_sin, l1_out)
shuffle_apply_rope(head_size, l1_cos, l1_sin, l1_out)
air.ops.store(l1_out.reshape(3 * head_size), OUT[o : o + 3 * head_size])
```

Five `air.extern`s replace five hand-built `FuncOp` declarations plus the loop
that stamped `link_with` and `llvm.emit_c_interface` on each. The declaration is
derived from the call, so only the scalar element types are stated — a Python
`1` cannot say whether the kernel wants `i32` or `index`.

The herd is `[herd_n, 1]` rather than `[1, herd_n]`, and the strip-mine is the
DSL's: one point per head with `shape=` pinning the core count, where the
predecessor wrote the outer `scf.for` and the `AffineMap` combining it with the
core coordinate by hand.

**A/B against the predecessor's IR:** same op hierarchy (one `air.herd`), the
same five private declarations, and byte-identical `func.call` signatures. The
only op-count difference is one extra `affine.apply`, emitted per DMA rather than
hoisted, which CSE folds.

## `ops.load` filling a view

`reshape`/`transpose` produce a view, and a view has always been a valid
*source* for `ops.store` — `store(t.transpose(1, 0), B[:, :])` is the whole of
data_transfer_transpose. Filling one was refused. That read as a rule and was an
omission: everything downstream already handles a view endpoint (that is how
store's source works), and `load` simply checked `isinstance(dst, Buffer)` where
`store` checks `(Buffer, BufferSlice)`.

rope_sincos is the first caller: its L1 tiles are `[3, head_size]` because that
is what `rope.cc` is compiled against, while its L3 tensor is one flat run per
head, so it has to reshape on the way *in*.

A view still does not excuse a shape mismatch — handing over a buffer whose shape
genuinely disagrees is refused exactly as before, naming both shapes.
`python/test/api/views.py` pins both, and pins that no `memref.collapse_shape` or
`memref.reshape` is emitted: a view is sizes and strides on the transfer, not an
operation on the buffer. Control-tested against the unfixed `ops.py`.

## The buffers are released explicitly, and that is load-bearing

The five `air.dealloc` calls at the end are not tidiness. **Left to place them
itself, the tracer frees each buffer after the last use it observes** — so
`l1_in` goes right after `vector_copy` while four more kernels are still running
on the others — and the result is *wrong on hardware*: a handful of Q/K rotation
pairs come back as huge values or NaN.

This was found because the conversion failed its lit while the predecessor passed
on the same box, and isolated by controlled A/B: the explicit deallocs are the
only difference between the passing and failing runs. In particular it is **not**
the new view feature — flat-tensor-plus-reshape and shape-exact-2-D-tensor both
fail without the deallocs and both pass with them.

**This is not automatic dealloc being broken in general.** 28 of the converted
examples free a buffer before their last op and all pass their lits, so the
trigger is narrower than that. I have not characterised it further, and I would
rather say so than guess. It seems worth chasing separately, since it is a
silent-wrongness mode in already-landed code.

## Gating

* `run_makefile_peano.lit` **PASSes on npu1 hardware**, and the predecessor passes
  on the same box, so the comparison is like for like.
* `python/test/api/` 25/25.
* All other `air.api` examples emit byte-identical IR against a baseline built
  from this branch's merge base.

### DIFF
--- a/programming_examples/rope_sincos/rope_sincos.py
+++ b/programming_examples/rope_sincos/rope_sincos.py
@@ -81,7 +81,17 @@ def build_module(num_heads, head_size, herd_n):
     # The kernels are specialised for head_size=48: the frequency table is
     # hardcoded and the symbol names carry the size.
     assert head_size == 48, f"Only head_size=48 is supported, got {head_size}"
-    assert (num_heads * head_size) % herd_n == 0
+    # The herd's iteration space is one point per head, strip-mined onto herd_n
+    # cores, so it is the *head count* that has to divide -- not the element
+    # count, which is what the predecessor checked because its herd was
+    # [1, herd_n] with the outer loop striding over every element. The weaker
+    # form lets head_size carry the divisibility: num_heads=6 with herd_n=4
+    # passes (6 * 48 is a multiple of 4) and then fails inside air.herd with
+    # "herd shape (4,) does not evenly divide the logical grid (6,)".
+    assert num_heads % herd_n == 0, (
+        f"num_heads ({num_heads}) must be a multiple of herd_n ({herd_n}): "
+        "each core takes a whole number of heads"
+    )
 
     head_stride = 3 * head_size
 
@@ -93,9 +103,10 @@ def build_module(num_heads, head_size, herd_n):
     )
     vector_copy = air.extern("vector_copy_bf16_144_16", link_with=EXTERN_OBJECT)
 
-    # [3 * num_heads, head_size] rather than one flat run: the same bytes,
-    # but a whole head is now a [3, head_size] region, which is exactly the
-    # L1 tile's shape, so both transfers are plain shape-matching ones.
+    # One flat run, as the predecessor declared it and as the host writes it:
+    # num_heads heads of Q, K and V concatenated. The L1 tiles are [3, head_size]
+    # instead, because that is rope.cc's shape, so the transfers go through a
+    # reshaped view rather than either side giving up its own shape.
     IN = air.tensor([num_heads * head_stride], bf16)
     OUT = air.tensor([num_heads * head_stride], bf16)
 
--- a/programming_examples/rope_sincos/rope_sincos.py
+++ b/programming_examples/rope_sincos/rope_sincos.py
@@ -1,197 +1,161 @@
 # Copyright (C) 2026, Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Both correct, both mine — fixed in `3d172e35`.

**The assertion.** You are right that it does not check what this conversion needs, and I reproduced the gap rather than just tightening it:

```
assert passes for num_heads=6, herd_n=4? True   but 6 % 4 = 2
ValueError: herd shape (4,) does not evenly divide the logical grid (6,);
            each dimension must strip-mine exactly
```

The reason it was wrong is worth naming: the predecessor's herd was `[1, herd_n]` with an outer loop striding over every *element*, so checking the element count was right for it. Here the iteration space is one point per *head*, so `head_size` can supply the divisibility on its own and hide a bad head count. It now checks `num_heads % herd_n` and says why:

```
AssertionError: num_heads (6) must be a multiple of herd_n (4):
                each core takes a whole number of heads
```

**The stale comment.** Also right, and it has a specific cause. While isolating the dealloc bug I tried a shape-exact spelling — L3 declared `[3 * num_heads, head_size]` so a head is a `[3, head_size]` region and no view is needed — to find out whether the reshaped-view DMA was what broke it. It was not (the deallocs were), so I reverted the code back to flat tensors and reshaped views, and failed to revert the comment with it. It now describes the flat run and says why the view is there.

I grepped for other survivors of that experiment and found none.

Gates re-run: `python/test/api` 25/25, `run_makefile_peano.lit` still PASSes on npu1 hardware.

PR #1912, opened 2026-08-26
[utils] Bump mlir-aie to 1.4.3.dev17+g4c8b56f

erwei-xilinx:
## Description

Bumps mlir-aie from `1.4.2.dev16+g7e00b57` to `1.4.3.dev17+g4c8b56f`.

The target is the merge commit of [mlir-aie#3580](https://github.com/Xilinx/mlir-aie/pull/3580) itself, which splits aiecc's two `--unified` behaviours apart: a device's cores are lowered once, but still compile to separate objects in parallel. That is the prerequisite for #1873, which asks aircc for the flag; before the split, `--unified` also merged the objects and regressed single-device designs 2.7x.

LLVM is unchanged. Both sides pin `56bcc1871734e6c375a254dec0ec74eb18d04a2e` / `DATETIME=2026080106`, so `utils/clone-llvm.sh` stays as it is.

## What the bump drags in

Three upstream changes touch op signatures we build:

- **#3577** (out-of-order S2MM DMA) inserts `out_of_order_id` into `aie.dma_bd` between `packet` and `burst_length`, and adds `out_of_order` to `aie.dma_start`.
- **#3546** adds `axcache` to `aie.dma_bd` and `aiex.npu.write_bd`.
- **#3559** packs the dynamic BD register block into one blockwrite, changing what `emit_txn_cpp` sees.

Four `create()` call sites in `AIRRtToNpuPass.cpp` and `AIRToAIEPass.cpp` are adapted positionally, and the `emit_txn_cpp` CHECKs follow the new blockwrite shape. No behaviour change intended on our side.

## Two AIE1 tests xfailed

mlir-aie has dropped AIE1. Two tests depend on it; the other 17 files naming `xcvc1902` still pass, so only the repeat-count path is affected.

`async_gemm_w_pingpong_to_locks` asks for a multi-pass transfer on a target whose `getMaxRepeatCount()` is 0. #3577's new `verifyDMARepeatCount` rejects it — and that verifier exposed an existing miscompile rather than causing one. The AIE1 backend branches on architecture before repeat is ever read:

```cpp
if (target_model.getTargetArch() == AIEArch::AIE1) {
  XAie_DmaChannelPushBdToQueue(..., /* BdNum */ bdNum);   // no repeat argument
} else if (op.getOutOfOrder()) {
  int repeatCount = op.getRepeatCount() + 1;              // AIE2 only
```

So the count never reached the emitted CDO: that transfer has been running once where the IR asked for twice.

`lower_herd_air_regions` is AIE1 only by accident — it passes no `device=`, so `air-to-aie` falls back to its `xcvc1902` default, and its CHECKs assert AIE1 lock semantics. I tried to retarget it instead of xfailing: npu2 additionally needs `row-offset=2` (NPU core rows start at 2), and even then it emits `AcquireGreaterEqual` where the test expects `Acquire`. Half-porting it would assert less than it does now, so it is xfailed with the other.

I also tried to *fix* the repeat-count case rather than xfail it, by unfolding an N-pass task into N single-pass ones. That trades an invalid repeat count for BD exhaustion — `'aie.mem' op has more than 16 blocks` on the pingpong test — so it is not in this PR.

## Follow-up this exposes

`air-to-aie` still defaults to `xcvc1902`, so any invocation omitting `device=` now targets an architecture upstream no longer supports. That default is exactly how `lower_herd_air_regions` got caught. Changing it is a behavioural change with its own blast radius, so it is left for a separate PR.

## Verification

Ryzen AI 9 HX 370 (Strix, npu2), Peano.

| check | result |
|---|---|
| `check-air-mlir` | 538 tests, 522 passed, **0 unexpected failures** (xfail 7 -> 9) |
| `llama32_1b_q4nx` `make compile` | passes |
| `llama32_1b_q4nx` `make verify` (hardware) | **PASS 2/2** |
| GEMV bf16 npu2 configs (hardware) | PASS |

Not covered: Chess (no Vitis on this host) and npu1 hardware.

### DIFF
--- a/mlir/lib/Conversion/AIRRtToNpuPass.cpp
+++ b/mlir/lib/Conversion/AIRRtToNpuPass.cpp
@@ -878,7 +878,9 @@ struct DmaToNpuPattern : public OpConversionPattern<airrt::DmaMemcpyNdOp> {
           /*sizes=*/ValueRange{}, /*strides=*/ValueRange{},
           /*static_sizes=*/nullptr, /*static_strides=*/nullptr,
           /*pad_dimensions=*/nullptr, /*bd_id=*/nullptr, pktAttr,
-          /*burst_length=*/nullptr, /*iteration=*/nullptr,
+          /*out_of_order_id=*/nullptr,
+          /*burst_length=*/nullptr, /*axcache=*/nullptr,
+          /*iteration=*/nullptr,
           /*offset_parameter=*/nullptr,
           /*offset_state_table_idx=*/nullptr, /*next_bd_id=*/nullptr);
     } else if (dynOffsetI32) {
@@ -4453,7 +4455,8 @@ struct AIRRtToNpuPass : public impl::AIRRtToNpuBase<AIRRtToNpuPass> {
             /* d0_zero_before */ 0, /* d1_zero_before */ 0,
             /* d2_zero_before */ 0,
             /* d0_zero_after */ 0, /* d1_zero_after */ 0,
-            /* d2_zero_after */ 0);
+            /* d2_zero_after */ 0, /* burst_length */ 0,
+            /* axcache */ nullptr);
         uint32_t addr = (dstColIndex << target_model.getColumnShift()) |
                         (0x1D004 + bdID * 0x20);
         {
--- a/mlir/lib/Conversion/AIRToAIEPass.cpp
+++ b/mlir/lib/Conversion/AIRToAIEPass.cpp
@@ -6491,13 +6491,15 @@ public:
         if (!channel_head) {
           channel_head = start_bb;
           auto b = OpBuilder::atBlockBegin(channel_head);
-          startOp = AIE::DMAStartOp::create(b, loc, dir, chan, rep,
-                                            /*pad_value*/ 0, first_bd, end_bb);
+          startOp =
+              AIE::DMAStartOp::create(b, loc, dir, chan, rep,
+                                      /*pad_value*/ 0,
+                                      /*out_of_order*/ false, first_bd, end_bb);
         } else {
           auto b = OpBuilder::atBlockBegin(start_bb);
           startOp = AIE::DMAStartOp::create(
-              b, loc, dir, chan, rep, /*pad_value*/ 0, first_bd,
-              channel_head->getTerminator()->getSuccessor(1));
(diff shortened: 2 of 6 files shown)


erwei-xilinx:
Converted both AIE1 tests to AIE2 rather than leaving them xfailed (a1b144b6). Both lower fine on AIE2, so skipping them was giving up coverage for no reason. `check-air-mlir` is now **524 passed / 0 failed**, with xfail back down to 7 — the bump costs no tests.

One trap worth recording. `async_gemm_w_pingpong_to_locks` pinned its placement in two places — `row-offset=3 col-offset=5` on the command line and `x_loc`/`y_loc` on the `air.herd` — and the attribute is what actually binds. Changing the flags alone still emitted column 5 and tripped npu1's 4-column limit:

```
error: 'aie.tile' op column index (5) must be less than the number of columns in the device (4)
```

Both now point at cols 0-1 / rows 2-3. The repeat count #3577 rejected is in range on AIE2 (`getMaxRepeatCount()` 0 -> 255), so nothing needs unfolding.

Two assertions changed, because AIE2 genuinely differs rather than to make things pass:

- `Acquire` -> `AcquireGreaterEqual` throughout; AIE2 has no plain `Acquire`.
- Locks per tile 6 -> 8, and no longer all zero-initialised — AIE2 semaphores carry the ping-pong depth: four `init 0`, two `init 1`, two `init 2`. I kept the count exact and matched init as `{{[0-2]}}` rather than `{{.*}}`, so a lock initialised outside that range still fails.

`lower_herd_air_regions` named no device at all and inherited the `xcvc1902` default; it now asks for npu1 with `row-offset=2`, since NPU core rows start at 2.

Both are perturbation-tested: rewriting `AcquireGreaterEqual` back to `Acquire` in the CHECKs makes each test fail, so the lock assertions still discriminate rather than matching anything.

Re-verified after the change: `check-air-mlir` 524/0, `llama32_1b_q4nx` `make compile` passes, `make verify` **PASS 2/2** on hardware.

The `air-to-aie` default-device follow-up still stands — `lower_herd_air_regions` inheriting `xcvc1902` is exactly the trap, and it will catch the next test that omits `device=`.

erwei-xilinx:
Closing and reopening to kick GitHub into dispatching CI — this PR has had zero check suites created since it was opened, while other PRs from the same fork the same day ran normally. No change to the branch.

PR #1913, opened 2026-08-26
[programming_examples] air.api: herd_dataflow, plus air.parallel

erwei-xilinx:
`herd_dataflow` onto `air.api`, the DSL surface it turned out to need, and a compiler crash found on the way.

## `air.parallel`

`air.parallel` was a stub that raised, on the reasoning that *"an `scf.parallel` reaches the spatial hierarchy only when a conversion pass selects it, which in general it does not"*. That is true of a herd grid and wrong at segment scope, where staging fans a memtile buffer out to a row of cores. Two things make that loop parallel rather than merely repeated, and `air.sequential` supplies neither:

- **The trip index names one slot of a channel bundle.** `air-place-herds` refuses to take that from a temporal loop — *"channel bundle indices must not be temporal scf.for induction variables"*.
- **The trips share one set of buffer descriptors.** Writing them out as a Python `for` unrolls into that many independent DMAs, which is a different design: herd_dataflow's four per-column L3 reads fit on npu1 inside the loop and do not fit unrolled, with *"no ShimNOCTile has sufficient DMA capacity for 0 input/2 output channels"*.

This is not a judgement call about style — the example's own checked-in `air.mlir`, which three of its six lits parse via `--mlir-source file`, has read `scf.parallel` with the loop IV as the bundle index all along:

```mlir
scf.parallel (%par1) = (%c0) to (%c4) step (%c1) {
  air.dma_memcpy_nd (%alloc_1[%c0, %apply] ...)
  air.channel.put @L2ToL1Chan1[%par1, %c0] (...)
}
```

`air.parallel` emits `scf.forall`, which the pipeline's `scf-forall-to-parallel` turns into exactly that. Bounds validation mirrors `air.sequential`. `errors.py` loses the case pinning the old `NotImplementedError` and gains one for a step that does not tile the extent — overrunning matters more here, because the trips are fan-out slots and the last one would address a destination that does not exist.

## The conversion

Three herds and a tile that flows through all of them without going back to L2: `herd_0` adds, `herd_1` copies and hands the result over a cascade channel, `herd_2` calls the hand-written `add_3_bf16`. A channel is a module-level symbol, so each stage finds what the previous one sent by name, with nothing threaded through the segment.

Both compute bodies are one line. `c[:] = a[:] + b[:]` replaces a doubly-nested loop building every 16-element strip out of two `memref.subview`s and a `memref.collapse_shape`; `c[:] = a[:]` replaces a scalar `memref.load`/`memref.store` loop over all 4096 elements, and is now vectorised at the same 16 lanes as the add.

`air.mlir` is regenerated from this builder, so `--mlir-source file` parses the same design the Python path builds. It is target-independent — npu1 and npu2 print identically — which is what lets one file serve both generations' lits. Against the predecessor's, the segment holds the same three `scf.parallel` loops and the same one put and one get per channel across all five channels.

## `air-fuse-channels`: do not index an empty put/get list

`AIRFuseChannels` reads `a_vec[0]` as the access pattern every other op is compared against, and `a_puts[0]`/`b_puts[0]`/`a_gets[0]`/`b_gets[0]` as the ops to merge, in all four cases without checking the list is non-empty. A channel reaches these with no puts, or no gets, once an earlier fusion in the same run has moved them onto another symbol, and the pass aborts on

```
std::vector<xilinx::air::ChannelGetOp>::operator[]: Assertion '__n < this->size()' failed
```

which in a build without assertions is an out-of-bounds read instead. Both sites are hit in turn — guarding only the first moves the abort to the second.

Reached from valid input: a segment that stages a row of tiles into one memtile buffer and puts each column's slice on a bundled channel with a constant index. That is what a plain Python loop produces in air.api, and while herd_dataflow is better written with `air.parallel`, nothing rejects the unrolled spelling, so the crash is reachable. The new test is the IR that form reaches this pass with; it aborts the pass without the fix.

## Testing

- Both npu1 `herd_dataflow` lits pass — python source and file source. The four npu2 lits are UNSUPPORTED on my box.
- `check-air-mlir` 525/525 (new `air-fuse-channels` case fails against the previous behaviour), `check-air-python` 40/40, `api/` suite 27/27.
- npu1 programming-examples sweep: 88 passed.
- 46 air.api examples emit byte-identical IR.

### DIFF
--- a/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
+++ b/mlir/lib/Transform/AIRDependencyScheduleOpt.cpp
@@ -4577,6 +4577,12 @@ private:
   template <typename T>
   bool areConsistentMemoryAccessPattern(std::vector<T> a_vec,
                                         std::vector<T> b_vec) {
+    // a_vec[0] is the pattern every other op is compared against, so there is
+    // nothing to compare when it is empty -- and indexing it is out of bounds.
+    // A channel reaches here with no puts, or no gets, once an earlier fusion
+    // in this same pass has moved them onto another symbol.
+    if (a_vec.empty())
+      return false;
     Value memref = a_vec[0].getMemref();
     SmallVector<Value> offsets = air::getOffsetsAsValues(a_vec[0]);
     SmallVector<Value> sizes = air::getSizesAsValues(a_vec[0]);
@@ -5379,10 +5385,16 @@ private:
         getChannelGetOpThroughSymbol(chan_a);
     std::vector<air::ChannelGetOp> b_gets =
         getChannelGetOpThroughSymbol(chan_b);
-    if (!b_puts[0]->getParentOfType<air::HerdOp>()) {
+    // Each side is merged only when both channels have one to merge. A channel
+    // whose puts or gets have already been moved onto another symbol earlier
+    // in this pass reaches here with an empty list, and there is simply
+    // nothing to do for that half -- indexing it is out of bounds.
+    if (!a_puts.empty() && !b_puts.empty() &&
+        !b_puts[0]->getParentOfType<air::HerdOp>()) {
       mergeChannelOpsTemporally(a_puts[0], b_puts[0], mergeByLBOrUB);
     }
-    if (!b_gets[0]->getParentOfType<air::HerdOp>()) {
+    if (!a_gets.empty() && !b_gets.empty() &&
+        !b_gets[0]->getParentOfType<air::HerdOp>()) {
       mergeChannelOpsTemporally(a_gets[0], b_gets[0], mergeByLBOrUB);
     }
   }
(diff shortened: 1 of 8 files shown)


erwei-xilinx:
Both comments addressed — one fixed, one declined with a reason.

**Stale staging bullet (`run.py:31`) — fixed in `eaebd034`.** Worse than stale: the bullet also claimed *"both end up as NUM_COLUMNS independent transfers"*, which is exactly the thing this branch disproves and the reason `air.parallel` exists. Left in, the file that demonstrates the distinction would have taught the opposite. Rewritten to say what the code does and why the Python `for` is deliberately not used.

**`--target` choices (`run.py:208`) — not changing.** Two reasons:

1. The value is already validated one layer down, with a message that names the accepted set from its single source of truth:
   ```
   $ ./run.py -p --target npu9
   ValueError: unknown target 'npu9'; expected 'auto' or one of ['npu1', 'npu2']
   ```
2. 55 of the 63 examples that take `--target` omit `choices` for that argument; only 8 have it. Adding it here would make this file the outlier and put a 56th copy of the accepted-value list next to a central one that already exists, which is what drifts.

If the project would rather have the fast-fail everywhere, that is worth doing as one sweep across all 55 — or better, by having the shared helper install the argument — rather than one file at a time.

PR #1925, opened 2026-08-27
[air] Let a hierarchy op's channel dependences be seen from outside it

erwei-xilinx:
`CanonicalizeAsyncOpDeps` drops an async edge when it can prove the two ends share neither a memref nor a symbol. `getSymbolRefsUsedByOp` gathers the symbols for that test, and it returned early for any `AsyncOpInterface` op without looking inside its regions.

`air.herd`, `air.segment` and `air.execute` are async **and** hold regions, and their channel traffic lives entirely in the region — the symbol never appears in the op's own attributes. So the helper reports an incomplete symbol set for exactly the ops whose symbols are hardest to see, the "these two share nothing" conclusion is reached on a false premise, and the edge is removed whenever the two ends also happen not to share a memref.

The fix is to walk the regions of any op that has them. Region-less async ops (`channel.put`/`get`, `dma_memcpy_nd`) are unaffected — their symbol is already in their attributes and there is nothing to walk.

## The observable difference

`air-opt -canonicalize` on the first test case:

```mlir
before:  air.herd @h async [%async_token_1]
after:   air.herd @h async [%async_token_1, %2]      // %2 is the channel.put
```

## What is and is not claimed

That the analysis is **unsound** is not in question: a helper whose contract is "the symbols this op uses" was returning a subset, and a transform that deletes edges was consuming it.

Whether that unsoundness currently produces a wrong result for anyone is a **separate question, and I have not demonstrated it**. No test in tree caught it — the suite passes identically with and without this change. It was found on a pipelined model where the deleted edge mattered, but that is not something this repository can check, so the tests here assert the analysis rather than a downstream symptom.

I'd rather put that plainly than dress it up. If reviewers or CI think the edge is redundant in practice and the pruning was legitimate, that is a reasonable position and this should be closed.

## Tests

Four cases in `mlir/test/Dialect/AIR/canonicalize_hierarchy_channel_dep.mlir`. Three fail without the fix:

- **herd**, **segment**, **execute** — a channel-mediated dependence into a region-holding op survives canonicalization
- **two herds on disjoint channels, sharing no memref** — the edge between them is *still pruned*

That last one is the one worth reading. Walking the regions makes the analysis see *more* symbols, so the risk this change carries is over-preservation — quietly turning into "never prune an edge into a hierarchy op" and pessimising every schedule. The case asserts a single token in `@hb`'s dependence list, with no comma, so the edge from `@ha` is provably gone.

## Checks

- `check-air-mlir` 545 tests / 527 passed / 7 expectedly failed / 0 failed
- `check-air-python` 38/38
- `git clang-format origin/main` clean
- Negative control: revert the source hunk alone and 3 of the 4 cases fail

### DIFF
--- a/mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
+++ b/mlir/lib/Dialect/AIR/IR/AIRDialect.cpp
@@ -425,9 +425,15 @@ static LogicalResult CanonicalizeAsyncOpDeps(OpT op,
         llvm::SetVector<SymbolRefAttr> result;
         auto opSymbolRefAccesses = getAllSymbolRefAccess(o);
         result.insert(opSymbolRefAccesses.begin(), opSymbolRefAccesses.end());
-        if (isa<air::AsyncOpInterface>(o))
+        // A region-holding op uses every symbol its body uses, whether or not
+        // it is itself async. air.herd, air.segment and air.execute all carry
+        // their channel traffic inside the region: the symbol never appears in
+        // the op's own attributes, so stopping at an async op hides it and the
+        // channel-mediated dependence looks like no dependence at all.
+        // Region-less async ops (channel put/get, dma) are unaffected -- their
+        // symbol is already in their attributes and there is nothing to walk.
+        if (o->getNumRegions() == 0)
           return result;
-        // If op isn't an air.async op, then collect symref uses in its regions.
         SmallVector<Region *> regions;
         for (auto &region : o->getRegions())
           regions.push_back(&region);
--- a/mlir/test/Dialect/AIR/canonicalize_hierarchy_channel_dep.mlir
+++ b/mlir/test/Dialect/AIR/canonicalize_hierarchy_channel_dep.mlir
@@ -1,0 +1,191 @@
+//===- canonicalize_hierarchy_channel_dep.mlir -----------------*- MLIR -*-===//
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: air-opt -canonicalize -split-input-file %s | FileCheck %s
+
+// A hierarchy op uses every channel its body uses. The symbol never appears in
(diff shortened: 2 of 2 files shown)


erwei-xilinx:
Closing this. CI settled it against the change, which is what it was opened to find out.

## The observation was right; the conclusion was wrong

`getSymbolRefsUsedByOp` really does return an incomplete symbol set for region-holding ops, and `CanonicalizeAsyncOpDeps` really does prune the `air.channel.put` -> `air.herd` edge as a result. Both are true.

What I got wrong is treating that as a defect to fix. **The pruning is required.** On hardware, the edge is a circular wait:

- the put is a DMA into a channel of finite depth
- it cannot complete until something drains it
- the drain is the `air.channel.get` inside the herd
- with the edge restored, the herd cannot start until the put completes

Producer and consumer have to overlap. Serialising them hangs the device.

## What CI showed

`main` is green on both hardware jobs across the five commits preceding this branch. With this change:

| job | failures |
|---|---|
| `amd8845hs` | 7 |
| `amdhx370` | 2 |

All `ERT_CMD_STATE_TIMEOUT` — the NPU kernel never completes. Across `26_vecmat_i8`, `12_matmul_transform_1x4_bf16`, `08_gemm_extern_vec`, `09_gemm_extern_vec_4x4`, `29_gemm_4_level_tiling_extern_vec_4x4_bf16`, `40_triton_vec_add`, `42_triton_softmax_bf16`. Not one fragile test — a broad hang, and not flakiness.

## For anyone who finds the same gap later

The region walk is missing **deliberately in effect**, whatever the original intent. If you make `getSymbolRefsUsedByOp` see into regions, you will restore this edge and deadlock every design where a hierarchy op consumes a channel its producer fills. If the incompleteness ever needs addressing, it has to be done without reintroducing that ordering — a narrower rule that distinguishes a channel dependence from other symbol sharing, or a change to how the put/get pair is lowered.

The tests on this branch are still a fair description of the *behaviour*; they simply assert the wrong thing is desirable.

Thanks to the reviewer who pushed on whether this fixed a real bug — the honest answer turned out to be no, and worse than no.

PR #1945, opened 2026-08-28
[air.api] Hotfix: conv1d_depthwise.build_module must return a module (#1927 regression)

erwei-xilinx:
## The failure

Nightly LLM perf CI has been red since 2026-08-28 (run 33149510099, air `6b9792a349d8`). `llms/lfm2_1_2b_q4nx` fails `run_npu2_verify`, `run_npu2_profile` and the prefill sweep at all four lengths:

```
air.backend.abc.AirBackendError: aircc compilation failed:
loc("air.mlir":1:1): error: expected operation name in quotes
```

`conv1d_depthwise.build_module` ended `return launch`, handing back the `air.api` `LaunchContext`. `llms/lfm2_1_2b_q4nx/lfm2_1_2b_q4nx_prefill.py:238` imports it and `:326` passes the result to `KernelCache.compile_and_cache`, which writes `str()` of it into `air.mlir`. The file therefore held one line — `<air.api._compile.LaunchContext object at 0x...>` — which is what aircc rejects at 1:1.

Credit to the reporter for the diagnosis and the repro.

## The fix

Returning the context is the *majority* convention, not a deviation: ~78 examples end `build_module` with `return launch`, and that is fine for an example nobody imports. It is wrong for the handful `llms/` imports, which hand the result straight to the cache. Those four — `gelu_and_mul`, `layer_norm`, `silu_and_mul`, `weighted_rms_norm` — split `build_launch` (the context) from `build_module` (a built module, explicit target). `conv1d_depthwise` now matches them. The module text is unchanged.

## Two things around it, because the return type is only half the defect

**A guard at the call site.** `XRTBackend.compile` `str()`s whatever it is given, so a type error surfaced as a parse error two passes downstream in a file nobody reads. It now rejects anything that is not a module, and names the remedy when the argument has one:

```
AirBackendError: XRTBackend.compile expects an MLIR module, got
air.api._compile.LaunchContext. Call .build(target=...) on it first.
```

**A PR-gated test.** `python/test/api/llms_builder_contract.py` builds every cross-directory builder `llms/` imports and reparses each with `air.ir.Module.parse` — what aircc does with the file. #1927 could not have been caught otherwise: it touched no `llms/` file, and `check-programming-examples-llms-*` runs only in `nightlyPerfBenchmark`, so the PR was green and the nightly went red hours later attributed to nobody. Adding an import to `llms/` now means adding a row.

The test is proven to fail on the bug, not merely to pass on the fix: reintroducing the defect gives `NOT MLIR (LaunchContext, 1 lines)` and names the builder.

## The scan turned up a second instance

Running the consumer scan for the fix found the same defect in **#1944**, still open: `flash_attention/kernel_fusion_based/attn_npu2.build_module` is imported by `llms/shared/infra/fa_headfirst.py:129` and passed straight to `compile_and_cache`. On `main` it returns an `ir.Module`; the conversion in #1944 makes it return a `LaunchContext`. Fixed on that branch (module text byte-identical, 715 lines). The test in this PR passes on plain `main` and would have caught it.

`channel_examples/dual_herd_packet_switch`, converted in the same PR as `conv1d_depthwise`, also ends `return launch` — but nothing imports it and its own `main` calls `.build()`, so it is not broken and is left alone.

## Gating

- Reporter's repro: **1 line → 40 lines**, non-empty, reparses.
- Module text byte-identical to the pre-edit builder (41 lines; baseline asserted non-empty first).
- The real lfm2 path — `build_conv1d(512, 2048, 8, ...)` → `KernelCache.compile_and_cache('conv1d', ...)` → aircc at `target_device=npu2` — produces `conv1d.xclbin` (21593 B) + `conv1d.insts.bin`. Control with the defect reintroduced fails at the new guard.
- `check-air-python` 46 pass (was 45), `check-air-mlir` 525 pass / 7 XFAIL.

**Not run, stated plainly:** the `conv1d_depthwise` lit is `REQUIRES: ryzen_ai_npu2` and my box is npu1 (Phoenix), so there is no hardware run here; and no lfm2 `make verify` end to end, which needs LFM2 weights I do not have. The npu2 aircc compile above covers the parse failure, not execution.

### DIFF
--- a/programming_examples/conv1d_depthwise/conv1d_depthwise.py
+++ b/programming_examples/conv1d_depthwise/conv1d_depthwise.py
@@ -44,7 +44,7 @@ HALO = K_TAPS - 1  # 2 rows of left context
 L1_BYTES = 65536  # per-compute-tile L1 on AIE2P
 
 
-def build_module(seq, channels, tile_s, np_dtype_in=bfloat16, herd_x=8, herd_y=1):
+def build_launch(seq, channels, tile_s, np_dtype_in=bfloat16, herd_x=8, herd_y=1):
     assert (
         channels % herd_x == 0
     ), f"channels ({channels}) must be divisible by herd_x ({herd_x})"
@@ -169,6 +169,25 @@ def build_module(seq, channels, tile_s, np_dtype_in=bfloat16, herd_x=8, herd_y=1
     return launch
 
 
+def build_module(
+    seq, channels, tile_s, np_dtype_in=bfloat16, herd_x=8, herd_y=1, target="npu2"
+):
+    """The MLIR module. Signature and return type are the llms/ builders' contract.
+
+    Most air.api examples let `build_module` hand back the LaunchContext and
+    leave `.build()` to their own `main`, which is fine for an example nobody
+    imports. The four that llms/ imports — gelu_and_mul, layer_norm,
+    silu_and_mul, weighted_rms_norm — instead split the two: `build_launch` for
+    the context, `build_module` for a built module with an explicit target.
+    lfm2's prefill hands the result straight to `cache.compile_and_cache`, so it
+    needs the module; returning the context stringified an object repr into
+    air.mlir and failed as a parse error two passes downstream.
+    """
+    return build_launch(
+        seq, channels, tile_s, np_dtype_in, herd_x=herd_x, herd_y=herd_y
+    ).build(target=target)
+
+
 def conv1d_reference(x_pad, w_tapmajor, seq, channels):
     """FP32 reference for the causal depthwise conv.
 
@@ -268,7 +287,7 @@ if __name__ == "__main__":
     if args.perf_iters < 0:
         parser.error("--perf-iters must be >= 0")
 
-    launch = build_module(
+    launch = build_launch(
         args.seq,
         args.channels,
         args.tile_s,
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Ran this on NPU2 (Strix, AIE2P) with the LFM2 weights, which closes the two gaps the description states it could not cover.

Patch applied to `main`, `ninja install`, then the real `llms/lfm2_1_2b_q4nx` path:

```
Compiled conv1d: 0.3s -> conv1d.xclbin
All 7 kernels compiled to .../build_peano/kernel_cache_2048/
```

`conv1d` compiles through aircc at `target_device=npu2` and produces an xclbin — the `expected operation name in quotes` failure is gone on hardware, not just at the builder call.

Then `make compile-decode` (decode_L2048 + decode_L2047) and `make verify` end to end:

```
[verify] Summary: {'n_layer_records': 0, 'topk_passed': 2, 'topk_failed': 0}
[verify] PASS
```

That is the exact lit the nightly has been failing (`run_npu2_verify`), green.

One note for the ledger: this clears lfm2's three lits, which includes the prefill sweep that exits the job. The nightly stays red on the other three models — `qwen3_4b_q4nx` and `qwen25_3b_q4` decode timeouts, and `llama31_8b_q4nx` / `qwen25_7b_q4nx` top-k. Unrelated to this PR; I am working through those separately.

PR #1947, opened 2026-08-28
[air.api] Convert all of programming_examples/llms/ to air.api

erwei-xilinx:
Moves all twelve raw-bindings builders under `programming_examples/llms/` onto `air.api` and deletes the thirteen copy-pasted duplicates that had accumulated across the int4, qwen and smolvla trees. `llms/` now has no `@module_builder` left in it.

## What is gated, and how

Two instruments, because they answer different questions and only one was used before.

**`air.insts.bin` — the host DMA/BD stream.** Every builder A/B'd against its predecessor at every shape its call sites use: **65/65 byte-identical**, plus 14/14 on the shapes re-run for this PR. Each comparison asserts both arms non-empty and runs a same-input determinism control first, so an "identical" cannot come from two failed builds.

**Per-core ELFs — the core program code, which `insts.bin` cannot see.** This is the instrument the earlier batches were missing. Across all twelve builder families:

| family | `insts.bin` | core code |
|---|---|---|
| `rope_2d`, `rope_1d` | identical | **identical** |
| `rms_1d` (×3 shapes) | identical | differs |
| `add_2d_to_2d`, `add_2d_to_1d`, `padded_add` | identical | differs |
| `bias_add_2d`, `bias_add_1d` | identical | differs |
| `qknorm_1d`, `qknorm_2d` | identical | differs |
| `gelu_2d`, stitcher add | identical | differs |

`rms_1d`'s difference is a loop peel of the accumulator init — same 16-wide bf16 vectors, same summation order, one fewer add of zero. **Every other difference traces to `air-ping-pong-transform` firing on the DSL's emission where it did not on the predecessor's: the converted builders systematically gain double-buffering.** That is also why `insts.bin` is identical — ping-pong changes core-side buffering and objectfifo acquire/release, not the shim BD program.

Numerics are preserved as far as generated code can show: vector widths unchanged, f32 accumulates intact, per-row reduction structure unchanged, and both transformations are exact for the operations involved.

**Throughput is not measured.** Each builder trades roughly 20% more program memory and 2× the L1 tile buffers for DMA/compute overlap. Plausibly a win; unverified. The shipping configurations are `herd_x=8` and npu2-only, so they cannot be measured on an npu1 box — this needs `check-programming-examples-llms-profile` in nightly.

**The stitched multi-launch modules are gated too.** `qwen25_0_5b`'s `build_o_ffn_head_module` and `build_down_add_module` both run `stitch_elf`, whose regex surgery over `str(module)` is the part the text changes could have broken. Against `main` both are structurally identical — same op census, channel symbols, public and private functions — differing only in line count.

## The twelfth builder

`o_gemv_ffn_int4_fused` is the large one: one launch, one segment, ten herds, two W→E cascade chains, three packet broadcasts, 19 channels, 138 channel operations. `air.elf` is **not** byte-reproducible (a same-input control differs in 182 bytes of PDI header ids/checksums and the XRT UUID), so it is gated on `aie.air.mlir` and the per-core ELFs, which are. The whole data-movement fabric matches the predecessor exactly:

```
aie.tile 40   aie.core 24   aie.lock 198   aie.use_lock 632   aie.dma_bd 190
aie.shim_dma_allocation 34   aie.memtile_dma 8
aie.flow 33         -- identical SET
aie.packet_flow 19  -- identical id multiset
```

Only `aie.buffer` differs, +8, one core-local buffer per LGU/LD core.

**Its numerics are not verified.** The design needs npu2's eight columns; its lit is `REQUIRES: ryzen_ai_npu2, peano` and can only run in nightly.

## DSL surface

Three additions, all required by that builder and none useful without it: `channel_type="npu_dma_packet"`, `broadcast_shape` alongside a packet channel (previously refused for every type on a rationale that only holds for cascade), and `air.herd(at=(column, row))` for the `x_loc`/`y_loc` that a cascade chain depends on.

Plus one behavioural fix: **channel declarations are now placed by construction order rather than first-use order.** `air-to-aie` hands out shim DMA channels walking the `air.channel` symbols in order, so on a design near the shim's limits the order decides whether an L3 endpoint gets an allocation at all. Blast radius measured: of 26 converted examples with more than one channel, 18 are byte-identical and 8 reorder declarations only — same declaration set, rest of the module unchanged — and 7 of those 8 produce a byte-identical `air.insts.bin` (the 8th fails to compile in *both* arms for an unrelated missing kernel object).

## A bug this found

Replacing each model's private residual-add with a call to the shared builder left four qwen prefills calling `build_named_add`/`build_padded_add` with no import — nine undefined names, a `NameError` the moment the builder runs. Three things should have caught it and each had a hole: the name resolves at *call* time so an import sweep passes; `llms_builder_contract.py` covers the six cross-directory builders, not the models' own; and `llms/` is nightly-only, so the PR goes green and the nightly reddens on `main` hours later. Same shape as #1927.

Fixed, and `python/test/api/llms_shared_builder_calls.py` closes the gap in the PR-gated suite by calling every model-local builder that reaches into `shared/builders`. Proven by revert: dropping one import takes it from 9/9 to 7/9 and names both broken builders.

Also included: `Makefile_o_gemv_ffn_int4_fused` was the only one of four Makefiles compiling `mv_int4_bf16.cc` missing `-D__AIE_API_AIE_ADF_HPP__`, so its peano kernel build failed outright against the current wheel. Unrelated to the conversion, latent because that lit is nightly-only.

## Testing

- `check-air-python` api suite: 55 tests, all passing (54 before, +1 new)
- `llms_builder_contract`: 6/6
- `llms_shared_builder_calls`: 9/9, proven by revert
- `black` and `pyflakes` clean on all changed files


erwei-xilinx:
Addressed the reviewer's `_UNSUPPORTED['channel_type']` catch in 291bee01 — it was correct.

Widening the guard to admit `npu_dma_packet` left the sentence beside it still listing packet and cascade as unavailable, so anyone hitting the error path for `npu_mmio` got told the wrong thing. That's the same staleness twice over: the first pass fixed the "Only 'npu_cascade' is available" clause and missed this one. So rather than correct the string again, the list is now **derived** — `_ALL_CHANNEL_TYPES` minus `_IMPLEMENTED_TYPES` — and adding a type to the implemented tuple drops it from the refusal automatically.

```
air.channel(channel_type='npu_mmio') is not implemented. Implemented:
'npu_cascade', 'npu_dma_packet', plus the default npu_dma_stream, which is
what you get by leaving channel_type off. Still unimplemented: npu_mmio,
gpu_symmetric_heap. Each has its own lowering and its own verifier rules --
mmio, for instance, requires an L3 put, an L1 get and a constant
memref.get_global source.
```

While there: `dest=` and `packet_ids=` both said they "need npu_dma_packet", which now reads as the reason they're refused when the real reason is that those knobs aren't implemented. Reworded.

api suite still 55/55.

erwei-xilinx:
Confirming the `math.rsqrt` finding, and it localises to one line.

`o_gemv_ffn_int4_fused.py:549` binds `rstd` to an **expression**, not a value:

```python
rstd = ops.cast(ops.rsqrt(acc[:] / float(K) + 1.0e-5), bf16)
l1_normed[:] = l1_res1[:] * rstd * l1_gamma[:]
```

A `BufferExpr` is a tree, so the second line splices the whole rsqrt subtree into the elementwise assignment and it is evaluated per vector step rather than once. Reproduced the two lowerings side by side on a standalone RMSNorm at K=2048, VEC=16, on `main`s air.api:

```mlir
// inline -- rsqrt is vector<16xf32>, INSIDE the elementwise loop
%11 = arith.addf %9, %10   : vector<16xf32>
%12 = math.rsqrt %11       : vector<16xf32>
%13 = arith.truncf %12     : vector<16xf32> to vector<16xbf16>
%14 = arith.mulf %5, %13   : vector<16xbf16>
%15 = vector.transfer_read %alloc_2[%arg10] ...
vector.transfer_write %16, %alloc_3[%arg10] ...

// materialized -- rsqrt is scalar f32, computed once and stored
%7  = arith.addf %6, %cst_17 : f32
%8  = math.rsqrt %7          : f32
%9  = arith.truncf %8        : f32 to bf16
memref.store %9, %alloc_7[%arg10] : memref<1xbf16, 2 : i32>
```

At K=2048/VEC=16 that is 128 rsqrt lookups per core instead of 1, which is the right order for the ~7%. The predecessor did `math_dialect.rsqrt` once and then `BroadcastOp`, so this is a conversion artifact rather than a design change.

Worth noting the other three converted RMSNorm sites in this PR already do it correctly -- `rms_gemv_rope_multi.py:160`, `rms_qkv_qknorm_rope_multi.py:170` and `:235` all allocate `rstd` and assign `rstd[:] = ...` before using `rstd[:]`. Only this one drops the alloc and the subscript, so the fix is to match its siblings:

```python
acc  = air.alloc([1], f32,  scope=h.private())
rstd = air.alloc([1], bf16, scope=h.private())
acc[:]  = ops.reduce_add(ops.cast(l1_res1[:] * l1_res1[:], f32))
rstd[:] = ops.cast(ops.rsqrt(acc[:] / float(K) + 1.0e-5), bf16)
l1_normed[:] = l1_res1[:] * rstd[:] * l1_gamma[:]
```

I have not built this branch to confirm the emitted IR for the fused kernel itself -- that needs this PR's DSL installed, and my box is mid-run on two other model gates. The mechanism above is measured; the 7% recovery on this specific kernel is not yet.

erwei-xilinx:
Correction to my comment above: **the rsqrt is not the ~7%.** #1957 measures the gap on npu2 hardware, min-of-500, arms interleaved, and closes it entirely — 1599 us -> 1489 us against 1488–1498 for the raw bindings — **with this builder unmodified**. Line 549 still binds `rstd` to an expression in that arm, so it cannot be what was costing the 7%. The causes are `bf16.default_vector_width` being 16 on npu2 where the hardware legalizes 32, and `ops.fill` lowering to `linalg.fill` and thence a store per element.

What I got right was the mechanism: `math.rsqrt` really does land as `vector<16xf32>` inside the elementwise loop rather than as a scalar computed once, and the three sibling RMSNorm sites in this PR really do materialize where this one does not. What I got wrong was the attribution — I reasoned from op counts to a percentage instead of measuring, said "right order for the ~7%", and that inference was simply wrong. Sorry for the noise.

So the 5-line change I proposed is now a consistency fix rather than a performance fix: it makes line 549 match `rms_gemv_rope_multi.py:160` and `rms_qkv_qknorm_rope_multi.py:170`/`:235`, and it should be justified on that basis or dropped, not on recovering the regression. Worth re-measuring on top of #1957 before anyone spends time on it — with the loop at 32 lanes the inline rsqrt runs half as often again, and it may well be free.

PR #1949, opened 2026-08-29
[flash_attention] Restore the build_module contract, and temporal_causal's numbers

erwei-xilinx:
Hotfix for two defects #1944 left on main. Both are invisible to per-PR CI.

## 1. Three of four builders stopped returning a module

Only `attn_npu2` kept a module wrapper after the air.api migration — because
`llms_builder_contract.py` forces it. Type-checked by importing each:

| builder | returns | |
|---|---|---|
| `attn_npu2` | `Module` | safe |
| `attn_npu2_seqfirst` | `LaunchContext` | **breaks** |
| `attn_npu2_temporal_causal` | `LaunchContext` | **breaks** |
| `attn_npu1` | `LaunchContext` | **breaks** |

Six call sites hand the result straight to `KernelCache.compile_and_cache`,
which stringifies it into `air.mlir`, so each dies with
`XRTBackend.compile expects an MLIR module`:

- `llms/llama32_1b/llama32_1b_prefill.py` (both branches)
- `llms/llama32_1b_int4/llama32_1b_int4_prefill.py`
- `llms/qwen25_0_5b/qwen25_0_5b_prefill.py`
- `llms/lfm2_1_2b_q4nx/lfm2_1_2b_q4nx_prefill.py`
- `llms/shared/infra/fa_temporal.py`
- `llms/smolvla/smolvla_vision_encoder.py`

`llms/shared/infra/fa_headfirst.py` is the only safe one, because it imports
the wrapped `attn_npu2`.

Models reached, directly or through `llama32_1b_prefill`: **llama32_1b,
llama32_1b_q4nx, llama32_1b_int4, llama32_3b, smollm2_1_7b, qwen25_0_5b,
lfm2_1_2b_q4nx, smolvla**. Not reached: gemma3_4b_q4nx, qwen3_0_6b/1_7b/4b,
qwen25_1_5b, qwen25_3b — all go through `fa_headfirst`.

Fixed at the source rather than at six call sites, exactly the way `attn_npu2`
already does it: rename the air.api entry to `build_launch` and add a
`build_module(**kwargs)` that returns `build_launch(**kwargs).build(target=...)`.
**No call site changes.**

## 2. temporal_causal is numerically wrong

Worse, because it is silent and it is q4nx's default (`TEMPORAL_CAUSAL_SKIP := 1`
is exported by its Makefile, and it is the fast path). `llama32_1b_q4nx`
prefill returns token **499** instead of **12366 (' Paris')**.

Bisected by swapping only `attn_npu2_temporal_causal.py`, with the compiler
build, the mlir-aie pin and the weights all held fixed and the kernel cache
cleared each run:

| `attn_npu2_temporal_causal.py` | result |
|---|---|
| post-#1944 (906 lines) | argmax **499**, MISS |
| 38d00992 (2009 lines) | argmax 12366, **PARIS** |
| seqfirst instead | argmax 12366, **PARIS** |

So `seqfirst`'s migration is numerically fine and this one is not. Reverted to
38d00992's implementation, which is `@module_builder` and therefore satisfies
the contract too. Re-migrating it is left to a follow-up that can gate on
`verify-paris`.

## 3. Why the existing guard missed it

`llms_builder_contract.py` caught this exact class for #1927 and is the right
mechanism — it needs no NPU, so it gates per-PR. It only missed #1944 because
its list did not include the FA builders `llms/` imports. Added the three
missing rows (**9/9**).

This matters because `buildAndTestRyzenAI.yml` runs the programming examples
with `--filter-out "llms/"`, and `check-programming-examples-llms-*` runs only
from `nightlyPerfBenchmark.yml` (cron 04:17) — which was already red on the
27th and 28th on SHAs predating #1944, so this would have landed on top of
existing redness.

## Verification

- `verify-paris` **PARIS on the default path** (no env override) and on
  `TEMPORAL_CAUSAL_SKIP=0`.
- `llms_builder_contract.py`: 9/9 builders return a module.
- `smolvla make compile`: 5 ELFs, with **no** call-site change.
- Examples: `attn_npu2`, `attn_npu2_seqfirst`, causal, gqa, elf — all PASS.
- `check-air-python`: 43 passed, 0 failures.

## Note for #1936

#1936 currently works around this at smolvla's two call sites with
`.build(target="npu2")`. Once this lands those become wrong (they would call
`.build()` on a `Module`), so I will drop them there — this PR makes them
unnecessary.


erwei-xilinx:
Superseded by #1950, which has the identical four-file change on a clean branch.

I built this branch with `git add -A <path>`, which swept in a large amount of untracked debris under `kernel_fusion_based/` — investigation .md notes, .csv, IR dumps, `debug_ir/`, ad-hoc test scripts. That is what the format check caught: every file it flagged (`test_cascade/test_ir.py`, `test_precision/`, `python/spensor/**`) is one I never edited. My mistake, not a problem with the fix.

#1950 is `origin/main` + exactly four files:
`attn_npu1.py`, `attn_npu2_seqfirst.py`, `attn_npu2_temporal_causal.py`, `llms_builder_contract.py`.

PR #1952, opened 2026-08-29
[air.api] Give a buffer region the operators a whole-buffer read already has

erwei-xilinx:
`Buffer.__getitem__` short-circuits: `b[:]` returns a `BufferExpr`, anything
narrower returns a `BufferSlice`. Arithmetic was defined on both; comparisons
and bitwise on `BufferExpr` only. So two spellings of the same read answered
different operators:

```python
b[:]   >= n    # works
b[3:4] >= n    # TypeError: '>=' not supported between instances of
               #            'BufferSlice' and 'int'
```

## Why it matters now

`attn_npu2.py:551` is the second spelling:

```python
head_next = ctr[2:3] + 1
wrapped   = head_next >= num_head_groups      # fine -- `+ 1` promoted it
if dv_chunks > 1:
    last_dv = ctr[3:4] >= dv_chunks - 1       # TypeError at trace time
```

`dv_chunks > 1` needs `head_dim > lkp`, so this branch is reached only by
**qwen3-4b, llama31-8b, qwen25-7b and llama32-3b** — every `fa_headfirst` model
with a 128-wide head. On main they fail to build:

```
File ".../flash_attention/kernel_fusion_based/attn_npu2.py", line 551, in _
    last_dv = ctr[3:4] >= dv_chunks - 1
TypeError: '>=' not supported between instances of 'BufferSlice' and 'int'
make: *** [.../qwen3_4b_q4nx/Makefile:235: compile-prefill] Error 1
```

Found by running the `qwen3_4b_q4nx` verify gate locally. It is a **new** break
relative to the last nightly, which got as far as decode.

## The fix

In air.api, not in the example — `attn_npu2` was written against the surface
this restores. The comparison and bitwise dunders forward through `_as_leaf()`
exactly as `_arith` already does.

`__eq__`/`__ne__` stay out, as on `BufferExpr`: defining `__eq__` sets
`__hash__` to `None`. Python derives the reflected *comparisons* by swapping
operands, so only the bitwise ones need `__r*__` forms. `__bool__` comes along
too — a region now answers the very operators that guard's message points at.

## Why no test caught it

Every comparison and bitwise test in the suite subscripts the **whole** buffer
(`a[:] >= b[:]`), so all of them take the `BufferExpr` path and none take
`BufferSlice`. `cmp_select.py` gains the partial-region cases: region-vs-region,
region-vs-scalar (the spelling that failed), the reflected form, and bitwise.

Note `a_comparison_can_name_one_element_of_a_buffer` already covers
`ops.equal(ctr[1:2], 0)` — the *function* spelling had been taught about
regions; the *operator* spelling had not. That is the asymmetry this closes.

## Verification

- `check-air-python`: **43 passed, 0 failed** (54 discovered, 11 unsupported).
- `qwen3_4b_q4nx`: `make compile` now completes, prefill included; it raised the
  TypeError before.
- Trace-built all four migrated `flash_attention/kernel_fusion_based` builders
  across the branch-selecting configs — head_dim 64/128/256, causal on/off, GQA
  2:2 and 4:1, `fused_qkv`, `n_images`, and 1/2/4/5/8/9 q-rounds (which select
  the triangle, merge, fold and row-split branches). **24 build**; the remainder
  are L1-budget rejections that say so. This was the only trace-time defect in
  the matrix.
- `black` clean.

### DIFF
--- a/python/air/api/_value.py
+++ b/python/air/api/_value.py
@@ -757,6 +757,70 @@ class BufferSlice(_StridedView):
     def __neg__(self):
         return -self._as_leaf()
 
+    # Comparisons and bitwise, completing the operator surface a region shares
+    # with a whole-buffer read.
+    #
+    # `Buffer.__getitem__` returns a BufferExpr for `b[:]` and a BufferSlice for
+    # anything narrower, so these two spellings of the same read went down
+    # different paths. Arithmetic was on both; comparisons and bitwise were on
+    # BufferExpr only. `b[:] >= n` therefore worked and `b[3:4] >= n` raised
+    # `TypeError: '>=' not supported between instances of 'BufferSlice' and
+    # 'int'` -- a distinction with no rule behind it, which every existing
+    # comparison test missed because they all subscript the whole buffer.
+    #
+    # __eq__/__ne__ stay out, as they do on BufferExpr: defining __eq__ sets
+    # __hash__ to None and changes what `slice == slice` means for ordinary
+    # Python. ops.equal / ops.not_equal are the spelling.
+    def __lt__(self, o):
+        return self._arith(o, lambda a, b: a < b)
+
+    def __le__(self, o):
+        return self._arith(o, lambda a, b: a <= b)
+
+    def __gt__(self, o):
+        return self._arith(o, lambda a, b: a > b)
+
+    def __ge__(self, o):
+        return self._arith(o, lambda a, b: a >= b)
+
+    def __and__(self, o):
+        return self._arith(o, lambda a, b: a & b)
+
+    def __rand__(self, o):
+        return self._arith(o, lambda a, b: a & b, reflected=True)
+
+    def __or__(self, o):
+        return self._arith(o, lambda a, b: a | b)
+
+    def __ror__(self, o):
+        return self._arith(o, lambda a, b: a | b, reflected=True)
+
+    def __xor__(self, o):
+        return self._arith(o, lambda a, b: a ^ b)
+
+    def __rxor__(self, o):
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Update after landing #1951 and re-running the gate end to end.

**qwen3_4b_q4nx `make verify` now PASSES on this branch** — `{"topk_passed": 2, "topk_failed": 0}`, `[verify] PASS`. Not just the compile: last night's `RuntimeError: decode dispatch pos6 state=ert_cmd_state.ERT_CMD_STATE_TIMEOUT` is gone too, which I attribute to the mlir-aie bump in #1946 (mlir-aie#3622, "Keep the empty-PDI reset alternating across dispatches") rather than to anything here — an odd empty-PDI reset count leaves the firmware caching the address so every dispatch after the first is un-reset, which is exactly a timeout at a random decode position.

So for tonight's nightly, of last night's 12 failures:

| cause | models | resolved by |
|---|---|---|
| conv1d LaunchContext | lfm2 verify/profile/prefill-sweep | #1945 |
| `o_ffn` memtile over-allocation | llama32_1b_q4nx, llama31_8b_q4nx, qwen25_7b_q4nx | not reproducible on the current pin — llama32_1b verify PASSES locally |
| decode timeout | qwen25_3b_q4 | #1943 (diagnoses this exact placement) |
| decode timeout | qwen3_4b_q4nx | verified PASS locally |

and smolvla, which #1951 just un-broke.

This PR is the remaining blocker: without it those four head_dim=128 models fail *earlier* than last night, at prefill compile instead of decode. Merged main in rather than rebasing.

PR #1966, opened 2026-08-31
[flash_attention] Gate the drain de-tiling at tile_size_q != dv_tile

erwei-xilinx:
The drain's de-tiling `reshape(dv_tile // M, tile_size_q // M, M, M)` is a no-op
whenever those two extents are equal, which is every shape that ships except
head_dim=256. That is how the swapped version reached main twice -- #1961 fixed
it in `attn_npu2_temporal_causal.py`, #1964 in the other three drains -- with a
green CI both times.

Nothing in the suite emits the asymmetric shape without a device: gemma3-4b is
the only model at head_dim=256, and its lit builds and runs the whole decode.

This adds an IR-only lit that emits `attn_npu2.py` at `fa_headfirst`'s
head_dim=256 tiling (lkp=32, lqp=256, num_q_tiles=8 so tile_size_q=32,
dv_tile=128) and FileChecks the `@Gp2L2` put:

    [4, 8, 16, 8] [64, 8, 256, 1]

Reaching that tiling from the CLI needs `num_q_tiles`, which `main()` hardcoded
to 4; it becomes `--num-q-tiles` with the same default, so no existing
invocation changes.

Verified both ways: passes on main, fails with the pre-#1964 reshape restored.

### DIFF
--- a/programming_examples/flash_attention/kernel_fusion_based/attn_npu2.py
+++ b/programming_examples/flash_attention/kernel_fusion_based/attn_npu2.py
@@ -614,6 +614,7 @@ def parse_args():
     parser.add_argument("--lkp", type=int, default=64)
     parser.add_argument("--lq", type=int, default=512)
     parser.add_argument("--lqp", type=int, default=256)
+    parser.add_argument("--num-q-tiles", type=int, default=4)
     parser.add_argument("--dk", type=int, default=64)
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Closing: this is a regression guard, not a fix for a failing test. Reopen if the coverage is wanted separately.

PR #1967, opened 2026-08-31
[ci] Record the runner's XRT, driver and firmware in perf.json

erwei-xilinx:
`perf.json`'s `toolchain` block names mlir-air, mlir-aie and Peano and nothing
beneath them. So when a failure reproduces only on the benchmark runner --
`qwen3_4b_q4nx`'s `ERT_CMD_STATE_TIMEOUT` is the live example -- there is no way
after the fact to tell whether XRT, the amdxdna driver or the NPU firmware moved
between a green night and a red one. Twice in the last week that question was
the one blocking a root cause, and twice it was unanswerable.

Adds, parsed best-effort from `xrt-smi examine`:

    "xrt_version": "2.23.0",
    "xrt_hash": "03f75c0f...",
    "amdxdna_version": "2.23.0_20260128",
    "npu_firmware": "1.1.2.64",
    "kernel": "6.14.0-37-generic"

No xrt-smi, or an output format change, leaves the fields out rather than
failing the publish -- provenance should never be able to break the run that
produces the numbers.

The block was written out four times (perf.json, the null-row path, sweep.json,
prefill_sweep.json); it becomes one dict built once.

Parser checked against real `xrt-smi examine` output.

### DIFF
--- a/.github/workflows/nightlyPerfBenchmark.yml
+++ b/.github/workflows/nightlyPerfBenchmark.yml
@@ -433,12 +433,44 @@ jobs:
             exit 1
           fi
           LLVM_AIE_VERSION=$(python3 -m pip show llvm-aie | awk '/^Version:/{print $2}')
+          # Runtime provenance under the pinned compilers: a decode hang that
+          # only ever reproduces on this runner is unattributable after the fact
+          # without it. Best effort -- a missing xrt-smi just leaves the fields out.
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Closing: instrumentation only, fixes nothing. Reopen if the provenance is wanted separately.

PR #1970, opened 2026-08-31
[air.backend] Recognize Gorgon Point as npu2

atassis:
## Problem

`xrt-smi` reports this part as `NPU Gorgon Point 1`, which contains none of the npu2 keywords, so the model list falls through. The list is duplicated in three places, kept in sync only by a comment, and all three miss it:

    python/air/backend/xrt.py        NPU_MODELS -> detect_target_device()
    test/lit.cfg.py                  npu2_models
    programming_examples/lit.cfg.py  npu2_models

They fail differently, which is why fixing one would not be enough:

- `detect_target_device()` falls back to npu1, and its own docstring names the consequence -- an `aie.device(npu1)` binary loads on npu2 without error and computes nothing.
- The lit configs take the "no recognized model" branch instead, so neither `ryzen_ai` nor `ryzen_ai_npu2` is added and **every NPU test reports UNSUPPORTED**. That is the same outcome those files' own comments record having already fixed once for Strix Halo and Krackan.

The silicon is npu2. This box is `1022:17f0 rev 10`, which `amdxdna_pci_drv.c` maps to `dev_npu4_info` -- already in the list as `npu4`/`Strix`. Only the marketing name is unmatched.

## Fix

Add the Gorgon Point name to all three lists, mirroring the same addition in mlir-aie's `hostruntime.py`, which the comments above each list already name as the alignment target.

## Test

Verified on the part against this box's real `xrt-smi` output, using the pre-patch lists as controls rather than by inspection:

- `detect_target_device()`: npu1 -> npu2 (the pre-patch arm also prints its own fallback warning)
- lit match: UNKNOWN -> npu2

`black` clean on all three files.

## Known limitation

The three lists remain duplicated. Having the lit configs import the backend's list instead of re-declaring it is the durable fix; it is a larger change than this one and I have kept it out of scope.

### DIFF
--- a/programming_examples/lit.cfg.py
+++ b/programming_examples/lit.cfg.py
@@ -123,7 +123,15 @@ if config.xrt_lib_dir and config.enable_run_xrt_tests:
         # aligned with mlir-aie's hostruntime.py and the AIR XRT backend
         # (NPU_MODELS). Robust to xrt-smi table-format changes and covers newer
         # parts (Strix Halo, Krackan) that the old strict regex missed.
-        npu2_models = ["npu4", "strix", "npu5", "strix halo", "npu6", "krackan"]
+        npu2_models = [
+            "npu4",
+            "strix",
+            "npu5",
+            "strix halo",
+            "npu6",
+            "krackan",
+            "gorgon point",
+        ]
         npu1_models = ["npu1", "phoenix"]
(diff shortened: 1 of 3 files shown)


erwei-xilinx (approved):
Thanks for looking into this, @atassis! Looks good to me.

PR #1975, opened 2026-09-02
[air-runner] Fix two stalls in channel retirement and tile admission

erwei-xilinx:
Two independent stalls in `air-runner`, each a case where the simulator reports
"simulation stalled" instead of a latency. Both were found by running a design
whose herds occupy more than one tile and whose herds on one hierarchy ask for
more tiles than it has.

### 1. A `channel.put` must retire on its own dispatch count

A put asked the *channel-wide* dispatch counter whether it had finished
dispatching its spatial instances. That counter is shared with the get, and the
get zeroes it on completion so the next round can start — so whenever the get
was executed first, the put read zero against its own total and could never
retire.

The op's own count is also the more precise question: `processed == total_count`
asks whether the channel has moved *this op's* worth of instances, which is only
the same question while a single put owns the channel.

### 2. Contended tiles are offered in a deterministic order

A herd whose input arrives on a channel read inside its own region is a graph
root: nothing in the dependence graph orders it against the other herds on the
same hierarchy, so they all ask for their tiles at once. Candidates are
collected by walking the already-processed vertices and taking their
successors, so the offer order is predecessor-completion order — well defined
across time, and arbitrary within a step.

That does not matter while every candidate fits. Once they do not it decides
everything, because a herd reserves its tiles when it is admitted and holds them
until its terminator runs. A herd that cannot start still occupies the
hierarchy, so a consumer offered the tiles ahead of its producer parks exactly
the tiles the producer is waiting for.

Two rules, and the pair is one mechanism:

- **break ties by ascending vertex id.** Ids follow the IR walk, so that is
  program order, which within a hierarchy is a topological order of the
  dataflow — producers first. Ties only: the existing order already carries
  completion time and that has to be preserved, because ranking purely by
  position starves the tail of a pipeline (the first stage of the next
  iteration is always earlier in the IR than the last stage of this one).
- **no bypass.** Once a herd has been refused its tiles, stop offering them to
  the herds behind it for the rest of the pass.

### Tests

Three new tests, and each of the three changes has one that fails without it:

| reverted | `channel_put_retire` | `candidate_order` | `no_bypass` |
|---|---|---|---|
| nothing (this PR) | pass | pass | pass |
| put retirement | **stall 209** | pass | pass |
| tie-break | pass | **stall 209** | pass |
| no-bypass | pass | **stall 415** | **stall 415** |
| all (main) | stall | stall | stall |

Each test documents what is load-bearing about its shape, because in every case
the fault hides under a small perturbation:

- `channel_put_retire` needs a priced op before the put (or the two sides start
  together and the put is executed first) and a herd extent within the port
  count (or the transfer takes several rounds and the order stops mattering).
- `candidate_order` needs each herd to wait on its own allocation with the
  allocations emitted in **reverse** stage order, which is what makes
  predecessor-completion order put consumers first. In stage order the offer
  order is already correct and the test passes for the wrong reason.

`ninja check-air-mlir`: 550/550. `git clang-format --diff origin/main` clean.
No behaviour change to any pre-existing test.

### DIFF
--- a/mlir/lib/Util/Runner.cpp
+++ b/mlir/lib/Util/Runner.cpp
@@ -15,6 +15,7 @@
 #include "llvm/ADT/APInt.h"
 #include "llvm/ADT/Any.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/ErrorOr.h"
 #include "llvm/Support/JSON.h"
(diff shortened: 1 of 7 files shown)


erwei-xilinx:
Thanks — addressed in d18284dd.

On the specific claim: `operator[]` could not actually have inserted there, because every key in `next_vertex_set` was written by the `try_emplace` loop on the line immediately above the sort. But the objection is fair as a matter of construction — a comparator that *can* mutate the container it reads is held up only by an adjacent loop, and if that loop ever moved or stopped covering every key it would be an inconsistent comparator rather than a visible bug.

So the map is out of the comparator entirely. The key is materialised into the sort input, using the iterator `try_emplace` already returns:

```c++
llvm::SmallVector<std::pair<uint64_t, Graph::VertexId>> by_age;
by_age.reserve(next_vertex_set.size());
for (auto v : next_vertex_set)
  by_age.emplace_back(c.first_ready_time.try_emplace(v, time).first->second, v);
llvm::sort(by_age);
for (auto [i, entry] : llvm::enumerate(by_age))
  next_vertex_set[i] = entry.second;
```

Same ordering — lexicographic on `(ready time, id)` is oldest-ready-first with an ascending-id tie-break — and one map lookup per candidate instead of one per comparison.

Re-verified: 550/550 `check-air-mlir`, `git clang-format --diff origin/main` clean, and the revert matrix is unchanged, so all three tests still fail without their respective fix.

| reverted | `channel_put_retire` | `candidate_order` | `no_bypass` |
|---|---|---|---|
| nothing (PR head) | pass | pass | pass |
| put retirement | stall 209 | pass | pass |
| tie-break | pass | stall 209 | pass |
| no-bypass | pass | stall 415 | stall 415 |

erwei-xilinx:
CI caught a real regression — `airrunner/matmul` and `airrunner/matmul_bf16_vectorized` stalled. Fixed in 74516a77, and rebased onto current main while I was there.

**Cause.** The candidate reordering was applied to the *whole* candidate list, but its justification only concerns herds. Sorting everything also changes which of several transfers is offered a port first, which is a behaviour change with no argument behind it — and `airrunner/matmul` has exactly one herd, so the deadlock this PR fixes cannot arise there at all and the design had no business being perturbed.

**Fix.** Only the herd slots are permuted now. Every non-herd candidate keeps its position and its relative order:

```c++
llvm::SmallVector<unsigned> herd_slots;
llvm::SmallVector<std::pair<uint64_t, Graph::VertexId>> by_age;
for (auto [i, v] : llvm::enumerate(next_vertex_set)) {
  if (!isa_and_present<air::HerdOp>(G[v].op))
    continue;
  herd_slots.push_back(i);
  by_age.emplace_back(c.first_ready_time.try_emplace(v, time).first->second, v);
}
if (herd_slots.size() > 1) {
  llvm::sort(by_age);
  for (auto [k, slot] : llvm::enumerate(herd_slots))
    next_vertex_set[slot] = by_age[k].second;
}
```

Tiles are the only resource a herd reserves ahead of running, so that is the whole scope of the rule. The head-of-line rule was already herd-only.

**Verification.** Beyond the suites: I ran `airrunner/matmul` with this branch's runner and with `origin/main`'s on identical input and diffed the emitted trace — **byte-identical**, 110592 bytes each. For any design without two herds contending on one hierarchy the change is now inert by construction, and that is checkable rather than argued.

The revert matrix is unchanged, so nothing was weakened to get there:

| reverted | `channel_put_retire` | `candidate_order` | `no_bypass` |
|---|---|---|---|
| nothing | pass | pass | pass |
| put retirement | stall 209 | pass | pass |
| tie-break | pass | stall 209 | pass |
| no-bypass | pass | stall 415 | stall 415 |

550/550 `check-air-mlir`, clang-format clean.

One note for anyone reproducing: `check-air-runner` is a separate target from `check-air-mlir`, and I had only been running the latter. That is on me — the two airrunner tests live in `test/airrunner/`, not `mlir/test/`.

erwei-xilinx:
Fixed properly in 4a8773a7. The previous push did not cure it and I had the cause wrong; both are worth writing down.

### Why I could not reproduce it

`test/airrunner/*` calls `air.compiler.util.Runner(...)`, which is the runner compiled into the **python extension** — not the `air-runner` binary. I had only been rebuilding the binary, so every local "main vs branch" comparison ran the same stale extension and gave identical results, including identical stalls. Rebuilding `AirAggregateCAPI` as well gives a faithful local reproduction: `origin/main` passes both tests, the previous branch head failed both. Everything below is measured against that.

### The actual cause

Retiring a put on **its own** dispatch count. That reads like the more precise question and is wrong, because the dispatch budget is channel-wide as well — `getRemainingDispatchesForDynamicDispatch` subtracts the *channel's* progress:

```c++
unsigned remaining = total - already_dispatched;   // already_dispatched is channel-wide
```

So where several puts share a channel, a later one legitimately dispatches fewer instances than its own spatial factor, sometimes none at all, and asking its own count leaves it waiting for instances it was never going to send. Both `airrunner/matmul` tests are exactly that shape — **two put ops on ten of their twelve channels**.

### The fix

Keep the channel-wide test and add only the case the get side has always had: the channel completed and its counters were cleared before this op was executed. Zero is indistinguishable from "not started" on the counter alone, so it is qualified by this op having dispatched something — a put that has moved nothing cannot be looking at its own completion.

```c++
if (processed == total_count ||
    (!processed && this->ctrl_g->g[it].dispatched_count)) {
  this->retireChannelOp(it);
}
```

The ordering change also got narrowed in the previous push (herd slots only, non-herd candidates keep position and relative order); that part was right and is unchanged here.

### Verification

Full matrix, python extension rebuilt for every row:

| variant | `matmul` | `matmul_bf16` | `channel_put_retire` | `candidate_order` | `no_bypass` |
|---|---|---|---|---|---|
| PR head | PASS | PASS | PASS | PASS | PASS |
| minus put retirement | PASS | PASS | stall 209 | PASS | PASS |
| minus tie-break | PASS | PASS | PASS | stall 209 | PASS |
| minus no-bypass | PASS | PASS | PASS | stall 415 | stall 415 |
| `origin/main` | PASS | PASS | stall | stall | stall |

Every pre-existing test passes in every variant, and each of the three changes is isolated by a test that fails without it. 550/550 `check-air-mlir`, clang-format clean.

PR #1980, opened 2026-09-04
[air-to-aie] Bump mlir-aie + llvm-aie, and size the AIE core stack for LLVM 22

erwei-xilinx:
`llvm-aie` has been held at a 21.x pin because the LLVM 22 nightly produced NaN / garbage outputs on `xrt/48_triton_matmul_ver4_strix_4x4_bf16_output`, `xrt/50_multi_launch_attention` and the Qwen LLM examples. The pin comments carried a revert checklist for those tests.

That was never a compiler bug. LLVM 22 grows the AIE core stack frame for these kernels from 960 to 1216 bytes (`paddxm [sp], #0x3c0` -> `#0x4c0` in the core ELF prologue). `air-to-aie` only emitted `stack_size` when the pass option differed from its default, so at the default the attribute was absent, the core fell back to the device default of 1024 bytes, and the frame overflowed into that core's own L1 buffers.

## Changes

- Emit `stack_size` unconditionally in `air-to-aie`, and raise the default to 2048, so the size AIR compiles against is explicit rather than inherited from the target model. The default moves in all three places that define one (`tools/aircc/aircc.cpp`, `Passes.td`, `python/air/backend/xrt{,_runner}.py`) — the Python layer only forwards `--stack-size` when it differs from its own default, so these have to stay in step.
- Bump mlir-aie to `10767b5` (wheel `1.4.3.dev55+g10767b5`) and llvm-aie to `22.0.0.2026090201+a36c62b9`, which is the llvm-aie that mlir-aie itself pins at that hash. mlir-aie's `clone-llvm.sh` is unchanged across the window, so no LLVM bump is needed.
- `RuntimeSequenceOp` gained optional `trace_buffer` / `trace_slices` attributes in mlir-aie #3652, so the two builders in `AIRRtToNpuPass.cpp` need the wider arity.
- Drop `aie2_fp32_conf_shim.h` from `block_quantized_i8/single_core`. llvm-aie 21 only *declared* `mul_elem_16_conf` / `mac_elem_16_conf` for aie2 (`aiev2_aie_api_compat.h`), so the example force-included a header supplying the bodies to avoid undefined symbols; llvm-aie 22 *defines* them in `aiev2_core.h`, which turns the shim into a redefinition error.
- Drop the now-stale revert checklists from the three pin sites.

The newer mlir-aie is what identified the stack bug: it measures each core's requirement and reports the shortfall by name (`stack_size = 1024 is insufficient: this core needs 1216 bytes`), turning this class of bug into a build error instead of silent corruption.

## Validation

NPU2 (Strix), all runs serialised through `lit`'s `flock /tmp/npu.lock` on an otherwise idle machine.

Three arms, interleaved run-by-run, N=10 each, single variable (mlir-aie held at the old pin):

| arm | xrt/48 | xrt/50 |
|---|---|---|
| Peano 21 / stack 1024 | 10/10 PASS | 10/10 PASS |
| Peano 22 / stack 1024 | 10/10 FAIL | 10/10 FAIL |
| Peano 22 / stack 2048 | 10/10 PASS | 10/10 PASS |

Full `test/xrt` suite: Peano 22 alone regresses exactly xrt/48 and xrt/50 and nothing else; with the stack fix there are zero new failures relative to the old pin. On the combined bump the suite is 67 passed / 1 failed, and that one (`xrt/57_l3_broadcast_packet_flow`, `'aie.dma_bd' op Packet ID field can only hold 5 bits`) fails identically on the old pin — pre-existing and untouched here.

`check-air-mlir`: 536 passed / 0 failed.

The bump is performance-neutral on its own: isolated `make bench-decode` (llama-1B, n=40, sd <= 0.05 ms) moved L=128 17.090 -> 17.067 ms, L=1024 17.545 -> 17.494, L=2048 18.080 -> 18.029.

Cost of the default: 1 KB of L1 per core, and buffer addresses shift by the stack delta. The full suite plus the MLIR gates show no design that stops fitting.

### LLM gates

Both run on NPU2 against the combined bump, via the same `make verify` top-k token gate (NPU vs HF bf16, 2 prompts x 32 greedy tokens, k=5) that the nightly uses:

| example | result | wall time |
|---|---|---|
| `llms/llama32_1b_q4nx` | `[verify] PASS` — topk_passed 2, topk_failed 0 | 33 s |
| `llms/qwen25_0_5b` | `[verify] PASS` — topk_passed 2, topk_failed 0 | 89 s |


erwei-xilinx:
Good catch — fixed in abcdbdc0.

`setStackSize` was inside the `if (!core)` branch, so a tile that already had an `aie.core` kept the device default. Moved the call out of that branch and guarded it on the attribute being absent, so a size already set on the core still wins:

```cpp
      // Size every core this pass touches, not just the ones it creates: the
      // device default is smaller than the frames Peano now generates, so
      // leaving it absent overflows the stack into this core's buffers. A
      // size already on the core was chosen deliberately, so it wins.
      if (!core.getStackSizeAttr())
        core.setStackSize(options.stack_size);
```

`getStackSize()` is `std::optional<int32_t>`, so the null-check is well defined.

Behaviour is unchanged for cores this pass creates — the attribute is set either way, only later in the sequence, and `aie.core` attributes serialize alphabetically so the emitted IR is identical. Re-verified: `check-air-mlir` 536 passed / 0 failed, and xrt/48 + xrt/50 (the two tests the stack fix targets) pass on NPU2.

PR #1981, opened 2026-09-04
[gemma4_e2b_q4nx] Gemma4-E2B on NPU2: a per-layer-embedding decode engine, gated per layer class

erwei-xilinx:
Adds Gemma4-E2B (text) as a Q4NX example, running FastFlowLM's `model.q4nx`
bundle on NPU2 through a new fused-decode engine that carries the per-layer
embedding branch the model needs.

**Scope: the decoder layer, gated per layer on device.** There is no prefill
engine and no token loop yet, so this example deliberately does not carry the
`run` / `ask` / `verify` / `profile` targets the mature `llms/` examples do,
and its `make help` says so. What is here is verified rather than sketched.

## Correctness

Coverage is **by layer class**, not by sample. Gemma4 crosses two attention
types with two KV/FFN regimes, and a layer from one class says nothing about
the others:

| layer | class | cos |
|---|---|---|
| 0 | sliding, own kv, 6144 | 0.999385 |
| 4 | full, own kv, 6144 | 0.998385 |
| 14 | full, own kv, 6144 | 0.992663 |
| 15 | sliding, kv←13, 12288 | 0.997008 |
| 19 | full, kv←14, 12288 | 0.991631 |
| 34 | full, kv←14, 12288 | 0.999172 |

The reference is numpy, itself gated per layer against FastFlowLM's own golden
activations (all 35 layers 0.986–0.9998). Every figure above is bit-stable
across repeats and was re-measured on the exact pinned toolchain
(mlir-aie `10767b50e`, llvm-aie `22.0.0.2026090201+a36c62b9`).

The last 20 layers share KV — they carry no k/v projection and attend the cache
of the last layer of their own type below the boundary — so scoring them needs a
two-layer build. `DECODE_KV_SRC` points a wave's readback at a lower slab; the
append is deliberately left on the wave's own slab (see below).

## Four bugs, with the evidence in the commit messages

**An AIR DMA channel split into more than one BD task runs once per PDI load,
not once per wave.** `getRepeatCounts` opens a task whenever the enclosing loop
trip count changes, and only a single-task channel gets the self-looping form;
with two or more, every chain ends at `aie.end` and nothing re-arms it. mlir-aie
does not diagnose this — `HasValidDMAChannels` collects channels into a
`DenseSet` and only checks its size, so duplicates are silently deduped. This
one fact explained a strict pass/fail alternation, a hard ~50% ceiling, and why
a fresh `hw_context` always passed (it reloads the PDI). Fixed by giving every
PLE channel exactly one task.

**`chess_storage(%chess_alignof(...))` is Chess-only and Peano ignores it**, so
`activation_lut_ab/cd` sat at alignment 4 while the exp LUTs next to them
carried a real `alignas`. `aie::lut` reads with vector loads, so whether
`getActivationBf16` worked was decided by where the linker landed the table —
correct on one core at `0x…400`, garbage on another at `0x…04`. This was never
gemma4-specific: `glu.cc` uses the same helper on every model and was aligned
only by luck.

**A dispatch that follows an NPU TDR completes but returns a wrong answer**,
reproducibly per layer, and is cured by neither re-uploading the inputs nor a
fresh `hw_context`. The gate now reports INCONCLUSIVE and exit 2 rather than
scoring it; a six-layer sweep had reported four FAILs that way, none real.

**A vocab-drain stall was the up core's placement.** Bisecting the packet
channel by destination isolated it to the one core whose packets traverse the
other two PLE tiles' switches; moving it one row clear went 32/40 → 80/80,
numerically identical. Every scheduling lever was re-measured on the fixed build
and none moved it.

Two of these are in shared code and fix other models too. The 1x4x1 attention
fixes also affect gemma3-4b, whose top-k verify gate was run on this branch and
**PASSES** (2/2 vs the bf16 reference).

## Also here

- `fused_decode_ple` gets a Makefile and README; the kernel recipe was previously
  not in the tree at all. Its `preflight-peano` checks for llvm-aie ≥ 22 because
  21.x does not fail to build this model — it builds it and returns all-NaN.
- A `peano_ge22` lit feature, for the same reason: a test gated only on `peano`
  would go green on a design that cannot compute anything.
- The example is registered in `hf_models.txt` (so the offline nightly pre-seeds
  the bundle) and in `LLM_HF_MODELS`. Weights come from the published, ungated
  `FastFlowLM/Gemma4-E2B-IT-NPU2` via `Q4NX_MODEL_SOURCE`, the same contract the
  other q4nx examples use.

## Notes for review

- `fused_decode_ple` is a fork of `fused_decode`. For every model *other* than
  gemma4-e2b it must emit byte-identical IR to its parent; that check is in the
  README and passes against current main, including after #1978.
- There is intentionally **no lit test in `fused_decode_ple/`**. It would build
  into the same directory as the model-level compile gate and race it under the
  unfiltered `check-programming-examples` suite, which has no `-j1`. (The same
  race already exists between `fused_decode`'s lit and gemma3-4b's — worth a
  separate look.)
- The KV-shared append stays on the wave's own slab rather than being suppressed:
  a per-wave append geometry that varies would split the channel into more than
  one BD task, which is the first bug above.
- Unrelated, found while wiring this up: `programming_examples/lit.cfg.py` builds
  the test PYTHONPATH as `os.path.join(config.xrt_dir, "python")`, but cmake sets
  `XRT_DIR` to the package directory (`.../share/cmake/XRT`) when XRT is found via
  its config package, so that path does not exist and every pyxrt-importing lit
  fails locally. Not touched here.


erwei-xilinx:
Superseded by #1982, which is the same 47 commits from a branch on this repo rather than the fork, so the LLM nightly perf workflow can be dispatched against it.

Issue #1984, opened 2026-09-10
gemma4_e2b_q4nx decode dispatch hangs nondeterministically on the nightly benchmark runner (ERT_CMD_STATE_TIMEOUT)


erwei-xilinx:
gemma4-e2b's fused-decode dispatch (`programming_examples/fused_decode_ple`) hangs
nondeterministically on the `amdryzenai5pro340` nightly runner. `xrt::run::wait()`
returns `ERT_CMD_STATE_TIMEOUT` (state=8). The same binaries run clean on a
development box.

This is what blocks `run_npu2_sweep.lit` for this model, which is otherwise at
parity with the other q4nx examples (PR #1982).

## Rate

Repeating one staged binary, no rebuild between attempts:

| waves | result |
|---|---|
| `UNI_DEC=4` | 4 fail / 10 |
| `UNI_DEC=5` | 4 fail / 10 |

Consistent with an independent ~10% failure probability per decode wave:
`1-(1-0.1)^n` predicts 34% / 41% at 4 / 5 waves and **97.5% at 35**, matching 7 of
7 observed failures at full depth (including all six contexts of a real sweep).
Retry logic cannot rescue a full-depth run.

Never reproduced on the development box in ~25 runs.

## Established, please do not re-propose without new evidence

- **Nondeterministic, not a depth threshold.** The runner gives
  `1P 2P 3P 4F 5P 6F 8F 35F`, and `UNI_DEC=4` and `5` *flipped* pass/fail between
  two runs of byte-identical binaries.
- **Not the build.** `insts.bin` md5 reproduces exactly local-vs-CI at depths
  3/4/5/6/8/35; the same depth built twice locally is identical.
- **Not mlir-aie provenance.** Rebuilding locally against CI's exact wheel
  (`mlir_aie==1.4.3.dev55+g10767b5`) reproduces neither hang.
- **Not the sliding-vs-full layer geometry mix.** `UNI_DEC=4` failed once and
  passed once.
- **Not the part or column count.** Depth 1 passes on the runner, and the device
  is emitted once and reused temporally, so the spatial footprint does not change
  with depth.

## Only remaining measured difference

| | runner | dev box |
|---|---|---|
| amdxdna / XRT | **2.21.0** | **2.23.0** |
| NPU firmware | 1.1.2.64 | 1.1.2.64 |
| Peano | 22.0.0 | 22.0.0 |
| power mode | Turbo | Turbo |

[...]

PR #1988, opened 2026-09-12
[gemma4_e2b_q4nx] Run Gemma4-E2B on the NPU: prefill and token-level decode

erwei-xilinx:
Gemma4-E2B (text) now runs end to end on the NPU: the prefill as a 22-ELF engine, and the decode token by token through a single 35-layer dispatch. Neither existed when this PR opened — it started as prefill-only, and the description below has been rewritten to match what the branch actually carries.

## Prefill

The whole prompt runs on the NPU — RMSNorm, Q/K/V, per-head QK-norm, the weightless value-norm, RoPE, MQA flash attention, the GELU-tanh GLU, the per-layer-embedding branch, and the LM head — with resident per-layer weight BOs.

`make prefill` scores the full 262144-wide logit vector against the CPU reference the example already ships (`gemma4_e2b_q4nx_weights.forward_prompt`):

```
first-token argmax=9079 'Paris' (expect 'Paris')
logit cosine vs CPU reference = 0.998582 (floor 0.99)
GATE PASS
```

Identical across two independent runs. The argmax alone is deliberately not the gate — on this model a doubled embedding scale and a per-layer-embedding read from the wrong tensor both still predicted `' Paris'` while it was numerically broken.

A **self-owner**, like `gemma3_4b_q4nx`: there is no bf16 Gemma4 for a thin q4nx wrapper to delegate `compile_all_kernels` / `run_transformer_block` to. It builds **22 ELFs** rather than Gemma3-4B's 8, because Gemma4's per-layer class map reaches the ELF shapes — every attention-shaped ELF at head_dim 256 and 512, every FFN-shaped one at 6144 and 12288, plus the three PLE stages.

### head_dim=512

No other model here builds a 512 head dim, and it needs no new kernel. `shared/infra/fa_headfirst` gains a `_FA_TILING[512]` entry of `(32, 256, 8, 1, 64)`, verified on device at **cos 0.999700** against a float64 SDPA reference. Earlier attempts at 512 failed only because they used `num_q_tiles=16`, which asks for 16 herd columns on an 8-column device — a column-count error, not an L1 or architectural limit.

The kernel bakes in a `1/sqrt(head_dim)` score scale and Gemma4's `ATTN_SCALE` is 1.0, so `sqrt(head_dim)` is folded into the `q_norm` weight, where it is applied in f32 before the bf16 store rather than rounding Q a second time.

### Two device constraints

Both surfaced as silent wrong answers, not compile errors, and both are documented at the code:

- **One GEMM per ELF.** Two GEMM slices stitched into a single ELF return partial NaN nondeterministically — same binary, same inputs, clean on 2 of 4 repeats. Q, K and V therefore each get their own ELF instead of sharing the siblings' fused `rms_qkv_qknorm_rope`. This also lets the 20 KV-shared layers skip the K/V dispatch outright rather than running it against zero weights.
- **A narrow GEMM takes fewer herd columns, not a smaller tile.** At N=256 the 4-column/64-wide shape was nondeterministic the same way; 2 columns × 128 is stable over repeats and more accurate (0.999851 → 0.999968). See `gemm_herd_n`.

### GEMM specs

D=1536 puts every Gemma4 shape outside the measured GEMM registry, so the example supplies its own spec function through the hook the builder already exposes for this (`gemm_spec_fn`, as qwen3_4b does). The rule is read off the registry rather than invented, and reproduces the recorded best method on **41 of 41** entries at M=2048, the regime the prefill runs in. One measured exception: the 12288-wide FFN needs `tile_k_l2=64`, because `tile_k_l2 * N` is the K-tile DMA stride and NPU2 caps a BD stride at 1048576.

## Decode

`gemma4_e2b_q4nx_inference.py` is the token-level driver. The prefill seeds the device KV cache and each token is then one dispatch of all 35 decoder layers, the PLE branch and the 4-bit LM head. This closes the example's parity gap with the other q4nx models: `run`, `ask`, `profile`'s decode half, and the shared `verify/` subsystem all needed it.

### The KV hand-off is the part with no sibling

Every device KV row is `REGION_W = 1024 = 2 CUs × DH_A(512)` and the single MQA head is replicated into both halves. But a sliding layer's 256-wide head does **not** sit contiguously in its 512-wide slot: it is scattered as `[real_lo | zeros | real_hi | zeros]` so that the rope kernel's fixed `(i, i+256)` pairing lands on the real `(i, i+128)` pairs — exactly the interleave `gemma4_e2b_q4nx_requant` already documents for the weights.

Seeding it contiguously does not fail. It produced a correct first token — which comes from the *prefill* — followed by `' Parisнии est le Humदा is a het de'`.

### So the gate covers the stop, not just the tokens

`make run` is greedy from `<bos> The capital of France is` and must produce `[9079, 236761]` (`' Paris.'`) and then stop on `<end_of_turn>`. That is not merely recorded from a device run: the CPU oracle re-run on the growing prompt emits exactly `[9079, 236761, 106]`. Passing, two clean repeats.

### Two decode templates, and they are not interchangeable

`compile-decode` builds the layer gate's one-layer template; `compile-decode-full` builds the token driver's 35-layer one. Pointing the driver at the gate's does not fail — it dispatches a 1/35-scale model — so the driver asserts `UNI_DEC`/`ATTN_MAXL` against the builder before dispatching.

## Benchmarks and CI wiring

Both curves the LLM benchmark page carries are wired up:

- `run_npu2_prefill_sweep.lit` — TTFT vs padded prompt length, at 512 / 1024 / 2048. Measured: 1184 ms (432 tok/s), 2177 ms (470 tok/s), 4719 ms (434 tok/s), first-token gate passing at every length.
- `run_npu2_sweep.lit` — decode tok/s vs KV depth, all 35 layers: 18.21 / 16.27 / 13.39 / 9.91 / 6.52 / 3.87 / 2.14 at 1k…64k; 128k out of reach.

`make profile` is now TTFT + decode tok/s end to end: **TTFT 4.505 s, 19.72 tok/s** over 64 tokens. It passes `--ignore-eos`, unlike the siblings, because this model answers the Paris prompt in two tokens and a run that honours the stop would report per-call setup rather than decode.

`verify_adapter.py` wires the example into the shared `verify/` subsystem against `unsloth/gemma-4-E2B-it` (an ungated mirror of `google/gemma-4-E2B-it`). `make diagnosis` passes through it. **`make verify-topk` does not**, and the cause is the prompt rather than the device: `verify_runner` tokenizes with `tok.encode()`, which adds no `<bos>` for this tokenizer, and without it the model goes out of distribution — the NPU's top-5 at the first divergence is `[4092, 4094, 4095, 4096, 4098]`, a run of consecutive ids. What settles it is that on the same prompt the CPU oracle for this bundle, with no NPU involved, emits `[786, 786, 786, 786]`. A degenerate q4nx model cannot match bf16 top-5 whatever runs it. This is written up in the README with the triage step (re-run the ids through `forward_prompt` before suspecting the decode).

## What is deliberately not here

`run_npu2_sweep.lit` marks every context `--expect-fail`, and there is no lit around `make run`, both for #1984: the full-depth 35-wave dispatch hangs nondeterministically on the benchmark runner (amdxdna 2.21.0) while being clean on a development box (2.23.0). A lit around the token gate would not go red there, it would *time out* and block the nightly, which `--expect-fail` cannot express. So `run_npu2_profile.lit` measures TTFT through `profile-prefill` and CI's on-device decode coverage stays the per-layer gate, which is one wave and reliable.

When #1984 is fixed, three things land together: `--expect-fail` comes off the sweep, the profile lit goes back to `make profile --n-tokens 64`, and a `run_npu2_verify_decode.lit` becomes possible.

## Review follow-ups addressed on the branch

- The engine's `compile-decode` tested only for the *existence* of `decode.xclbin`/`decode.insts.bin` after its generator ran, so a failed rebuild with last build's artifacts still on disk reported success and handed the consumer a stale template. The outputs are now removed before the generator runs.
- The decode weight cache was keyed on geometry alone, so switching `Q4NX_MODEL_SOURCE` silently reused the previous bundle's packed weights. The key now carries `Q4nxModel.fingerprint()`.
- `sweep_decode.py`'s `--expect-fail` forgave *any* non-ok status, so a `build_fail` / `no_template` / `stale_template` at an expected-fail context was indistinguishable from the known #1984 timeout. The exemption is now restricted to dispatch-time statuses; build and staging failures stay hard. (Also escapes two `%` in that file's argparse help, which made `--help` raise.)

### DIFF
--- a/programming_examples/llms/gemma4_e2b_q4nx/Makefile
+++ b/programming_examples/llms/gemma4_e2b_q4nx/Makefile
@@ -7,9 +7,9 @@
 # Prefill runs end to end on the NPU and is gated against the CPU reference
 # (`make compile-prefill`, `make prefill-paris`). The decoder layer runs on the
 # NPU too but is gated per LAYER (`make layer-gate`) rather than per token:
-# there is no token-level chatbot driver yet, so the `run` / `ask` / `verify` /
-# `profile` targets the other llms/ examples carry are deliberately absent
-# rather than present and broken.
+# there is no token-level chatbot driver yet, so the `run` / `ask` / `profile`
+# targets the other llms/ examples carry are deliberately absent rather than
+# present and broken.
 #
 # The device path is the fused_decode_ple engine (../../fused_decode_ple), a
 # fork of fused_decode carrying the per-layer-embedding branch this model needs.
@@ -31,10 +31,15 @@ PREFILL_DRIVER := $(srcdir)/gemma4_e2b_q4nx_prefill.py
 BUILD_DIR := build_peano
 
 # Padded prefill length. Every prefill ELF is built FOR this length, so changing
-# it is a rebuild, not a runtime knob -- which is the axis run_npu2_prefill_sweep
-# walks.
+# it is a rebuild, not a runtime knob. Only 2048 has been run; a prefill sweep
+# over other lengths is follow-up work.
 CTX ?= 2048
 export Q4NX_SEQ_LEN := $(CTX)
+
+# Kernel cache, UNDER THE DIRECTORY MAKE WAS INVOKED FROM. The lit tests each
+# run from their own build dir and each start with `make clean`; a cache shared
+# through $(srcdir) would let one test delete the other's ELFs mid-dispatch.
+export Q4NX_CACHE_DIR := $(CURDIR)/_q4nx_cache_seq$(CTX)
 
 # Weight source: HF repo id (model.q4nx bundle) or a local dir/file. Read by the
 # drivers as Q4NX_MODEL_SOURCE, exactly as in the other q4nx examples.
@@ -274,5 +279,5 @@ clean:
 	@$(MAKE) --no-print-directory -C $(DEC) clean
 	rm -rf $(srcdir)/__pycache__ $(srcdir)/build_peano $(srcdir)/build_chess 2>/dev/null || true
 	rm -f $(srcdir)/*.npz $(srcdir)/decode_L*.xclbin $(srcdir)/decode_L*.insts.bin 2>/dev/null || true
-	rm -rf $(srcdir)/_q4nx_cache_seq* $(srcdir)/build_peano 2>/dev/null || true
+	rm -rf $(srcdir)/_q4nx_cache_seq* $(CURDIR)/_q4nx_cache_seq* $(srcdir)/build_peano 2>/dev/null || true
 	@echo "Gemma4 caches and the PLE decode build removed."
--- a/programming_examples/llms/gemma4_e2b_q4nx/Makefile
+++ b/programming_examples/llms/gemma4_e2b_q4nx/Makefile
@@ -3,11 +3,13 @@
 #
 # Gemma4-E2B (text) Q4NX on NPU2 (MLIR-AIR).
 #
-# THIS EXAMPLE IS PARTIAL, and the target list says which half is which. The
-# decoder layer runs on the NPU and is gated there per layer (`make layer-gate`,
-# `make compile-decode`); there is no prefill engine and no token-level chatbot
-# yet, so the `run` / `ask` / `verify` / `profile` targets the other llms/
-# examples carry are deliberately absent rather than present and broken.
+# THIS EXAMPLE IS PARTIAL, and the target list says which half is which.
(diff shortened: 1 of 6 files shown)


erwei-xilinx:
Addressed in 3c0068104.

PR #1989, opened 2026-09-15
[air.api] Convert fused_decode to air.api

erwei-xilinx:
`programming_examples/fused_decode/fused_decode.py` is the decode engine ten
q4nx models share, and it was the last large design still written against the
raw `air.dialects` bindings. This moves it onto `air.api`.

It is a spelling change. The AIR module every model builds is unchanged, and
so is everything compiled from it.

## Gate

A 30-config matrix — ten models × {decode wave, lm-head wave, DYNSEQ} — built
through the builder's own `FUSED_DECODE_EMIT_ONLY` hook and diffed against a
pre-conversion capture. Each model's environment is read from its own
Makefile's `DECODE_ENV`, so the matrix cannot drift from what the real builds
use. DYNSEQ is in it because it is the variant that reshapes the `func`
signature, which a conversion is likeliest to disturb.

Three difference classes are semantic no-ops. Each was cleared by comparing
compiled artifacts before being normalised away:

- air.api ends a buffer's life at its last use, so it emits `memref.dealloc`
  for the segment-scope L1 buffers the raw builder simply leaked.
- A dynamic slice offset lowers through `affine.apply` where the raw bindings
  hand-built `arith.muli`. `--lower-affine --canonicalize`, which the pipeline
  runs anyway, makes the two converge exactly, modulo an `overflow<nsw>` hint.
- Which constant a region materialises first depends on which op asked first.

**Two artifact facts worth recording.** The decode ELF is *not* reproducible —
two builds of the same tree give different md5s — so only `decode.insts.bin`
can carry an A/B there. But `insts.bin` is shim BDs and sync, and says nothing
about what the cores compute; the per-core ELFs under
`air_project/elfs_seg_core_*/` *are* reproducible and are what covers that.
Both were checked.

| | covered | result |
|---|---|---|
| AIR module | 30 configs | identical after normalisation |
| shim instruction stream | llama-3.2-1b, qwen3-8b, lfm2-1.2b | `insts.bin` byte-identical |
| core code | llama-3.2-1b (27 cores), lfm2-1.2b (28) | byte-identical |
| on device | llama32_1b_q4nx | first token 12366, clean EOS, "The capital of France is Paris." |

## What the conversion found

Four things were wrong in `air.api`, and each is fixed rather than worked
around:

- **The L1 budget summed buffers a body had already released**, which rejected
  llama-3.1-8b at 68 KB against a 64 KB tile. It now honours `air.dealloc`.
- **`_charge_shared_l1` applied the emitting herd's rank to every shared buffer
  in the segment**, which cannot be decided when one segment holds herds of
  different rank — fused_decode's are 1-D and 2-D at once.
- **A herd-shared kernel argument was always narrowed to a per-core slab.**
  "Shared" says more than one core sees the buffer; it does not say the cores
  partition it. fused_decode's proj pairs are a flat tile two adjacent cores
  alias, and there is no slab to narrow to.
- **A herd refused the second object file a body reached into.** A core's link
  set *is* a set: `aie-assign-core-link-files` walks the call edges to the
  declarations carrying `link_with` and collects a de-duplicated list. The
  core-level attribute is the deprecated spelling, which that pass merely
  migrates into the same set. The hybrid's conv herd stages a ShortConv layer
  and runs rope on one tile, which is two objects by construction.

The interface checks also move from `build()` to `compile()`. Both read
`self.outputs`, which exists for `CompiledKernel.__call__`: "declares an
output" is "`__call__` has something to return", and "inputs before outputs" is
the positional split XRT is handed. A design that takes the module from
`mlir()` and drives XRT itself asks neither, and fused_decode's ABI — fixed by
host bindings that already implement it — puts inputs after outputs.

## Also here

Two cleanups, each its own commit and each verified the same way:

- A shadowed `_kv_body`: two defs of the same name in a row, the first a paste
  of `_qk_body`'s body under `_kv_body`'s signature. It reads `a_q`, `a_m` and
  `a_cc`, which are not in scope there, so calling it would raise `NameError`.
- Six `# fixed config` constants, each pinned to one value and each keeping the
  branch its other value would have taken. None is read from the environment
  or assigned anywhere else. The largest were real code: `MULTIBLK`'s off-arm
  carried the whole L=1 single-block KV staging path, which cannot run at all
  since the kernels it calls were deleted long ago. That deletion emits
  **byte-identical** IR across all thirty configs — not merely identical after
  normalisation.

## Left on the raw bindings

One transfer, and not for tidiness. The launch-scope weight feed does not
survive conversion, measured twice. Rebuilding its base drops the W_SPLIT
group-relative subtraction and changes qwen3-8b's instruction stream. Keeping
the base exactly and moving only the puts is *also* wrong, on every model: the
two paired puts become independent `affine.apply`s off one base, where the raw
form chains the second `arith.addi` off the first put's offset, and
`coalesceShimDmaOrder` stops merging them — llama-3.2-1b goes from 158032 to
169676 bytes of instruction stream, twice the BDs, which is precisely what
`peelOffset`'s comment in `AIRRtToNpuPass.cpp` says happens silently. Nothing
about the index value is wrong; it is the offset *chaining* the coalescer
pattern-matches on.


erwei-xilinx:
Addressed in ff16029f. Five of the six findings were real; the sixth does not hold.

**Rank-0 tensor is not callable** — right, and the fix belongs where the assumption is. `CompiledKernel.__call__` hands every tensor to `XRTBackend.load`, which allocates a BO and passes the BO rather than an integer, so a bare scalar argument was something that path cannot invoke. Emitting one is still correct — fused_decode's DYNSEQ context length is exactly that, and it takes the module from `mlir()` and binds its own arguments — so `_check_interface` now turns it away alongside the other two things `__call__` assumes, rather than the declaration refusing it.

**Spatial provenance** — right, and the most consequential of the six. `air.sequential` refuses a bound built from a tile coordinate because the trip count would then differ between cores, and a loop with a channel operation in it deadlocks on the ones that run fewer trips. An `IndexExpr` carries that on its leaf; a raw value arrived with nothing, so a converted body's raw `tx` walked straight past the check. `coerce_index` now marks the leaf spatial when the value *is* one of the coordinates currently bound. Arithmetic derived from a coordinate through raw ops stays undetectable, and the comment says so rather than implying otherwise.

**Floating-point SSA argument** — right, the pass-through sat below the `is_float` branch, which turns away anything that is not a Python number. Moved above it.

**`b.released` in the combined L1 check** — right, and it was inconsistent with the change in the same PR: `alloc()` filtered its own holder's buffers but neither it nor `_charge_shared_l1` filtered the enclosing segment's, so a buffer freed after one herd still counted against a private tile allocated in the next. Both filtered now.

**`air.segment(params=...)`** — right, it used `list(params)` where the herd uses the shared parser, so `params=<one rtp>` raised for not being iterable. Reusing `_parse_herd_params`.

**`int()` on an `IntegerAttr`** (suppressed comment on `_index.py:372`) — this one does not hold on these bindings. `int(IntegerAttr(7 : index))` returns `7`; it does not raise and does not fall through to the leaf path. Checked directly, and the behaviour is also what the change was observed to produce: before it, converting the KV re-block made the GLU down-memtile loop materialise its constants in a different order, because its bound had become dynamic. After it, that stopped. Left as is.

Two regression tests added for the behaviours a later change could quietly undo: a rank-0 argument refused by `compile()`, and a raw coordinate still refused as a loop bound.

fused_decode is unaffected by all of this — gate 30/30 across the model matrix, `decode.insts.bin` byte-identical to main.

erwei-xilinx:
Round 2 addressed in 77dc2c4a. All three were real.

**Channel reuse across builds** — confirmed by probe before fixing: a `Channel` built once at import scope and used by two launches emitted `air.channel @shared_chan` in the first module and *not* in the second, leaving every `put`/`get` there naming a symbol its module does not define. The guard was a bool on the Channel object; it is keyed on the module now. This predates the branch and was reachable without `declare()`, but `declare()` is a second public way into the same state, so it belongs here. Test added that builds twice and checks the declaration is in both.

**Segment and launch coordinates** — right, and the reason is exactly the one given: a herd keeps its core position in `_coords` while a segment and a launch expose theirs as `leaves`, so my identity check only ever saw the herd's. All three are consulted now.

Agreed on excluding `launch.wave`, and worth stating why rather than just doing it: a dispatch index is the same on every core, so it is not a per-core quantity at all — which is what makes it a legal bound. fused_decode's attention cores loop on `ceil(L/16)` derived from exactly that. Folding it in would have broken them.

**`params=` diagnostic** — right, the parser is shared and hard-coded `air.herd`, so `air.segment(params=[...])` was reported against the wrong construct. It takes the name of the one being built.

Three tests added for what a later change could undo: a channel declared in both of two builds, a raw segment coordinate refused as a loop bound, and a segment `params=` error that names the segment.

fused_decode unaffected — gate 30/30, `decode.insts.bin` byte-identical.

PR #1992, opened 2026-09-15
[ci] Publish the decode binaries the nightly sweep dispatched

erwei-xilinx:
A number on the [benchmark page](https://xilinx.github.io/mlir-air/llms/) currently cannot be traced back to the bytes that produced it: the sweep builds a decode template pair per context, dispatches it, and discards it. This uploads them.

### Why now

This is what #1984 needs to move. That defect is a dispatch timeout reproducible **only** on the benchmark runner, and the investigation record already notes that a matching `insts.bin` md5 does **not** license "same binary" — the instruction stream is byte-identical local-vs-CI, while the CDO, which lives in the xclbin's `AIE_PARTITION` section, has never been compared at all. It cannot be compared without the artifact.

The confound this is meant to settle is concrete. CI builds against the mlir_aie **wheel** (`1.4.3.dev55+g10767b5`); a development box typically builds mlir-aie **from source**; and those two are on record as producing materially different designs from an identical instruction stream (d=35: 18.59 tok/s sd 0.03 vs 9.81 sd 33.8). Any local-vs-runner A/B that does not control for it is comparing two things at once — including the recent one that appeared to clear amdxdna/XRT 2.21.0 as the cause, which I ran on a from-source toolchain and therefore cannot separate from the wheel.

### Design

- **gemma3_4b_q4nx is collected alongside gemma4_e2b_q4nx as a control.** Closest sibling design, and it sweeps clean on the same runner. A byte difference present in both is the toolchain; one present only in gemma4 is the defect.
- **An md5 manifest covers every model's templates**, not just the two copied out. It is small, and it makes "is this gemma4-specific?" answerable without another CI round trip.
- **Device, driver, XRT and firmware are recorded.** This job has never logged them — only power mode and platform profile — so every runner-vs-local comparison so far has been reconstructed by archaeology, and some of it could not be reconstructed at all.

### Validation

The collection step was dry-run against a simulated CI tree before being committed: per-context prefixes keep `L1024` and `L2048` from colliding across the sweep's per-context workdirs, both models are captured, and the manifest is complete. Artifact size is ~20 MB.

Guarded on `!cancelled()` only, so a narrowed or non-main dispatch still produces it — that is the case an investigation actually runs.


erwei-xilinx:
Not landing this — pushed the branch to origin and dispatched it directly instead (run 35038069393). Diagnostic only.

PR #1993, opened 2026-09-17
[llms] Unbreak the decode sweep: drop the MULTIBLK guard in decode_geometry

erwei-xilinx:
The nightly LLM benchmark has been red since 2026-09-16 ([run 35055996477](https://github.com/Xilinx/mlir-air/actions/runs/35055996477), [run 35182309565](https://github.com/Xilinx/mlir-air/actions/runs/35182309565)). 10 of its 11 `run_npu2_sweep.lit` tests fail identically, before any NPU dispatch:

```
File ".../llms/bench/decode_geometry.py", line 115, in _rope_w_elems
    if not fd.MULTIBLK:
AttributeError: module '_fused_decode_geom' has no attribute 'MULTIBLK'
```

Two commits: the fix, and the gate that should have caught it.

## Cause

`MULTIBLK` was a `fused_decode.py` module constant pinned to `True` ("fixed config: decode is always multi-block"). #1989 inlined the always-taken branch and dropped the name — correct for the builder, but `decode_geometry.py` imports that builder to derive host BO sizes, and `_rope_w_elems` was the one remaining reader of the constant.

## gemma4 is the exception, not the cause

`gemma4_e2b_q4nx` is the only sweep that stayed green, because `PLE_BUILDERS` routes `gemma4-e2b` to the `fused_decode_ple` fork, which still defines `MULTIBLK`. That asymmetry makes the failure read as model-specific when it is the opposite — every model *except* gemma4 is broken.

gemma4's `verify` has reported `pass` every night since 09-13, last night included. The one `"verify_status": "fail"` row for it on the published page is the stale 2026-09-05 PR-branch datapoint already documented in `nightlyPerfBenchmark.yml`'s header, from before the model existed on main.

## Why the guard is deleted rather than defaulted

This function restates how the builder sizes its rms tensor, and `fused_decode.py` now sizes it unconditionally:

```python
UNI_DEC * RMS_LAYER + (UNI_DEC if ROPE_W_PER_LAYER else 1) * ROPE_W_LEN + K
```

A `getattr(fd, "MULTIBLK", True)` would paper over the next such divergence instead of failing on it.

## The gate

Nothing in CI ran `decode_geometry.py --check`, which is why a two-second pure-Python failure cost two ~4.5 h nightlies to surface. The second commit wires it in as `llms/bench/run_geometry_check.lit`, collected by a new `check-programming-examples-llms-geometry` target and invoked from `buildAndTestRyzenAI.yml`.

`REQUIRES:` is empty on purpose — the check imports the decode builders and does arithmetic, so it needs no NPU, no xclbin, no HF weights and no secrets. Gating it on `ryzen_ai_npu2` or `peano` the way the model lits do would make it silently unsupported on exactly the runners where a cheap gate is worth most. It also gives that job's LLM `requirements.txt` install something to serve — every target it was installed for is excluded there.

It is not a smoke test: `--check` re-derives `w_elems`, the qwen3-8b / llama-3.1-8b / qwen2.5-7b weight splits and the rope-per-layer geometry against expected numbers, so a change that keeps the import working but shifts a size fails here too.

## Verification

**The fix.** Geometry derived for all 11 models the sweep runs, at the `--bench-model` / `--vocab-chunk-i2` / `--n-layers` each one's lit passes:

| | before | after |
|---|---|---|
| gemma3-4b, lfm2-1.2b, llama-3.1-8b, llama-3.2-1b, llama-3.2-3b, phi4-mini, qwen2.5-3b, qwen2.5-7b, qwen3-4b, qwen3-8b | `AttributeError` | derives |
| gemma4-e2b | derives | derives (unchanged) |

llama-3.2-1b reproduces `bench_decode.cpp`'s built-in `w_elems=386662400` and `rms_size=67648`.

**The gate**, through real lit and FileCheck against a configured `programming_examples` build, both directions:

- discovery goes 410 → 411 tests; the target's `--filter` selects exactly that one; it PASSES
- re-introducing the deleted `MULTIBLK` guard makes it FAIL with the nightly's exact `AttributeError`; removing it passes again

So the gate is known to fail on the bug it exists for, not merely to pass on green. That run was on a host with no NPU (`xrt-smi` absent), no HF token, no Chess and Peano LLVM 21 — which is the point: it is the LLM signal that survives where every other `llms/` test is unsupported.

A `sweep`-suite dispatch of the nightly workflow on this branch is the on-hardware confirmation; results will be posted here.


erwei-xilinx:
## On-hardware confirmation

Dispatched the nightly workflow's `sweep` suite on this branch: [run 35249006404](https://github.com/Xilinx/mlir-air/actions/runs/35249006404) — **success**, 11/11 sweeps passed, 2732s of test time. Same runner as the nightly (`amdryzenai5pro340`).

Before this PR, 10 of those 11 died in `geometry()` before reaching the device, producing zero datapoints. They now produce **80 measured points, all `status: ok`**, across the full 1k–131k context axis:

| model | ok | tok/s @1k | tok/s @131k |
|---|---|---|---|
| llama32_1b_q4nx | 8/8 | 68.67 | 5.89 |
| lfm2_1_2b_q4nx | 8/8 | 67.59 | 5.88 |
| llama32_3b_q4nx | 8/8 | 28.13 | 2.24 |
| phi4_mini_q4nx | 8/8 | 23.56 | 1.95 |
| qwen25_3b_q4 | 8/8 | 21.75 | 1.68 |
| qwen3_4b_q4nx | 8/8 | 20.88 | 1.67 |
| gemma3_4b_q4nx | 8/8 | 20.61 | 1.99 |
| qwen25_7b_q4nx | 8/8 | 13.36 | 2.04 |
| llama31_8b_q4nx | 8/8 | 13.19 | 1.75 |
| qwen3_8b_q4nx | 8/8 | 12.70 | 1.57 |
| gemma4_e2b_q4nx | 0/8 | — | — |

gemma4's eight points are all `expected_fail`, which is exactly what its lit asks for (`--expect-fail 1024,2048,4096,8192,16384,32768,65536,131072` — every context). So the one model that stayed green through the outage is the one whose sweep measures nothing: it passed because it is expected to fail at every point, and it reached that verdict through the PLE builder that still had `MULTIBLK`. The other ten had to derive geometry to get anywhere, and couldn't.

The artifact landed as `perf-partial-35249006404` rather than the publishable name, and no `perf-history` commit was made — the `github.ref == 'refs/heads/main'` gate held, so this dispatch could not repeat the stale-row incident in `nightlyPerfBenchmark.yml`'s header.

Per-context `xclbin_md5` differs at every point, so these are real per-context builds and dispatches, not a cached result replayed.

PR #1994, opened 2026-09-17
[AIRRunner] Fix an LLP64 truncation that broke 26 runner tests on Windows

erwei-xilinx:
`air-runner` fails 26 of its 56 `Util/Runner` tests on Windows. Linux is unaffected, and no CI caught it.

## Cause

```cpp
double data_rate;                          // the member
void set_data_rate(long bytes_per_cycle);  // the setter
```

Both callers pass a `double`. The narrowing through `long` is lossless on LP64 Linux, where `long` is 64 bits. Windows is LLP64, where it is **32 bits** — and the arch models carry `bytes_per_second` values of `4e9` and `1e11`, both of which overflow int32. MSVC's out-of-range double→integer conversion yields `INT32_MIN`, so affected links came back at **-2147483648 bytes/s**.

A negative rate gives a negative transfer cost. `getTransferCost` returns `(uint64_t)ceil(seconds * cps)`, so the negative cycle count wrapped to ~2⁶⁴, simulated time jumped to `18446744073709536365` (= 2⁶⁴ − 15251), and the run stalled after one event:

```
Error: simulation stalled at 18446744073709536365 cycles without reaching the launch terminator.
```

`dma.mlir` emitted 1 of its expected 32 `DmaMemcpyNdOp` events.

Confirmed by instrumenting the cost function rather than inferred — it reported `bps=-2.147484e+09` (exactly `INT32_MIN`) with `vol=16384 dw=2 bytes=32768`, against a JSON value of `1e11`.

## Fix

`set_data_rate` takes `double`, matching the member and both call sites. It was the only bare `long` left in the runner.

**No behaviour change on Linux.** Every `bytes_per_second` / `bytes_per_cycle` in the tree is `1e9`, `4e9` or `1e11` — integers far inside double's exact-integer range — so a value that used to arrive via a 64-bit `long` now arrives identically via `double`. (The one real difference, that `double` no longer truncates a fractional rate, is unreachable: there are none.)

## Why nothing caught it

The Windows CI job built and ran **nothing** — no test suite at any point. A Windows-only miscompile of the cost model could not turn any job red. Linux CI runs `check-air-mlir` and was green throughout.

So this also adds a `Test` step running `check-air-mlir` there: pure `air-opt` / `air-runner` + FileCheck, no NPU, no XRT, no Vitis, ~30s.

`check-air-python` is deliberately left out — `api/errors.py` passes but takes ~74s against the suite's 30s lit timeout, so adding it would go red for a slowness reason unrelated to correctness. Worth fixing separately.

## Gate

On Windows against `origin/main` at the pinned toolchain (`mlir_aie 1.4.3.dev55+g10767b5`):

| | before | after |
|---|---|---|
| `Util/Runner` | 26 failed, 28 passed | **0 failed, 54 passed** |
| `check-air-mlir` | 511 passed, 26 failed | **537 passed, 0 failed**, exit 0 |


erwei-xilinx:
Addressed the review comment — and it was right, with a bit more behind it than the comment stated.

`data_rate` is bytes per **second**: `getTransferCost` divides bytes by it to get seconds, then scales by the clock to get cycles, so reading it as per-cycle would be wrong by a factor of the clock frequency. The parameter is renamed `bytes_per_second` accordingly.

On where the stale name came from: the `port(llvm::json::Object &)` constructor just above reads a `"bytes_per_cycle"` JSON key. **No arch model in the tree has that field** — every one spells it `bytes_per_second` — and nothing in the tree constructs a port that way, so that constructor is the last user of the older spelling. I noted it in the comment rather than deleting it: removing an unused constructor is a separate change from fixing an overflow, and I'd rather not fold it into a bug fix. Happy to do it as a follow-up if you'd like.

No functional change — `check-air-mlir` stays at 537 passed / 0 failed on Windows.

---

Separately, the Windows job is now genuinely green and genuinely running tests. The first push of the Test step went red, but not on a test: lit treats a missing `psutil` as fatal rather than degrading, so with the suite's `--timeout` it refused to run anything at all. `psutil` was already listed in `utils/requirements_dev.txt` directly under the comment *"lit requires psutil to set timeouts"*, but that file reaches pip as `-c`, and a constraints file installs nothing — the same hole the step's own comment describes for cmake and ninja. Fixed in ca1f2919.

CI now reports, on windows-2022:

```
[0/1] Running the air mlir regression tests
Total Discovered Tests: 555
  Passed           : 537 (96.76%)
  Unsupported      :  11 (1.98%)
  Expectedly Failed:   7 (1.26%)
```

Matching the local numbers exactly. As far as I can tell that is the first time this repo's test suite has run on Windows in CI, so the LLP64 fix is now confirmed by CI rather than only on one machine.

PR #1995, opened 2026-09-18
[llms] One benchmark table: sweep every LLM on both axes, split the VLA out

erwei-xilinx:
The nightly LLM page carried two tables — one scalar row per model, and curves
for the fused-decode (q4nx) models only. That split was read as "bf16 models
can't sweep context length". They can: nothing in those designs fixes the
length. What fixed it was a hardcoded `seq_len = 2048` in each driver.

All eleven autoregressive models now publish a decode curve and a prefill
curve. The scalar table has no rows left and is omitted. `smolvla`, which was
never an LLM, gets its own table.

### Decode tok/s vs context

| Model | 1k | 2k | 4k | 8k | 16k | 32k | 64k | 128k |
|:--|--:|--:|--:|--:|--:|--:|--:|--:|
| llama32_1b | 9.01 | 7.32 | 6.24 | 2.24 | 1.40 | 0.79 | 0.33 | 0.16 |
| llama32_1b_int4 | 13.98 | 11.99 | 8.72 | 2.65 | 1.26 | 0.61 | 0.31 | 0.08 |
| llama32_1b_q4nx | 68.97 | 64.10 | 55.20 | 42.00 | 28.10 | 16.40 | 9.10 | 5.77 |
| llama32_3b | 3.86 | 3.11 | 1.66 | 0.93 | 0.34 | 0.19 | — | — |
| qwen25_0_5b | 10.83 | 9.12 | 7.76 | 3.69 | 2.27 | 1.21 | 0.66 | 0.35 |
| qwen25_1_5b | 6.57 | 5.38 | 3.02 | 1.83 | 1.01 | 0.54 | 0.30 | 0.17 |
| qwen25_3b | 3.31 | 3.04 | 1.82 | 1.09 | 0.60 | 0.32 | 0.18 | 0.04 |
| qwen3_0_6b | 8.69 | 6.39 | 2.65 | 1.39 | 0.76 | 0.41 | — | — |
| qwen3_1_7b | 5.31 | 4.38 | 2.02 | 1.33 | 0.73 | 0.40 | — | — |
| qwen3_4b | 2.60 | 2.22 | 1.10 | 0.62 | 0.31 | 0.17 | — | — |
| smollm2_1_7b | 5.32 | 4.11 | 3.53 | 1.33 | 0.76 | — | — | — |

**The bf16/int4 slope is host attention, not NPU KV streaming**, which is what
the identical curve measures for a q4nx model. Comparable as published latency;
not comparable as a statement about the NPU design. Said so in every lit.

### Prefill TTFT vs padded length (ms)

| Model | 512 | 1k | 2k | 4k |
|:--|--:|--:|--:|--:|
| llama32_1b | 324 | 578 | 1101 | 2288 |
| llama32_1b_int4 | 330 | 586 | 1110 | 2305 |
| llama32_3b | 692 | 1267 | 2735 | 6422 |
| qwen25_0_5b | 372 | 524 | 929 | — |
| qwen25_1_5b | 714 | 1259 | 2514 | 5355 |
| qwen25_3b | — | — | 4645 | 10428 |
| qwen3_0_6b | 420 | 798 | 1668 | 3762 |
| qwen3_1_7b | — | — | 2332 | 5100 |
| qwen3_4b | 1482 | 2918 | 6047 | 13454 |
| smollm2_1_7b | 504 | 892 | 1699 | 3498 |

### What's in it

- `CTX` / `CONTEXTS` make variables, `--seq-len` / `--bench-prefill` /
  `--bench-decode` on ten drivers, replacing the hardcoded 2048.
- `shared/infra/decode_bench.py` + `prefill_bench.py` — one copy of each
  measurement, taking the model's step function as an argument.
- `bench/sweep_decode_runtime.py` — the decode sweeper for runtime-context
  designs. Separate from `sweep_decode.py` because the mechanism differs (that
  one builds a template per context; here one process walks the whole list off
  one build), but it writes the same record, so the table cannot tell them
  apart.
- `kernel_registry`: **39 new GEMM shapes** at M=512/1024, 43 runs, 43/43 PASS.
  Purely additive — 167 → 206 shapes, zero existing rows modified.
- A VLA table with columns named for what they hold (Action chunk, Prefix).

### Cache and measurement hazards closed

`KernelCache` keys entries on kernel name alone, so one directory shared across
lengths serves another length's ELFs. Build dirs are now per-CTX, the verify
adapters per-`max_seq`, and `sweep_prefill.py`'s guard reads the built length
off a `prefill kernels (seq_len=N)` line.

Both benches report **median and min, never a mean**. The host decode upcasts
the whole KV cache per token (1.9 GB/token at 8k on a 0.6B) and occasionally
collides with kernel reclaim: one 12-sample point ran 657–862 ms eleven times
and 46846 ms once. An earlier single-pass revision published 60550 ms for
smollm2 at 4096 and got a 33x "MHA cliff" written into a comment on the
strength of it; the point re-measures at 3498 ms under active contention.
`median >> min` is now the signal for that, and it is carried into the history.

### Known gaps, each reported rather than worked around

- `qwen25_3b` / `qwen3_1_7b` at 512/1024 → `stitcher_mixed_tile_m`. The
  registry's best method flips with length (square Q/O proj: fused-cast/tile_m=64
  at 2048, drain/tile_m=32 at 512/1024 — both pre-existing rows). Both models
  fuse several GEMMs into one ELF assuming a shared tile_m. A stitcher change.
- `qwen25_0_5b` at 4096 → `method_tile_mismatch`. `M=4096 K=896 N=896`
  drain@tile_m=64 is a **real** measurement (rebuilt here at 2943 GFLOP/s,
  faster than drain@32's 2323 and fused-cast@64's 2361) but is not linkable:
  `compile_gemm_mm` bakes DIM_M into mm.o, the stitcher pre-builds only
  `_m32`/DIM_M=32 and `_m64`/DIM_M=64, and `gemm_method_spec` wires drain to
  `_m32`. The schema holds one entry per method name, so drain@32 and drain@64
  cannot coexist — a registry-owner decision.
- `sweep_prefill.py`'s `--expect-fail` used to forgive *any* status. It now
  gates on a named set, so a length listed for a known gap cannot swallow a real
  build break. `wrong_seq_len` is deliberately excluded.

### Verification

`make verify` passes on all ten touched models. `test_perf_publish.py` gains
checks that a VLA model is listed once and not under an LLM column, that an
all-swept page drops the empty scalar table, and that a decode-swept model with
no prefill curve keeps its TTFT — that last one inverted from what it asserted,
because what it asserted was a bug that deleted nine models' TTFT.

All numbers measured on a Ryzen AI 9 HX 370, not the nightly runner.


erwei-xilinx:
Fixed in 3d85d619 — thanks, that was a real one.

`llama32_3b` had `elif args.bench_prefill:` twice in its dispatch chain, so the
second copy was dead. Same root cause as the duplicate `--bench-prefill`
argparse entry already fixed in 851f3e64: one script patched ten drivers and
re-applied an insertion to a driver that already had it. The argparse duplicate
raised on every invocation so the parser-check caught it; this one is silent, so
it didn't.

Scanned the other nine drivers for both shapes of the defect — adjacent
duplicate `if`/`elif` conditions, and any flag appearing more than once in the
dispatch chain after `build_session`. This was the only occurrence.

Re-exercised all three bench paths on `llama32_3b` after the fix:

- `make profile-prefill CTX=512` → 703 ms
- `make profile-decode CONTEXTS=1024` → 258.074 ms median / 239.386 ms min
- `make verify` → PASS

PR #1996, opened 2026-09-18
[llms] Give the last two models a prefill curve, emptying the scalar table

erwei-xilinx:
Follow-up to #1995. The scalar table still had two rows after it, and what they
published was worse than nothing:

| Model | Context | TTFT (ms) | Decode (tok/s) |
|:--|--:|--:|--:|
| qwen25_3b_q4 | **5** | 4057.0 | ↓ |
| qwen3_4b_q4nx | **24** | 5097.0 | ↓ |

Context 5 and 24 are the profile prompt's token counts. So those TTFTs are
near-empty-context measurements sitting in the same section as curves taken at
512–4096, inviting a comparison that isn't valid.

#1995 kept those rows on purpose — before it, appearing in the decode sweep
deleted the whole row and the model's only TTFT with it. But keeping a number
is only an improvement if the number means something.

Both models already had a working `profile-prefill` target and a decode sweep.
**Only the lit was missing.** Adding the two files gives every model both
curves, leaves no row in the scalar table, and the renderer then drops it
entirely — verified by re-rendering run 35307304793's artifacts with these two
curves merged in.

Measured on a Ryzen AI 9 HX 370, first-token gate passing at every point:

| Model | 512 | 1k | 2k | 4k |
|:--|--:|--:|--:|--:|
| qwen3_4b_q4nx | 1711 | 2861 | 5886 | 13077 |
| qwen25_3b_q4 | — | — | 4774 | 10312 |

`qwen25_3b_q4`'s 512/1024 are expected-fail as `stitcher_mixed_tile_m`. It runs
the bf16 `qwen25_3b` prefill (only its decode is q4), so it inherits that
example's limitation: the registry's best method for the square Q/O proj flips
from fused-cast (tile_m=64) at 2048 to drain (tile_m=32) at 512/1024, and the
fused stitcher assumes one tile_m across the GEMMs it merges. Both registry rows
are measured and correct — the stitcher is what cannot express the flip.

Its length knob is `SEQ_LEN` rather than `CTX`, hence `--seq-env`.
`sweep_prefill.py` appends `CTX=` to the make line regardless and this Makefile
ignores it; the built-length guard is what makes that safe to rely on.

### One thing worth picking up separately

The q4nx prefill drivers time themselves instead of going through
`shared/infra/prefill_bench.py`, so their points carry no median/min — 10 of 19
models have it, all of them the ones #1995 touched.

On run 35307304793, `lfm2_1_2b_q4nx` published **81871 ms at 4096 against 1112 ms
at 2048** — 74x for a 2x length increase. With no min beside it there is no way
to tell a real cliff from one contended sample. That is exactly the mistake a
single-pass prefill bench already caused once, for `smollm2_1_7b` (60550 ms,
which re-measured at 3498 ms under active contention). Noted in the
`qwen3_4b_q4nx` lit.


erwei-xilinx:
Correction to the last section of the description, pushed as ea4bc108.

I cited `lfm2_1_2b_q4nx`'s 81871 ms at 4096 as something that *couldn't be told
apart* from a contended sample without a min. That was the wrong framing — it
can be told apart by re-measuring, which I should have done before writing it.

It reproduces:

| | 2048 | 4096 | ratio |
|:--|--:|--:|--:|
| Krackan runner (run 35307304793) | 1112 ms | **81871 ms** | 74x |
| idle second box (HX 370, load 0.10) | 1315 ms | **81794 ms** | 62x |

The 2048 point differs **18%** between the two machines — ordinary spread. The
4096 point differs **0.1%**. Two boxes don't land on the same number under
contention; that's a deterministic fixed cost.

So it's a **real cliff**, not a measurement artifact, and it wants its own
investigation rather than a robustness fix. Unifying the q4nx drivers onto
`shared/infra/prefill_bench.py` is still worth doing on its own merits, but it
would not have caught this — the lit note now says so.

Nothing else in this PR changes; the two new lits and their numbers stand.

erwei-xilinx:
Taking the fix, correcting two of the three claims — 2982f573.

**Right, and the reason I'm taking it:** `profile-prefill` delegates to `bench`,
which passes `--bench-l $(BENCH_L)` with `BENCH_L ?= 2048`. The lit only
redirected the engine knob (`--seq-env SEQ_LEN`), so `BENCH_L` stayed pinned at
2048 for every point. Added `--bench-env BENCH_L`.

**Wrong — "the 4096 point is published under the wrong x-axis value."** It
wasn't. The prefill pads to the engine length, so feeding 2048 ids to a 4096
engine costs the 4096 engine's time. Measured both ways:

| | driver's own line | wall | published TTFT |
|:--|:--|--:|--:|
| without `--bench-env` | `L=2048 … 199 tok/s` ❌ | 10312 ms | 10312 ✅ |
| with `--bench-env BENCH_L` | `L=4096 … 399 tok/s` ✅ | 10274 ms | 10274 ✅ |

0.4% apart. `sweep_prefill.py` derives tok/s from the requested length, not from
the driver's line, so the record was correct. The driver's *printed* tok/s was
off by the ratio — that part is real and the fix corrects it.

**Wrong — "the 512/1024 builds hit `assert N <= self.seq` before producing
`stitcher_mixed_tile_m`."** They fail at *build*, before any bench runs, which
is why they already reported the intended status.

**But this is the most useful part of the report**, because it's a latent trap
rather than a current one: with `BENCH_L` pinned at 2048, the day the stitcher
limitation is fixed, 512 and 1024 would start failing on that assert instead — a
fresh failure for an unrelated reason, arriving exactly when the real one was
fixed. That alone justifies the change.

Re-swept all four lengths with the fix: 512/1024 `expected_fail`
`stitcher_mixed_tile_m` as before, **2048 → 4807 ms**, **4096 → 9695 ms**,
first-token gate passing at both. Lit numbers updated.

`qwen3_4b_q4nx` doesn't need this — its recipe is `Q4NX_BENCH_L=$(CTX)` inline,
and `CTX` is what `sweep_prefill.py` puts on the make line.

PR #1998, opened 2026-09-18
[gemma4_e2b_q4nx] Give the decode stack its own L1 bank

erwei-xilinx:
Draft. Partial fix for #1984 — the dominant half. Not ready to merge: the correctness gate hasn't been run (see Gates below), and a second defect remains.

## What #1984 actually is

Two independent bugs, previously conflated. This PR fixes the first.

## Bug 1 — L1 bank occupancy (fixed here)

L1 is four 16 KiB banks. `DECODE_STACK` decides where the stack ends, and the allocator fills the rest of that bank with buffers. Measured on an NPU2 Krackan, gemma4 decode at 34 waves / `ATTN_MAXL=1024`, failures out of 24 dispatches vs bytes of buffer sharing bank 0:

| `DECODE_STACK` | shares bank 0 | fail |
|---|---|---|
| 8192 | 6144 B | 13/24 |
| 10240 (default) | 6144 B | 10/24 |
| 12288 | 3072 B | 4/24 |
| 14336 | 2048 B | 5/24 |
| **16384** | **0 B** | **0/24** |

A threshold on whether anything shares the bank, not a gradient in headroom. `16384` is chosen for being *exactly one bank*, not for being large.

### It is not stack overflow

Worth stating, because a timeout-under-load reads like a blown stack:

- deepest call chain is ~4 KiB against the 10240 default
- removing ~4 KiB of real frame (`attn_kv_fin` stages a whole head slab it only needs one tile of) moves the rate 42% → 33%, i.e. nothing
- `8192` is a *smaller* stack with the *same* sharing and the same failure rate — a headroom story can't produce that
- a stack that genuinely overflows looks nothing like this: `1024` fails **24/24, deterministically**

### Not the driver, OS, or silicon

It reproduces on Windows, where the kernel driver is the KIPU/WDDM stack rather than Linux `amdxdna`, on a byte-identical CI-built xclbin (md5 verified). #1984's "only remaining measured difference: amdxdna 2.21.0 vs 2.23.0" doesn't survive that. gemma3-4b at the same depth, context and KV volume is 24/24 clean on the same machine.

## Bug 2 — still open, not addressed here

With the stack fixed, a residual remains that is caused by PLE being *instantiated at all*. All at `stack=16384`, depth 35, `ATTN_MAXL=1024`:

| variant | fail |
|---|---|
| no-PLE | **0/32** |
| PLE, full | 4/32 |
| PLE, weight feed off | 4/32 |
| PLE, feeds off + `@pliOut` off | 3/32 |
| PLE, `PLE_BYPASS` | 5/32 |

Nothing *inside* PLE's dataflow matters. Ruled out: depth 35 (no-PLE at d35 is 0/32), bank sharing (zero at 16384), and shim await pressure — the stripped PLE build emits **51 `dma_await_task`, identical to the clean no-PLE build**, and still fails. The only structural difference left is 4 tiles in column 3, including shim `(3,0)`, which the no-PLE build never instantiates.

Two conclusions recorded in `fused_decode_ple.py` don't survive a clean baseline: "no-PLE control 10/10" and "rung 5 (no feed at all) is 40/40" were both measured at the default stack, where bank contention dominated. Those sweeps were reading bug 1.

## Why the model spec, not the Makefile

Three q4nx Makefiles already carry `DECODE_STACK`, so the Makefile is the current convention. The spec dict is better here:

- which stack is right follows from the model's dimensions — exactly what those three Makefile comments say ("at K=4096 … the 10240 default does not fit"). `UNI_LM` resolves this way one line below.
- it removes a duplication: the values also live in `decode_geometry.py`'s check tables, which must re-state them via `env_extra` to derive a model's geometry. Two copies that can drift.

Env still wins, so sweeps and bisects are unaffected.

## Blast radius

Only gemma4 moves. Verified by resolving `STACK_SIZE` per model with no env override:

```
gemma4-e2b   (ple fork) = 16384
gemma3-4b    (ple fork) = 10240
llama-3.2-1b (parent)   = 10240
```

## Effect

Production path (PLE, 35 waves, `ATTN_MAXL=1024`): **22/24 failures → 5/32**. Good dispatches run at 50 ms, the correct speed.

## Gates

- [x] `black` clean; only gemma4's resolved stack changes
- [x] builds and dispatches on NPU2 Krackan
- [ ] **`make verify` for gemma4** — not run. Stack size changes allocation, not arithmetic, so numerics should be unaffected, but that's reasoning and verify is the gate. Needs the Linux runner (HF weights + full pipeline).
- [ ] sweep still `--expect-fail`s every context until bug 2 lands


erwei-xilinx:
Closing: obsolete. The failure this PR raises `DECODE_STACK` to avoid is the chain-lock deadlock, fixed generically in #2008, and the stack size was perturbing it rather than fixing a bank-conflict defect.

Measured on a tree that is clean #2000 + #2008 with this PR **not** applied, so `DECODE_STACK` sits at the old 10240 throughout. All at `default` power mode, `DECODE_NO_PLE=1` — the same no-PLE configuration the table in this PR's description was measured in:

| arm | stack | #2008 | result |
|---|---|---|---|
| control | 10240 | **off** | **39/64** — 25 failures, states 6 and 8, scattered |
| A | 10240 | on | **64/64** |
| A2 | 10240 | on | **128/128** |

The control is the point: it reproduces this PR's failure on today's machine, at ~39% against the 10/24 in the table above, and the only thing that changes between the arms is whether the labeller declines ping-pong into the serialized fan-in. Both arms came from one compiler binary, separated by a temporary env hatch that is not in any branch.

The PLE build is likewise clean at 10240 on the same tree: 300/300 on a build from the tip, and 2760/2760 earlier.

That also explains what the comment in this PR admits it could not. A bank conflict should cost arbitration cycles, not correctness, and qwen2.5-7b and qwen3-8b share a bank and are clean — they do not build this chain. Stack size changes L1 layout, which changes timing, which changes whether the arbiter race is lost. Hence the threshold-not-gradient shape, and hence "this pairs with something in gemma4's own geometry".

No follow-up needed here; the 16384 value is not required once #2008 lands.

PR #2002, opened 2026-09-19
[AIR] Replace the memtile fan-in/fan-out daisy chain with a per-slot rendezvous

erwei-xilinx:
The second defect behind #1984 — the one that survives the `DECODE_STACK` fix in #1998. gemma4-e2b's PLE decode hangs in firmware TDR on NPU2 Krackan: 7 of 20 dispatches at `default` power mode, `ERT_CMD_STATE_TIMEOUT` and `ERT_CMD_STATE_ABORT` at the 7.0/9.0 s watchdog.

## Cause

A cyclic wait between AIR's serialized fan-in chain lock and a shared switchbox arbiter.

`buf143`/`buf162` on `mem_tile_2_1` gather the four projection columns: a chain-lock fan-in set, 4 writers, 2 ping-pong slots. `pickChainBdLocks` daisy-chained them, `cap → W0 → W1 → W2 → W3 → R → cap`, imposing a compile-time total order on arrivals that are independent at runtime — four free-running herds in four columns.

All four gather streams egress to the shim row and travel west. Switchbox (2,0) has 9 masters and 6 arbiters, so three pairs must share, and the pathfinder put W0 and W2 on arbiter 1 (W1 and W3 on arbiter 2). A master port is backpressured until its S2MM channel's lock frees, and an arbiter granted to a stalled packet cannot grant another slave. So when ping-pong lets the east columns run a round ahead, W2's packet wins arbiter 1, stalls on W1's chain token, and holds the arbiter W0 needs — W0 being the writer the whole chain is waiting on.

It predicts the per-herd bisect cell by cell:

| blk0 PP | blk1 PP | pass |
|---|---|---|
| no | yes | 0/20 |
| yes | yes | 4/20 |
| yes | no | 20/20 |
| no | no | 19/20 |

Only run-ahead of the *late* stages matters, because an earlier stage holding the arbiter cannot close a cycle — the later stage was waiting on it anyway. The PLE is the amplifier, not the cause: it lengthens the reader's credit-return path so the stall window is wide enough to overtake.

## Fix

Each buffer **slot** gets its own (capacity, signal) pair instead of chaining the stages. For slot `s`, the multi side acquires `cap[s]` by 1 and releases `sig[s]` by 1; the single side acquires `sig[s]` by N and releases `cap[s]` by N, with `cap[s]` primed to N. Each slot is a counted barrier and the N participants are mutually unordered, so an early arrival costs latency and can never hold a resource an earlier stage needs. Run-ahead stays bounded to `pp_slots`.

Indexing by slot is what makes a counted lock work here, and is why the daisy chain looked necessary: one shared counting semaphore cannot bound per-participant run-ahead, and an AIE2 BD has exactly one lock-acquire and one lock-release field, so an N-way join is not expressible either. Per-slot counted locks need neither.

## Scope

- **Refeed buffers keep the daisy chain** (`ChainLockSet::serialized`). Their credits come from `air.refeed_count`, not the participant count, and they never get a ping-pong twin. `memtile_chain_lock_v2_refeed.mlir` passes untouched.
- **`air.no_chain_lock` stays honored for fan-out only**, so tagged designs keep their lowering. It is now a no-op in practice, but changing that isn't this PR's business.
- `use_lock_race_condition_fix_v2` defaults off; only `fused_decode` and `fused_decode_ple` enable it.

## Validation

On device at `default` power mode, same tree and geometry apart from the lock template:

| build | result |
|---|---|
| daisy chain (control) | **7/20** — 11 TIMEOUT, 2 ABORT |
| per-slot rendezvous | **40/40**, mean 65.1 ms |

Both emit identical logits. No performance change — the control's non-hanging dispatches run 64.5–64.6 ms against the fix's 65.1 ms mean.

`check-air-mlir` 557 tests, 539 passed, 0 failed. `check-air-cpp` 1/1. clang-format 17 clean. `fused_decode`, the other v2 user, compiles clean.

Four tests pinned the old chain and now assert the new invariant — every participant takes the *same* lock for a given slot. `memtile_chain_lock_v2_fanin_n3.mlir` is new: it pins the credit count to the participant count rather than the constant 4, and covers an odd N.

## Follow-up

`run_npu2_sweep.lit` has all eight contexts in `--expect-fail` and `run_npu2_profile.lit` skips decode, both marked temporary. Flipping them needs #1998, #2000 and this all landed, so it's a separate PR.


erwei-xilinx:
Closing: the fix in this PR is unsound.

A counted semaphore tracks arrivals, not distinct participants. With `cap[s]` primed to N, a single writer can consume all N credits across successive rounds — W0 alone taking t0, t2, t4, t6 drives `sig[0]` to 4 — and the reader then drains a slot whose other subregions are stale. Verified against the emitted gemma4 IR, so the "Validation" table above does not mean what it says: the pass rate improved for a reason other than the barrier being correct.

It is not patchable. Per-participant gating needs N private locks the reader must release, and an AIE2 BD has exactly one lock-acquire and one lock-release field, so the N-way join is not expressible. That constraint is what made the daisy chain look necessary in the first place.

The root cause in the "Cause" section above still holds and is unaffected by this — the defect is in the proposed remedy only. The hang is fixed instead by declining ping-pong for herds that feed a serialized fan-in chain, which removes the run-ahead rather than changing the lock template. Measured 2760/2760 against 24/160 on the unguarded build. That change will come as its own PR.

PR #2006, opened 2026-09-20
[flash_attention] Cut latency-bound elementwise passes in attn_npu2.cc

vule20:
## Summary

Three per-tile elementwise passes in [attn_npu2.cc](programming_examples/flash_attention/kernel_fusion_based/attn_npu2.cc) (`mul_r_gp`, `max_g_bf16`, `exp_g_minus_u`) were
latency-bound: each ran as one serialized load -> op -> store chain per 32-lane vector
(17-27 cycles per vector, mostly nops), with the per-row broadcast vector built from four scalar
loads, broadcasts and inserts per group. This rewrites them so independent vectors overlap.
Output is **bit-identical** to the current kernel.

## Change

- `mul_r_gp`: the row vector comes from three `interleave_zip` stages on one vector load; the 8 column
  blocks of a group are loaded, multiplied and stored as separate fully unrolled batches.
- `max_g_bf16`: unrolled max tree over the column blocks, one 8x8 `aie::transpose`, three vector max
  stages, one v8 store (replaces a dependent max chain, 4 scalar `reduce_max` and 4 scalar stores).
- `exp_g_minus_u`: the 16 row vectors are built once per call into a 1 KB static table; each group's 8
  vectors run as independent unrolled chains. The group loop stays rolled: unrolling it too gives
  17168 B of text, over the 16384 B program memory.

Each new path is behind `#if lqp == 64 && dv == 64` (`mul_r_gp`) or `#if lqp == 64 && lkp == 64`
(the other two), with the original code kept as the `#else` branch, so other tile shapes are
unchanged. Every FlashAttention config in the tree compiles with per-tile `lqp=lkp=dk=dv=64`, so the
guards are satisfied by all of them: **every FA user runs the new code.**

Program memory: text 14304 -> 14352 B (+48 B), plus 1 KB `.bss` for the `exp_g` table.

## Performance (NPU2, Ryzen AI MAX+ 395)

FlashAttention as used by the SmolVLA vision encoder (seq 1024, 12 heads, head_dim 64, non-causal,
bfp16), one call, avg of 50 runs:

| | before | after |
|---|---:|---:|
| ELF | 2540 us | 2130 us (-16%) |
| xclbin | 1890 us | 1384 us (-27%) |

In SmolVLA (`make profile`, 3 alternating rounds x 15 reps) the `flash_attn` device time drops from
33.4 to 27.5 ms per image (-17.7%).

## Correctness

Run on NPU2 hardware, patched tree vs the unmodified tree:

- All nine NPU2 lit tests' `make run` commands print `PASS!`: seqfirst, elf, gqa, gptoss_20b, llama3_8b,
  qwen25_7b (head_dim 128), temporal_causal, causal (LK=LQ=16384), infinite_bd_loop. `mean_rel_L1`
  is identical to four digits (4.573e-02 seqfirst, 3.440e-02 temporal_causal).
- The `#else` (original) branches were exercised by forcing the guards off in a copy of the source and
  re-running six of the tests: all pass, text back to 14304 B.
- Output compared **bit for bit** against the unmodified kernel, same inputs: non-causal (12 heads),
  GQA (12 q / 6 kv), head_dim 128 (8 q / 2 kv), and causal (seq 1024 and 2048, head_dim 64 and 128):
  0 differing elements in every case.

## Not covered

- I ran the lit tests' `make` commands directly, not through `lit`.
- The LLM models that link this kernel (llama, qwen, ...) were not run end to end (no weights on the
  test machine); their FA configs are covered by the lit tests above.
- NPU1 is unaffected (`attn_npu1.cc` is a separate file) and was not run.


vule20:
For context, this takes the SmolVLA vision encoder's flash_attn from 33.4 to 27.5 ms per image on NPU2, with bit-identical output. could a maintainer approve the run?

erwei-xilinx:
Hi @vule20, thanks for working on this! I reproduced this on NPU2 — 1812 µs on `main` vs 1291 µs here, 12 heads / seq 1024 via `profile-xclbin`. The win holds up, and I think it should land.

One thing I'd push back on is the `#if`s. By your own argument in the description, every FA config in the tree satisfies the guards, which means the three `#else` bodies are unreachable in every build and every test we have. Forcing the guards off by hand and re-running is the right check, but it isn't repeatable by anyone else, so those branches will drift out of correctness without anyone noticing.

The alternative is to make the fast paths generic so there's nothing to fall back to. That turns out to be mostly a matter of writing the bounds as expressions — `gi < 16` is `lqp/4`, `v[8]` is `col_blocks`, and so on. `max_g_bf16` genuinely needs 8 column blocks for the transpose, but that can be had with a rolled v64 max instead of the hand-unrolled tree.

Worth knowing before you go that way: it exposes a stack problem. Once the fast `exp_g_minus_u` is reachable above `lkp=64`, the full unroll keeps `2*col_blocks` vectors live and spills — 2624 B at `lkp=96`, 3712 B at 128, against a 2048 B default stack with `fused_softmax` nested above. `unroll_count(8)` fixes it at no cost, since at `lkp=64` the trip count is 8 and the codegen comes out byte-identical to the full unroll.

I have this as four local commits on my dev machine on top of `ba684a2c` if you want them: <link>. No tile-shape `#if` left, +79/−119 against `main` instead of +150, all nine NPU2 configs passing with `mean_rel_L1` unchanged. Happy to hand it over, fold it in, or drop it if you'd rather keep this PR as-is.

vule20:
Thanks, agreed. Every FA config in the tree compiles with 64-wide tiles, so the #else branches are unreachable and would only drift. I'lldrop the guards and make the fast paths generic.

If you can share your local commits (a branch or patch is fine), I'll build on them rather than redo the work. Either way I'll re-verify on NPU2 hardware before pushing. I'll push the update to this branch and report the results here.

vule20:
Thanks for the catch on the stack usage. I just pushed to `2f4cc865`. 

What were changed?:

1.  the three fast paths are unconditional now, no `#if`/`#else` left. `max_g_bf16` uses a rolled `v64` max loop over `col_blocks` instead of the unrolled 8-way tree
2. `exp_g_minus_u` chunks the column loop by `exp_col_chunk(col_blocks)` (largest divisor of `col_blocks` up to 8) instead of a hard-coded 8, giving byte-identical codegen to the old guarded path at `lkp=64` and bounds the live-vector count (so the stack problem you flagged doesn't reappear if this ever runs above `lkp=64`)
3. `mul_r_gp` is the  same structure written against `lqp`/`dv`. Text 14352 -> 14336 B, `.bss` unchanged.

Re-verifieed on NPU2: all 9 lit tests passed. xclbin 1382.7 us, ELF 2093.2 us here vs 1384/2130 before

erwei-xilinx (approved):

PR #2007, opened 2026-09-20
[layer_norm] Add a C++ row kernel for the affine LayerNorm (opt-in)

vule20:
## Summary

The `air.api` LayerNorm is latency-bound: it walks a 768-element row in 16-lane steps, chaining two
f32 reductions through an L1 scratch buffer before the epilogue, one row at a time on 8 tiles
(~2700 cycles per row). This adds an opt-in C++ row kernel with the same arithmetic order that uses
32 lanes and normalizes 4 rows together so their independent accumulate / reduce chains overlap.
The default path is unchanged.

Batching rows per DMA does not help the original (581 / 582 / 583 us for 1 / 4 / 8 rows per DMA), so it
is compute latency, not transfers. A wider herd (8x2, 8x4, 4x4) does not compile: it needs more shim
DMA channels than exist.

## Change

- `layer_norm_rows.cc`: `ln_rows_bf16(x, param, y)` for `LN_ROWS` rows of `LN_N` bf16 elements.
  mean = bf16(sum_f32(x)/N); d = x - mean; var = sum_f32(bf16(d*d))/N; rstd = bf16(rsqrt(var+eps));
  y = bf16(bf16(d*rstd)*weight + bias), the last multiply-add as one accumulate (one rounding, as the
  DSL version's vector codegen does). Rows are processed 4 at a time (`LN_GROUP`), fully unrolled.
  4 is the best point: larger groups spill accumulators to the stack and get slower, and 8 does not run.
- `layer_norm.py`: `build_launch` / `build_module` gain `ext=False, rows=4`. With `ext=True` the herd
  links `layer_norm_rows.o` and normalizes `rows` rows per DMA with the kernel; the default
  (`ext=False`) path is untouched. New `--ext` / `--rows` options on the example's CLI. Unsupported
  shapes raise a `ValueError` (N not a multiple of 32, rows per tile not a multiple of `rows`).
- `shared/infra/external_kernels.py`: `compile_layer_norm_rows(n, rows)` builds the object, next to
  `compile_attn_npu2` and friends. (`layer_norm.py --ext` imports it through a `sys.path` tweak to
  `../llms`; happy to move the helper if you would rather the example not reach into `llms/shared`.)
- `Makefile` target `run_ext` and `run_makefile_peano_ext.lit`, next to the existing ones.

## Performance (NPU2, Ryzen AI MAX+ 395)

Standalone, M=3072, N=768 over 8 tiles (avg of 50 runs):

| | DSL (current) | C++ kernel |
|---|---:|---:|
| ELF | 731 us | 443 us (-39%) |
| xclbin | 583 us | 303 us (-48%) |

Motivation: the SigLIP encoder in `llms/smolvla` runs two of these per layer. With the kernel wired in
(a separate change) the NPU device time per image dropped 92.7 -> 90.3 ms and end to end by about 10 ms.

## Numerics: not bit-identical, as accurate

The fp32 sums are reduced in a different order than the DSL's, so **the output is not bit-identical**:
at 3072x768 about 14% of elements (in about 20% of rows) differ by one bf16 ulp. Accuracy is the same
or slightly better against an f32 reference:

- the example's own gate: `mean_rel_L1` 4.317e-03 (C++) vs 4.437e-03 (DSL), rtol 1.6e-2 / atol 6e-2, PASS;
- cosine vs the f32 reference 0.999950 for both at 3072x768; largest error one bf16 ulp.

## Tests (NPU2)

- The example's gate, `make run` and `make run_ext`, both `OUTPUT_FORMAT=xclbin` and `elf`: all four PASS.
- A matrix comparing the DSL and the C++ kernel on identical inputs against an f32 reference: shapes
  1024x768, 3072x768, 1024x1536, 2048x2048 on xclbin and 1024x768, 3072x768 on ELF, each with six data
  regimes (normal, large mean offset, tiny values, large values, constant rows, rows of mixed scale).
  36 of 36 pass (no NaN/inf, including zero-variance rows; C++ mean error <= 1.5x the DSL's and in
  practice equal or lower). The large-offset regime is inherently poor for both (cosine 0.95, from the
  bf16 rounding of the mean in the original design); the two match there.
- Default path unchanged: at 3072x768 the DSL output is bit-identical between this tree and unmodified
  `main` (0 of 2,359,296 elements differ).
- Formatting: `git clang-format` (17.0.1) and `black` clean.


vule20:
this kernel is used by the SigLIP encoder in llms/smolvla (two LayerNorms per layer). I have a small follow-up ready that switches SmolVLA over to it, with NPU device time per  image 94.5 -> 85.3 ms (about 10%) together with a loop-tiling change, bringing  end-to-end latency reuduction for SmolVLA by 7.5% . I'll open it after this lands. Btw, the workflows are waiting for approval on this PR, could a maintainer approve the run? Thanks

erwei-xilinx (approved):

PR #2013, opened 2026-09-22
[smolvla] Use the C++ LayerNorm kernel and unrolled loop tiling for the fused ELFs

vule20:
## Problem

Profiling `llms/smolvla`'s SigLIP vision encoder on NPU2 turned up two separate inefficiencies in how the
fused vision ELFs run:

1. **LayerNorm is latency-bound.** The stock `air.api` LayerNorm loop spends ~2700 cycles per row, almost
   all of it latency: it walks a 768-element row in 16-lane steps, chaining two f32 reductions through an
   L1 scratch buffer before an epilogue, one row at a time across 8 tiles. Batching rows per DMA doesn't
   help (581/582/583 us for 1/4/8 rows) — the cost is compute latency, not data transfer, so a C++ kernel
   that overlaps independent rows' reduction chains is the fix (added separately as #2007, now merged).
2. **The fused ELFs' loop tiling was too coarse.** `runtime_loop_tiling_sizes` controls how much of a
   GEMM's launch-iteration loop is unrolled into the ELF's control stream, and dispatch time tracks that
   stream's size (the firmware parses it on every launch). The default `2,2` leaves `vit_o_ffn`'s control
   stream at 830 KB; a wider tiling shrinks it and speeds up every dispatch.

## Change

Two changes to how the vision ELFs are built, both environment-selectable through a new `smolvla_fuse.py`:

1. **LayerNorm** (`SMOLVLA_LN_EXT`, default `1`): LN1 (`vit_ln_qkv`), LN2 (`vit_o_ffn`) and the
   post-LayerNorm use the C++ row kernel from #2007 (`0` keeps the old `air.api` loop).
2. **Loop tiling** (`SMOLVLA_OFFN_TILING`, `SMOLVLA_LNQKV_TILING`, default `12,6`): `runtime_loop_tiling_sizes`
   of the two fused ELFs, previously `2,2` — `12,6` matches the launch grids' own extents, so the loop is
   fully unrolled.

## Why `12,6`

`vit_o_ffn`, batched over the three camera images (standalone ELF, avg of 60 runs):

| tiling | control stream | `vit_o_ffn` |
|---|---:|---:|
| 1,1 | 1252 KB | 12.0 ms |
| 2,2 (before) | 830 KB | 9.40 ms |
| 4,4 | 748 KB | 8.6-9.0 ms |
| 6,6 and up | 715 KB (fully unrolled) | 8.4-8.7 ms |
| **12,6** (the launch grids' own extents) | 715 KB | **8.43 ms** |

`vit_ln_qkv` alone: 3188 us at `2,2`, 2689 us at `12,6`. `12,6` is at or near the best of every value tried
(`6,3`, `12,3`, `8,8`, `12,12` were within noise of it). `omit_pingpong=all` was tried too and is far worse
(13 ms), so that stays off.

## Cache

The ELFs bake both choices in, and the kernel cache isn't keyed on sources, so they're part of the cache
directory name (`build/vision_kernel_cache_ln_o12x6_q12x6`) — a cache built with other settings is never
reused. **The first run after this change compiles the ELFs (about two minutes).** A setting left at its
old default adds nothing to the name.

## Correctness

- `make verify` passes with the defaults: cosine 0.99933, nMSE 0.00698, max_abs 0.0329 (before: 0.99853,
  0.00578, 0.0317; gate is cosine >= 0.99, nMSE <= 0.04). The C++ LayerNorm kernel isn't bit-identical to
  the `air.api` one (different f32 reduction order), so individual metrics move a little in both directions
  — always within gate, and closer to it in aggregate here.
- The tiling change on its own is bit-identical: at `12,6` the action chunk matches `2,2` to 16 digits
  (cosine 0.9993277192115784 in both).
- Both `SMOLVLA_LN_EXT=0` and `2,2` (the old defaults) still build and pass — checked while developing, so
  the change is a pure opt-in that doesn't remove the old path.

## Performance

NPU2 (Ryzen AI MAX+ 395), NPU `pmode=turbo`, CPU governor/EPP `performance`, `make profile`, 3 alternating
rounds x 15 reps, this branch vs unmodified `main`:

| | main | this PR |
|---|---:|---:|
| end to end, NPU vision + CPU rest | 587.8 ms | 558.0 ms (-5.1%) |
| vision stage, 3 images | 315.6 ms | 287.6 ms |
| NPU device time per image | 94.5 ms | 85.3 ms (-9.7%) |
| `vit_ln_qkv` per image | 18.5 ms | 14.5 ms (-22%) |
| `vit_o_ffn` per image | 42.0 ms | 37.1 ms (-12%) |
| `flash_attn` per image | 32.1 ms | 26.41 ms ( thanks to #2006 |

Every round of this PR (553 / 566 / 555 ms) is below every round of `main` (590 / 590 / 584 ms). Pure CPU
was 533 ms in that session, so the NPU path is still slightly slower than CPU on this machine (0.956x vs
`main`'s 0.908x) — closing that gap further is what #2006 (FlashAttention) and future work target.

## Not covered

- I ran `make compile` / `verify` / `profile` directly, not through `lit`.
- Only NPU2 (Strix Halo) was measured; the tiling values were chosen by measurement on this machine.
- `12,6` is tied to the current shapes (batch of 3 images x 1024 tokens): the launch grids are 12 x
  `{2,6}`. Other batch sizes would want their own values, hence the environment overrides.
- A related silent-miscompile finding for non-all-ones tilings on a launch with three grid dimensions
  (this PR doesn't use one) is written up in a separate issue, not filed yet.


erwei-xilinx (commented):
The measurements are convincing, and reusing the `ext=` path from #2007 is the right call. Five things I'd like addressed before this merges.

**1. The description covers two changes, but the branch contains three.** Commits `ab8ac01` and `755dfcf` add `_build_launch_q_in_segment` and the `SMOLVLA_FA_QSEG` flag, which together account for 380 of the 477 added lines and are not mentioned anywhere in the PR body. Please either split them into their own PR or describe them here — as it stands, anyone reviewing from the description will miss most of the diff.

**2. `_build_launch_q_in_segment` duplicates `build_launch` almost entirely.** 244 of its 334 lines are identical, including the full softmax chain and the cascade merge. Your docstring says only the schedule differs, and that is borne out by the code: the grid changes and the segment and herd bodies gain an `air.sequential` wrapper. That is a parameter rather than a second function. Since this design was modified twice in the past week (#2006, #2014), keeping two copies means every future fix has to be applied twice, with nothing to catch it when one of them is missed.

**3. A comment now contradicts the code it sits above.** `smolvla_vision_encoder.py:114` still states that drain herds require `runtime_loop_tiling_sizes=[2,2]` for BD-ID recycling, immediately above the two functions this PR changes to `12,6`. The same claim appears in `o_ffn_multi.py:232` and `:551`. If the explanation is that `12,6` unrolls the loop completely and leaves no BD IDs to recycle, that is worth stating explicitly, and all three comments should be corrected.

**4. `12,6` can be computed rather than hardcoded.** The PR describes it as the launch grids' own extents, and `smolvla_vision_builders.py:276` already has `tile_m` and `tile_n` from the registry spec, with `herd_m` and `herd_n` available as parameters. Deriving the value there, and keeping the environment variables as an override, would decouple the change from `VISION_N_IMAGES == 3`.

**5. Smaller points.** `SMOLVLA_FA_QSEG` is read in two separate places even though `smolvla_fuse.py` exists to prevent that. `LN_ROWS` is baked into the ELF but absent from the cache key, so changing it would silently reuse a stale cache. Nothing exercises the smolvla side or the `SMOLVLA_LN_EXT=0` fallback, so that path will quietly rot. I'd also like the three-dimensional-grid miscompile you mention to be filed before this knob is raised.

vule20:
Thanks for the review. Now addressed all five in three new commits (`5da546cf`, `d8fb8b2a`, `018b7c0a`).

1. Rewrote the PR description to cover all three commits, including the FA q-in-segment schedule and the
   `SMOLVLA_FA_QSEG` wiring.

2. `5da546cf` extracts the byte-identical pieces of the two functions — kernel externs, DMA/cascade channel
   declarations, Q/K/V/GP tensor declarations, and the cascade-merge compute — into four shared helpers
   (`_declare_flash_kernels`, `_declare_flash_channels`, `_declare_flash_tensors`, `_make_cascade_merge`),
   called from both `build_launch` and `_build_launch_q_in_segment`. I kept the causal-masking machinery and
   the actual loop-nesting (launch-grid axis vs. `air.sequential`) separate, since that genuinely is a
   different schedule rather than duplication — merging it would trade a real win for a much riskier merge
   of different control flow.

   Verified this is a pure mechanical extraction: printed the MLIR module before/after at 6 configs
   (default, causal, GQA, q_in_segment, GQA + q_in_segment, and an odd cascade-stage shape) and diffed —
   byte-for-byte identical at every one. Also re-ran this file's two lit tests' underlying commands on real
   NPU2 hardware after the refactor (default schedule and GQA 12q/6kv q_in_segment); both `PASS!`.

3. Your read was right — `12,6` fully unrolls the GEMMs' runtime loops (>= their own grid extent), which is
   exactly what avoids needing BD-ID recycling; `[2,2]` only needs it because it's smaller than the extent.
   Traced this through `AIRDependencyScheduleOpt`'s `findLargestFactor(tripCount, requested)` to confirm the
   mechanism, then rewrote the comment in all three places (backbone `o_ffn_multi.py` docstring + inline,
   and the SmolVLA encoder) to say so explicitly instead of stating `[2,2]` as a flat requirement.

4. `018b7c0a` adds `vit_ln_qkv_runtime_tiling` / `vit_o_ffn_runtime_tiling` in
   `smolvla_vision_builders.py`, right where you pointed (`tile_m`/`tile_n` from the registry spec,
   `herd_m`/`herd_n` as parameters): `grid_m = M // (tile_m * herd_m)`, `grid_n = max` across the ELF's GEMM
   slices of `N // (tile_n * herd_n)`. Confirmed it reproduces `[12, 6]` from the current shapes with no
   hardcoding. `SMOLVLA_OFFN_TILING` / `SMOLVLA_LNQKV_TILING` are override-only now.
   
   5. Same commit:
   - `SMOLVLA_FA_QSEG` is now read once (`FA_Q_IN_SEGMENT` in `smolvla_fuse.py`), imported by both the
     compile path and the cache-dir naming.
   - `LN_ROWS` is now in the cache-dir suffix (`_r{LN_ROWS}`, guarded on `LN_EXT` since it's a no-op
     otherwise).
   - Smoke-tested `SMOLVLA_LN_EXT=0` and `SMOLVLA_FA_QSEG=1` — both build clean via `make compile` with
     correctly-distinct cache directories. Didn't add these as automated lit tests, which is a real gap I'm
     leaving open rather than papering over — happy to follow up with that separately if you'd like it in
     this PR instead.
   - The 3-D grid miscompile: writing it up as its own issue now, will link it here once filed.

Please let me know if any further changes are needed.

erwei-xilinx (approved):

